PR #5, opened 2020-03-28
Broadcast message string clean EOL + format.

Tanjitsu:
Just a small pull for the broadcast message cleanup.
Added format for {nextmap} in string.

### DIFF
--- a/rcon/broadcast.py
+++ b/rcon/broadcast.py
@@ -16,8 +16,7 @@ if __name__ == '__main__':
  
     while True: 
         for time_sec, msg in msgs:
-            if msg == '/nextmap\n':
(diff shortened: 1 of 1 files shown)


rcon/broadcast.py
```diff
@@ -16,8 +16,7 @@
  
     while True: 
         for time_sec, msg in msgs:
-            if msg == '/nextmap\n':
-                msg = "Next map: {}".format(ctl.get_next_map())
+            msg = msg.rstrip("\r\n").format(nextmap=ctl.get_next_map())
```
MarechJ: Could you update the docs in the readme to explain the new format as well?

PR #6, opened 2020-03-28
Random broadcasts

Tanjitsu:
Shows a Button in settings to control Broadcast messages being randomized.
Only shows the button if you have a valid message file being broadcast.

Only issue is the dropdown state not changing when you click it.
(it updates if you refresh the page).

### DIFF
--- a/rcon/broadcast.py
+++ b/rcon/broadcast.py
@@ -4,6 +4,7 @@ import random
 
 from rcon.extended_commands import Rcon
 from rcon.settings import SERVER_INFO
+from rcon.models import init_db, enter_session
 from rcon.user_config import get_user_config, set_user_config
 
 if __name__ == '__main__':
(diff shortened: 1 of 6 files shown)


Tanjitsu:
Button state is now fixed.

rcon/broadcast.py
```diff
@@ -1,23 +1,36 @@
...
-            print(f"{time_sec} second(s): {msg}")
-            ctl.set_broadcast(msg) 
-            time.sleep(int(time_sec)) 
+    if os.path.isfile(f"{mount_path}broadcasts.txt"):
+        with open(f"{mount_path}broadcasts.txt") as f: 
+            default_msgs = [l.split(' ', 1) for l in f]
+            default_msgs = [m for m in default_msgs if len(m) == 2] # filtering out badly formated lines
+
+        if type(default_msgs) is list:
+            set_user_config('randomize_broadcast_available', True)
```
MarechJ, line 24: I wrote some code abstracting the config of the broadcast and taking care of default in `rcon/user_config.py` so we should remove this one before merging.

rcon/extended_commands.py
```diff
@@ -284,6 +285,15 @@ def do_perma_ban(self, player, reason):
         with invalidates(self.get_players, self.get_perma_bans):
             return super().do_perma_ban(player, reason)
 
+    def set_randomMessages(self, bool_str, log_info=True):
```
MarechJ, line 288: You added it so that it's exposed in the http API? 
While it's convenient to add it like that maybe we should make it leave in its own module. This classes is already super lengthy,

rcongui/src/components/SettingsView/padlock.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import FormGroup from '@material-ui/core/FormGroup';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+import Switch from '@material-ui/core/Switch';
+
+export default function Padlock({handleChange, checked, label}) {
```
MarechJ, line 6: I don't understand why we see changes that I merged a while back 🤔

rcongui/src/components/SettingsView/randomBroadcasts.js
```diff
@@ -0,0 +1,113 @@
...
+import { toast } from "react-toastify"
+import "react-toastify/dist/ReactToastify.css";
+import { showResponse } from "../../utils/fetchUtils";
+import InputLabel from "@material-ui/core/InputLabel";
+import MenuItem from "@material-ui/core/MenuItem";
+import FormControl from "@material-ui/core/FormControl";
+import Select from "@material-ui/core/Select";
+import Grid from "@material-ui/core/Grid";
+
+let active = false;
```
MarechJ, line 11: Hmm, why do you need a non-constant variable in the global scope? That does not seem right

rcongui/src/components/SettingsView/randomBroadcasts.js
```diff
@@ -0,0 +1,113 @@
...
+
+fetch(`${process.env.REACT_APP_API_URL}get_randomMessagesAvailable`)
```
MarechJ, line 13: Also making a call here? Why not in a componentDidMount ?

rcongui/src/components/SettingsView/randomBroadcasts.js
```diff
@@ -0,0 +1,113 @@
...
+
+    handler(value)
+    {
+      this.setState({ random: value })
+
+      this.props.broadcastChange(value)
+    }
+
+    render() {
+        if (!active) return null;
```
MarechJ, line 87: Ah, is it somewhat of a backward compatibility reason?

PR #10, opened 2020-06-13
Discord chat webhook

tuokri:


rcon/discord_chat.py
```diff
@@ -0,0 +1,58 @@
...
+        _id = int(resp["id"])
+        token = resp["token"]
+        return _id, token
+    except Exception as e:
+        logger.error("error parsing Discord chat webhook url: %s", e)
+        return None, None
+
+
+# Only parse ID and token and apply decorator if webhook URL was provided.
+if DISCORD_CHAT_WEBHOOK_URL:
```
MarechJ, line 55: You could also check that in `post_chat_message_to_discord` directly.
It will waste a few CPU cycles but it's also safer (it would not prevent the app from starting in case there's a problem in parse_webhook_url for example)

rcon/discord_chat.py
```diff
@@ -0,0 +1,58 @@
...
+
+from rcon.game_logs import on_chat
+
+DISCORD_CHAT_WEBHOOK_URL = os.getenv("DISCORD_CHAT_WEBHOOK")
+
+logger = logging.getLogger(__name__)
+
+
+def post_chat_message_to_discord(_, log):
+    if not DISCORD_CHAT_WEBHOOK_URL:
```
MarechJ, line 17: how about:

```
if not all([DISCORD_CHAT_WEBHOOK_URL, ...])
````

rcon/discord_chat.py
MarechJ: Actually you already check the value there, so why not moving the parsing up there?
Just add `@lru_cache` on `parse_webhook_url` no to repeat the parsin everytime and make `WEBHOOK_ID, WEBHOOK_TOKEN` local variables?

rcon/discord_chat.py
tuokri: Good idea.

rcon/discord_chat.py
tuokri: Ok. I've implemented this change now.

PR #11, opened 2020-06-22
Added server list button shown when multiple servers register 

Tanjitsu:
Currently doesn't show until you visit the front end.


rcongui/src/components/Header/serverList.js
```diff
@@ -0,0 +1,112 @@
...
+          .catch(error => toast.error("Unable to connect to API " + error));
+    }
+
+    getLocation(servers)
+    {
+      if (servers && servers.length)
+      {
+        for (const key in servers)
+        {
+          if (window.location.port === servers[key].port) return servers[key].port;
```
MarechJ, line 71: Why do you bother with that? Why not just using `window.location.origin` directly?

MarechJ: Or just `window.location` if you actually want the path as well

rcongui/src/components/Header/serverList.js
```diff
@@ -0,0 +1,112 @@
...
+      return false;
+    }
+
+    onClick(port)
+    {
+      if (port)
+      {
+        const parts = window.location.href.split(':');
+
+        window.location.href = parts[0] + ':' + parts[1] + ':' + port + window.location.pathname + window.location.hash;
```
MarechJ, line 84: You don't need the split. location is already split and has the properties you need
```
window.location.href = `${window.protocol}//${window.location.hostname}:${port}${window.location.pathname}${window.location.hash}`
```

rcongui/src/components/useStyles.js
```diff
@@ -174,6 +174,12 @@ const useStyles = makeStyles(theme => ({
   popover: {
     pointerEvents: 'none',
   },
+  serverList:{
+    position: "absolute",
```
MarechJ, line 178: Not sure if that's going to play well with the rest

rconweb/api/instances.py
```diff
@@ -0,0 +1,30 @@
+import logging
+import os
+import redis
+from rconweb.settings import ENVIRONMENT
+
+INSTANCE_KEY_PREFIX = "rcon_instance_{}"
+
+def _get_main_redis_db():
```
MarechJ, line 8: Ahaha so it worked out of the box? I expected I left some mistakes

rconweb/api/views.py
```diff
@@ -254,6 +255,16 @@ def players_history(request):
         "failed": failed
     })
 
+@csrf_exempt
+def get_instances(request):
+    instances = get_registered_instance()
+    return JsonResponse({
+        "result": instances,
+        "command": "get+instances",
```
MarechJ, line 263: typo _ instead of +

rcongui/src/components/Header/serverList.js
```diff
@@ -0,0 +1,112 @@
...
+
+    render() {
+        const { servers } = this.state
+        const { classes } = this.props
+
+        if (servers.length > 1)
+        {
+          return (
+              <React.Fragment>
+                  <Grid item className={classes.serverList}>
```
MarechJ, line 97: You probably want a `<Grid container` wrapping that no?

rcongui/src/components/Header/serverList.js
```diff
@@ -0,0 +1,112 @@
...
+    >
+      {defaultValue !== undefined ? (
+        <MenuItem value={defaultValue}>
+          <em>{defaultText}</em>
+        </MenuItem>
+      ) : (
+          ""
+        )}
+      {values.map(a => (
+        <MenuItem value={a.port}>{a.title}</MenuItem>
```
MarechJ, line 36: add a `key={a.port}` property, otherwise react is not happy

rcongui/src/components/Header/serverList.js
Tanjitsu: Yeah that function prob is not needed if you just pass window.location.path as the current value to the select ?

rcongui/src/components/useStyles.js
Tanjitsu: Works fine on my live version but your welcome to do it the correct way with grid :D

rconweb/api/instances.py
Tanjitsu: Took me a while to work out what StricRedis was but just importing and fromUrl seems to do the job.

rconweb/api/views.py
Tanjitsu: Ah ok, i just assumed that was your preferred formatting so i didn't look any deeper :D

rcongui/src/components/Header/serverList.js
Tanjitsu: key= and value= ?

rcongui/src/components/Header/serverList.js
Tanjitsu: I have 0 knowledge of your templating system Grid. i also have no need to learn it :*(

rcongui/src/components/useStyles.js
MarechJ: Did you test in responsive mode on mobile? Cause that's usually when things go wrong with the header. It's currently a bit messed up already if the server name is too long. 

Let's merge this in "various" so that I can make some tweaks before it hits master.

rcongui/src/components/Header/serverList.js
MarechJ: Ye it's just that container takes care of the alignment of items inside it. Not totally sure what's required here. It depends if the parent of this defines a container or not

rcongui/src/components/Header/serverList.js
MarechJ: It's just to give the html node a unique key so that react knows if it should re-render it the next time. It's an optimization thing

rcongui/src/components/Header/serverList.js
MarechJ: Ye try that

MarechJ:
> 
> 
> Currently doesn't show until you visit the front end.

Ye, so my guess it that Gunicorn is lazyloading the app.

Just add this option in the entrypoint.sh where gunicorn is started: https://docs.gunicorn.org/en/stable/settings.html#preload-app

MarechJ:
@Tanjitsu Did you test this branch btw? Should me merge, do we need to change more?

PR #15, opened 2020-07-29
Added variables for broadcasting all vips and a random vip

FlorisE:
Added {{vips}} and {{randomvip}} broadcasting variables.

### DIFF
--- a/rcon/broadcast.py
+++ b/rcon/broadcast.py
@@ -3,7 +3,6 @@ import time
 import logging
 import random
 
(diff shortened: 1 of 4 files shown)


FlorisE:
Added some more variables (admins, owners, seniors, juniors) and added support for variables to the welcome message and (non-auto) broadcast.

To avoid a circular reference I moved one import from the top of the file to the method in which it was used:
`from rcon.extended_commands import Rcon`

Issue #18, opened 2020-08-03
Link perma bans and blacklist


MarechJ:
Currently when you apply a permaban to someone this person also gets blacklisted (useful for multi servers as they share the same DB).

However if you remove that perma ban the blacklist stays and you have to remove it manually.

To implement this both the ban and the unban commands should manage database records using the steamid to identify the player, and if a link exists remove the blacklist on unban. 
The code should be fail proof, meaning that failing to unblacklist should not result in a failure for the unban, same cause for the ban

FlorisE:
I'll pick this up alongside with updating the reason field in the Blacklist by Steam ID form on the Settings page

FlorisE:
Still working on this, it's a bit of a pain to debug as there are many interrelated use cases:
- Ban -> Blacklist
- Blacklist -> Ban (if player is currently in the server associated with the hll_rcon_tool instance)
- Unban -> Unblacklist
- Unblacklist -> Unban (on the server associated with the hll_rcon_tool instance)
- Blacklisted (when joining server) -> Ban
- Button to remove bans for previously blacklisted players (on the other servers), only active if there are actually any applicable cases. We could also trigger this automatically when the tool is restarted.

Also Steam is down now...

MarechJ:
Done

PR #33, opened 2020-08-06
Feature/link perma bans and banlist

FlorisE:
Implementation of #18 

Also includes a small change to the broadcast.py to allow (slightly) more accurate timing

### DIFF
--- a/rcon/models.py
+++ b/rcon/models.py
@@ -50,8 +50,8 @@ class PlayerSteamID(Base):
                                     uselist=True, order_by="desc(PlayersAction.time)")
     blacklist = relationship("BlacklistedPlayer", backref="steamid",
                              uselist=False)
-    ban_log = relationship("PlayerBanLog", backref="steamid",
-                           uselist=False)
+    ban_logs = relationship("PlayerBanLog", backref="steamid",
+                           uselist=True)
     flags = relationship("PlayerFlag", backref="steamid")
 
     def get_penalty_count(self):
@@ -97,7 +97,7 @@ class PlayerSteamID(Base):
             ],
             penalty_count=self.get_penalty_count(),
             blacklist=self.blacklist.to_dict() if self.blacklist else None,
-            ban_log=self.ban_log.to_dict() if self.ban_log else None,
+            ban_logs=[d.to_dict() for d in self.ban_logs or []],
             flags=[f.to_dict() for f in (self.flags or [])]
         )
 
@@ -206,10 +206,9 @@ class PlayerBanLog(Base):
     id = Column(Integer, primary_key=True)
     playersteamid_id = Column(
         Integer, ForeignKey('steam_id_64.id'),
-        nullable=False, index=True,
-        unique=True
+        nullable=False, index=True
     )
-    ban_log = Column(String, unique=True)
+    ban_log = Column(String)
     created = Column(DateTime, default=datetime.utcnow)
 
     def to_dict(self):
--- a/rcon/models.py
+++ b/rcon/models.py
@@ -209,13 +209,11 @@ class PlayerBanLog(Base):
         nullable=False, index=True
     )
     ban_log = Column(String)
-    created = Column(DateTime, default=datetime.utcnow)
 
     def to_dict(self):
         return dict(
             steam_id_64=self.steamid.steam_id_64,
-            ban_log=self.ban_log,
-            created=self.created
+            ban_log=self.ban_log
         )
 
 
--- a/rcon/models.py
+++ b/rcon/models.py
@@ -50,6 +50,8 @@ class PlayerSteamID(Base):
                                     uselist=True, order_by="desc(PlayersAction.time)")
     blacklist = relationship("BlacklistedPlayer", backref="steamid",
                              uselist=False)
+    ban_log = relationship("PlayerBanLog", backref="steamid",
+                           uselist=False)
     flags = relationship("PlayerFlag", backref="steamid")
 
     def get_penalty_count(self):
@@ -95,6 +97,7 @@ class PlayerSteamID(Base):
             ],
             penalty_count=self.get_penalty_count(),
             blacklist=self.blacklist.to_dict() if self.blacklist else None,
+            ban_log=self.ban_log.to_dict() if self.ban_log else None,
             flags=[f.to_dict() for f in (self.flags or [])]
         )
 
@@ -195,6 +198,27 @@ class BlacklistedPlayer(Base):
             is_blacklisted=self.is_blacklisted,
             reason=self.reason
         )
+
+
+class PlayerBanLog(Base):
+    __tablename__ = 'player_ban_log'
+
+    id = Column(Integer, primary_key=True)
+    playersteamid_id = Column(
+        Integer, ForeignKey('steam_id_64.id'),
+        nullable=False, index=True,
+        unique=True
+    )
+    ban_log = Column(String, unique=True)
+    created = Column(DateTime, default=datetime.utcnow)
+
+    def to_dict(self):
+        return dict(
+            steam_id_64=self.steamid.steam_id_64,
+            ban_log=self.ban_log,
+            created=self.created
+        )
+
 
 
 class PlayersAction(Base):
--- a/rcon/player_history.py
+++ b/rcon/player_history.py
@@ -7,7 +7,7 @@ import math
 
 from rcon.models import (
     init_db, enter_session, PlayerName, 
-    PlayerSteamID, PlayerSession, BlacklistedPlayer, 
+    PlayerSteamID, PlayerSession, BlacklistedPlayer, PlayerBanLog,
     PlayersAction, PlayerFlag
 )
 from rcon.game_logs import on_connected, on_disconnected
@@ -207,6 +207,22 @@ def save_end_player_session(steam_id_64, timestamp_ms):
         sess.commit()
 
 
+def add_ban_log(steam_id_64, ban_log):
+    with enter_session() as sess:
+        player = get_player(sess, steam_id_64)
+        if not player.ban_log:
+            sess.add(PlayerBanLog(steamid=steam_id_64, ban_log=ban_entry))
+        else:
+            # for some reason there was already an entry in the playerbanlog, just update it
+            player.ban_log.ban_log = ban_log
+        sess.commit()
+
+
+def ban_and_add_log(rcon, steam_id_64, name, reason):
+    _, ban_log = rcon.do_perma_ban_and_return_ban_log(name, reason, by='BLACKLIST')
+    add_ban_log(steam_id_64, ban_log)
+
+
 def ban_if_blacklisted(rcon, steam_id_64, name):
     with enter_session() as sess:
         player = get_player(sess, steam_id_64)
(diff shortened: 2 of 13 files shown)


rcon/player_history.py
```diff
@@ -217,13 +232,17 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
...
 
         if player.blacklist and player.blacklist.is_blacklisted:
-            logger.warning("Player %s was banned due blacklist, reason: %s", str(
+            logger.warning("Player %s is blacklisted and has been banned, reason: %s", str(
                 player), player.blacklist.reason)
-            rcon.do_perma_ban(name, player.blacklist.reason)
             # TODO save author of blacklist
             safe_save_player_action(
-                rcon=rcon, player_name=player, action_type="PERMABAN", reason=player.blacklist.reason, by='BLACKLIST'
+                rcon=rcon, player_name=name, action_type="PERMABAN", reason=player.blacklist.reason, by='BLACKLIST'
```
MarechJ, line 239: It was a mistake from me to use that function because it makes a call to `get_player_info` and the dude is already gone so it won't work. I guess it kinda works because the info is cached but that's just cheer luck. 
Since we have the  actual DB record instance in that scope `player = get_player(sess, steam_id_64)` we might as well use that directly to store the action.

rcon/player_history.py
```diff
@@ -217,13 +232,17 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
...
             )
+            try:
+                ban_and_add_log(rcon, steam_id_64, name, player.blacklist.reason)
+                sess.commit()
+            except:
+                logger.warning("Something went wrong banning player %s", str(player))
```
MarechJ, line 245: `player.to_dict()` for be more useful to log, cause I think you'll just get the object id here otherwise

rcon/player_history.py
```diff
@@ -275,10 +294,31 @@ def add_player_to_blacklist(steam_id_64, reason, name=None):
                     steamid=player, is_blacklisted=True, reason=reason)
             )
 
+        try:
+            if rcon and name and ban_if_in_game:
```
MarechJ, line 298: Depending on the Rcon here makes things a bit intertwined don't you think?

Also it's a big change in behaviour, so if we keep that let's annonce it quite explicitly in the release

rcon/player_history.py
```diff
@@ -275,10 +294,31 @@ def add_player_to_blacklist(steam_id_64, reason, name=None):
...
+                logger.warning("Something went wrong banning player with steam id %s", steam_id_64)
+
         sess.commit()
 
 
-def remove_player_from_blacklist(steam_id_64):
+def _find_and_remove_perma_bans_by_ban_log(rcon, ban_logs, sess):
+    counter = 0
+    perma_bans = rcon.get_perma_bans()
+    for ban_log_entry in ban_logs:
```
MarechJ, line 312: So in this function you could totally use the Name instead of the ban_log right? 
And as an additionally safe guard you could just chec that the time or the date roughly matches. This way you don't need the whole database layer which make the whole thing much simpler and less prone to go out of sync with the server (if for some reason the server loose its history, or if people use a different client, etc...)

Your thoughts?

rcon/player_history.py
```diff
@@ -289,9 +329,72 @@ def remove_player_from_blacklist(steam_id_64):
             raise CommandFailedError(f"Player {player} was not blacklisted")
 
         player.blacklist.is_blacklisted = False
+
+        if rcon:
+            _find_and_remove_perma_bans_by_ban_log(rcon, player.ban_logs, sess)
```
MarechJ, line 334: try except?

rcon/player_history.py
```diff
@@ -289,9 +329,72 @@ def remove_player_from_blacklist(steam_id_64):
...
+
+    return return_list
+
+
+def banned_unblacklisted_players(rcon):
+    with enter_session() as sess:
+        return [p.to_dict() for p in _banned_unblacklisted_players(rcon, sess)]
+
+
+def unban_unblacklisted_players(rcon):
```
MarechJ, line 389: This file is getting quite length, maybe it would be worth extracting all the blacklist related code to a new file?

rcon/player_history.py
```diff
@@ -275,10 +294,31 @@ def add_player_to_blacklist(steam_id_64, reason, name=None):
...
 
 
-def remove_player_from_blacklist(steam_id_64):
+def _find_and_remove_perma_bans_by_ban_log(rcon, ban_logs, sess):
+    counter = 0
+    perma_bans = rcon.get_perma_bans()
+    for ban_log_entry in ban_logs:
+        if ban_log_entry.ban_log:
+            if ban_log_entry.ban_log in perma_bans:
+                rcon.do_remove_perma_ban(ban_log_entry.ban_log, 'BLACKLIST_REMOVE', False)
```
MarechJ, line 315: You'll want to invalidate the cache of perma bans for each call to a do_perma or do_remove_perma

rcon/recorded_commands.py
```diff
@@ -34,17 +35,27 @@ def do_temp_ban(self, player, reason, by):
...
         except:
             logger.exception("Unable to blacklist")
         return res
 
+    def do_remove_perma_ban(self, ban_log, by, unblacklist=True):
+        res = super().do_remove_perma_ban(ban_log)
+        if unblacklist:
+            try:
+                remove_player_from_blacklist_using_ban_log(ban_log)
+            except:
```
MarechJ, line 55: Can you narrow that down to specific exception(s)?

rcongui/src/components/PlayersHistory/index.js
```diff
@@ -281,7 +281,15 @@ const FilterPlayer = ({classes, constPlayersHistory, pageSize, total, page, setP
                             blacklisted={player.blacklisted}
                             flags={List(player.flags.map(v => Map(v)))}
                             onflag={() => setDoFlag(player)}
-                            onBlacklist={() => setDoConfirmPlayer({ player: player.steam_id_64, actionType: "blacklist" })}
+                            onBlacklist={
+                              () => setDoConfirmPlayer(
+                                {
+                                  player: player.steam_id_64,
+                                  actionType: "blacklist",
+                                  name: player.names ? player.names[player.names.length - 1] : null
```
MarechJ, line 289: Why arbitrarily picking the last name here?

rcongui/src/components/RconSettings/blacklist.js
```diff
@@ -68,16 +64,22 @@ class Blacklist extends React.Component {
...
-        </Grid>
+                <Grid item xs={2} spacing={1} className={`${classes.padding} ${classes.margin}`} justify="center" alignContent="center">
+                    <Button 
+                       color="secondary"
+                       variant="outlined"
+                       disabled={!steam_id || !reason }
+                       onClick={
+                         () => { 
+                           submitBlacklistPlayer(steam_id, name, reason)
+                             .then((res) => !res.failed && this.setState({ steam_id: "", name: "", reason: "" }))
```
MarechJ, line 76: Hmm, it may be a bit risky here to assume that this function with return a promise, or that it will return the results (in my experimental branch with the accounts, they won't)

Since the caller is responsible for sending the data, maybe it should also be responsible for managing the state, and this component just receives the values as props?

rcongui/src/components/RconSettings/rconSettings.js
```diff
@@ -39,6 +42,14 @@ class RconSettings extends React.Component {
             }))
     }
 
+    async getBannedUnblacklistedPlayers() {
```
MarechJ, line 45: Does that belong in the UI? It's something to fix the sate right? Maybe it's best done at startup or something of the sort?

rcongui/src/components/PlayersHistory/index.js
FlorisE: By adding the name we can check if the blacklisted player is currently in the server. I assumed that the last name was the most recent one. We could ban the player if in game using just the steam id but it seems like that would be expensive as it would require repetitive calls to `get_player_info` (unless we already have a very recent cache for that?).

rcongui/src/components/RconSettings/blacklist.js
FlorisE: If we want to keep both files while we are not sure that RconSettings provides a promise to Blacklist, then maybe it's better to assume the request succeeded and clear the fields regardless of the response from the server. Anything else would require RconSettings to micromanage the state of Blacklist.

rcongui/src/components/RconSettings/rconSettings.js
FlorisE: It's for server admins to unban players who were previously blacklisted and got banned, but are now unblacklisted and should hence be unbanned. This happens automatically on the game server instance associated with the rcon tool instance, but not on other servers which share the same DB. We could trigger it on start up as well, but it might still be useful to also have a manual action in case the rcon tool doesn't get restarted often?

rcon/models.py
```diff
@@ -197,6 +200,24 @@ def to_dict(self):
         )
 
 
+class PlayerBanLog(Base):
```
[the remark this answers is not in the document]
FlorisE: Trying to reproduce the ban log seems a bit brittle due to time zone differences / network delays / unsynced clocks, etc. I think storing the ban log is more straightforward.

rcon/player_history.py
FlorisE: It's a minor feature so we can strip it out

rcon/models.py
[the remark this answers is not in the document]
MarechJ: I suppose both approaches have a down sides. I do prefer those with less state. 
Also now that BM supports HLL a bit, people will start to use both client you can expect bans not to be done from this process and that's where shit goes wrong.
Ultimately it's your time, so your call :)

rcon/player_history.py
FlorisE: Well, one big "out of sync" problem is when the server and the RCON tool use different time zones or the time is misconfigured on the RCON tool, but this is actually not a problem when using the ban log. If the server lost its history then this function won't be called, because it is only called with actions related to the blacklist. If people use a different client, then they also pass the blacklist mechanism altogether, so I don't think it's a major issue (though we might want to set up some process to import existing bans into the blacklist/ban log).

rcongui/src/components/PlayersHistory/index.js
MarechJ: No unfortunately the last name is not necessarily the most recent one. For example if:
1. Connects as "foo"
2. Connects as "bar"
3 Connects as "foo"

Bar will still be the last one, actually that's an issue for going stateless as well, as we don't have the link between player_sessions and player_names, that's something we'd need to add actually (in [save_session](https://gitlab.trading.imc.intra/amsbod/compliance/six/-/blob/master/six_reporting/modules/trade_selector/tas_utils.py)

So bottomline, yes it'd be better with steam ids. If you use the Rcon.get_players. ids are in there and are cached so it's not so expensive. (Please not that occasionnaly there's a bug where HLL does not return the steamid, rare but happens)

rcongui/src/components/RconSettings/blacklist.js
MarechJ: That's ok though? It's the spirit of High Order Component. where children component are 'dumb' ones doing render only.

rcon/player_history.py
MarechJ: For the timezone issue you could add a tolerance of 24h then?

> they also pass the blacklist mechanism altogether

Fair enough

rcongui/src/components/RconSettings/blacklist.js
FlorisE: In my latest version I made it a fully controlled component, please check it again to see if I made any mistakes. It has some downsides because the rconSettings.js becomes feature envy, but it's best if we want to keep Blacklist.js as mostly render only.

rcongui/src/components/PlayersHistory/index.js
FlorisE: I see. I think this is only needed if we want to directly ban a blacklisted player who is currently in the server, but it seems to be more trouble than it's worth so I will strip it out.

FlorisE: I see. I think this is only needed if we want to directly ban a blacklisted player who is currently in the server, but it seems to be more trouble than it's worth so I will strip it out.

rcon/models.py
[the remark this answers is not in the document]
FlorisE: True, I didn't expect people to use both tools (why would you when this tool is more feature rich), but it seems like people are actually using both. I wonder if there is some way of getting the most recent server time without banning someone, so we can check if the server time matches the rcon tool time closely enough.

MarechJ:
Can you fix the small comments around Exception catching and such before we merge?

FlorisE:
> Can you fix the small comments around Exception catching and such before we merge?

What should still be fixed regarding to this?

MarechJ:
This is already integrate now that the bans by steamid are implemented on the game server side.

PR #38, opened 2020-08-20
Replace npm by yarn (much faster).

gbottieau:
Remove useless directories after build (to reduce container size)


rcongui/Dockerfile
```diff
@@ -2,26 +2,30 @@ FROM node:13-stretch
...
-RUN npm run build
-
 COPY entrypoint.sh .
+
 RUN chmod +x /code/entrypoint.sh
 
+RUN yarn run build
+
+RUN rm -rf src/
+RUN rm -rf public/
```
MarechJ, line 26: Are you sure static assets are in the build then?

rcongui/Dockerfile
```diff
@@ -2,26 +2,30 @@ FROM node:13-stretch
...
+RUN rm -rf node_modules/
```
MarechJ, line 27: By doing that you also screw with the caching right? Did you test?

rcongui/Dockerfile
```diff
@@ -2,26 +2,30 @@ FROM node:13-stretch
...
 COPY nginx.conf /etc/nginx/sites-available/default
 
 RUN mkdir /pw
 
 COPY package.json .
-COPY package-lock.json .
+COPY yarn.lock .
 
-RUN npm install
+RUN yarn install
```
MarechJ, line 13: Is there really a difference here? npm improved a lot recently and yarn is from evil facebook :P

rcongui/Dockerfile
gbottieau: Capture
Yes public directory is copied into build one.

rcongui/Dockerfile
gbottieau: Yes I tested. No cache screwed at all :)

rcongui/Dockerfile
gbottieau: NPM (two 'npm i' cmd) Done in 40.897s and Done in 31.601s
YARN (two 'yarn install' cmd) Done in 57.59s. Done in 0.41s.

'yarn is from evil facebook'  Yes... 🤣  but evil is already in the place (React 👀 ?)

Anyway, if you want want to keep npm, I can just leave 'RUN rm xxx' cmds 👍

rcongui/Dockerfile
MarechJ: Ok, so the installed packages will still be in the previous layers on the build machine but not in the final image? Good one then

rcongui/Dockerfile
MarechJ: Eheh., indeed. 
Ye let's stick with npm please. I'm used to the tool :)

gbottieau: done

PR #39, opened 2020-08-20
Fix bad splice usage to remove text history from settings.

gbottieau:
Fix bad splice usage to remove text history from settings.
Add sort on texts history.
Fix minor typo.


MarechJ (commented):
Ah nice. I noticed a bug where the wrong message would get deleted. Is that is? 
Thanks for contributing

MarechJ (approved):

gbottieau:
> 
> 
> Ah nice. I noticed a bug where the wrong message would get deleted. Is that is?
> Thanks for contributing

Yes this is it :)

PR #49, opened 2020-09-13
Feature/Add config to auto-ban based on recent VAC bans

timraay:
Implemented a config that allows admins to automatically ban players that received a VAC/Game ban within the past N days

### DIFF
--- a/.env
+++ b/.env
@@ -55,6 +55,9 @@ STEAM_API_KEY=
 # Automatically ban a player if they had a VAC ban less than N days ago
 # Settings this to 0 will disable it.
 BAN_ON_VAC_HISTORY=
+# The reason to be given for the ban
+# You may use the variables {DAYS_SINCE_LAST_BAN} and {MAX_DAYS_SINCE_BAN}
+BAN_ON_VAC_HISTORY_REASON=VAC ban history ({DAYS_SINCE_LAST_BAN} days ago)
 
 # SERVER 2
 HLL_HOST_2=
--- a/.env
+++ b/.env
@@ -52,6 +52,8 @@ DISCORD_SEND_TEAM_KILL_UPDATES=yes
 VOTE_MAP_INSTRUCTIONS=To vote write the map number in the chat
 # The steam API is used to display country flags and VAC bans. Get a key there: https://steamcommunity.com/dev/apikey
 STEAM_API_KEY=
+# Automatically ban a player if they have been VAC banned within the specified amount of days.
+BAN_ON_VAC_HISTORY=
 
 # SERVER 2
 HLL_HOST_2=
--- a/.env
+++ b/.env
@@ -52,7 +52,8 @@ DISCORD_SEND_TEAM_KILL_UPDATES=yes
 VOTE_MAP_INSTRUCTIONS=To vote write the map number in the chat
 # The steam API is used to display country flags and VAC bans. Get a key there: https://steamcommunity.com/dev/apikey
 STEAM_API_KEY=
-# Automatically ban a player if they have been VAC banned within the specified amount of days.
+# Automatically ban a player if they had a VAC ban less than N days ago
+# Settings this to 0 will disable it.
 BAN_ON_VAC_HISTORY=
 
 # SERVER 2
--- a/rcon/player_history.py
+++ b/rcon/player_history.py
@@ -1,3 +1,4 @@
+import os
 import logging
 import datetime
 from functools import wraps
@@ -12,6 +13,9 @@ from rcon.models import (
 )
 from rcon.game_logs import on_connected, on_disconnected
 from rcon.commands import CommandFailedError
+
+from rcon.steam_utils import get_player_bans, STEAM_KEY
+AUTOBAN_TRIGGER_THRESHOLD = os.getenv('BAN_ON_VAC_HISTORY')
 
 from rcon.cache_utils import ttl_cache, invalidates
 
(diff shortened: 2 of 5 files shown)


.env
```diff
@@ -52,6 +52,8 @@ DISCORD_SEND_TEAM_KILL_UPDATES=yes
 VOTE_MAP_INSTRUCTIONS=To vote write the map number in the chat
 # The steam API is used to display country flags and VAC bans. Get a key there: https://steamcommunity.com/dev/apikey
 STEAM_API_KEY=
+# Automatically ban a player if they have been VAC banned within the specified amount of days.
```
MarechJ, line 55: I'd rephrase that to somthing like: "if they had a VAC ban less than N days ago" to avoid any ambiguty.
Also make 0 the default value and specify in the command that 0 means "disabled"

rcon/player_history.py
```diff
@@ -227,6 +231,31 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
             )
 
 
+def ban_if_has_vac_bans(rcon, steam_id_64, name):
+    try:
+        autoban_threshold = int(AUTOBAN_TRIGGER_THRESHOLD)
+    except ValueError: # No (proper) value is given
+        return
```
MarechJ, line 238: If you go for the 0 as default then you could log that exception. That'd avoid some awkward trouble shooting sessions in the tech support channel :)

rcon/player_history.py
```diff
@@ -13,6 +14,9 @@
 from rcon.game_logs import on_connected, on_disconnected
 from rcon.commands import CommandFailedError
 
+from rcon.steam_utils import get_player_bans, STEAM_KEY
+AUTOBAN_TRIGGER_THRESHOLD = os.getenv('BAN_ON_VAC_HISTORY')
```
MarechJ, line 18: It's also a possibilty to default to 0 here

rcon/player_history.py
```diff
@@ -227,6 +231,31 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
...
+        if not player:
+            logger.error("Can't check VAC history, player not found %s", steam_id_64)
+            return
+
+        bans = get_player_bans(steam_id_64)
+        if bans == {}: return # Player has no bans
+        last_ban = bans['DaysSinceLastBan']
+        if int(last_ban) <= autoban_threshold:
+            reason = f"VAC ban history ({str(last_ban)} days ago)"
+            logger.warning("Player %s was banned due VAC history, last ban: %s days ago", str(player), str(last_ban))
```
MarechJ, line 252: That's an INFO rather than a warning

rcon/player_history.py
```diff
@@ -227,6 +231,31 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
...
+
+        if not player:
+            logger.error("Can't check VAC history, player not found %s", steam_id_64)
+            return
+
+        bans = get_player_bans(steam_id_64)
+        if bans == {}: return # Player has no bans
+        last_ban = bans['DaysSinceLastBan']
+        if int(last_ban) <= autoban_threshold:
+            reason = f"VAC ban history ({str(last_ban)} days ago)"
```
MarechJ, line 251: As mentionned in the issue it'd be better to make that configurable. Not everyone wants english :)

rcon/player_history.py
```diff
@@ -227,6 +231,31 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
...
+
+    with enter_session() as sess:
+        player = get_player(sess, steam_id_64)
+
+        if not player:
+            logger.error("Can't check VAC history, player not found %s", steam_id_64)
+            return
+
+        bans = get_player_bans(steam_id_64)
+        if bans == {}: return # Player has no bans
```
MarechJ, line 248: Just to be on the safe side i'd change that to:

```
if not bans:
    return
```

rcon/player_history.py
MarechJ: Also it'd be good to change the name of that variable a bit so that it contains the unit, DAYS in that case

rcon/player_history.py
```diff
@@ -227,6 +231,31 @@ def ban_if_blacklisted(rcon, steam_id_64, name):
...
+        last_ban = bans['DaysSinceLastBan']
+        if int(last_ban) <= autoban_threshold:
```
MarechJ, line 250: Are you sure this is castable to an int? I thought this could be null?

I'd rename those a bit to be clearer
```
int(last_ban) <= autoban_threshold:
```

to something of the sort:

```
days_since_last_ban <= max_days_since_ban
```

rcon/steam_utils.py
```diff
@@ -39,8 +39,7 @@ def get_player_country_code(steamd_id):
     return profile.get('loccountrycode', 'private' if is_private else '')
 
 
-@ttl_cache(60 * 60 * 24, cache_falsy=False, is_method=False)
-def get_player_has_bans(steamd_id):
+def get_player_bans(steamd_id):
```
MarechJ, line 42: could be good to cache this one as well? at least for a short period of time?

MarechJ (changes requested):
Thanks for the contribution. It looks alright, just a bunch of small comments before we merge :)

rcon/player_history.py
timraay: The message for getting banned by the blacklist wasn't configurable either. Should I make that configurable as well then?

MarechJ: Hmm, yes it is? It's a parameter from the GUI, then it get stored in the DB :) Unless you're talking about something else.
The one for VACs however should be from the `.env` as this is fully automatic. We could set it from the GUI but I think it's overkill TBH and too much extra work.

MarechJ (approved):
Thanks for the contribution, it's much appreciated

PR #50, opened 2020-09-17
Added window protocol to external playerView resources

Tanjitsu:
Request secure URL if using secure address.

### DIFF
--- a/rcongui/src/components/PlayerView/playerList.js
+++ b/rcongui/src/components/PlayerView/playerList.js
@@ -124,7 +124,7 @@ const getCountry = (player) => {
       alt={country}
       title={country ? getName(country) : ""}
       style={{ height: "12px" }}
-      src={`http://catamphetamine.gitlab.io/country-flag-icons/3x2/${country}.svg`}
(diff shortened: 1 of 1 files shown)


rcongui/src/components/PlayerView/playerList.js
```diff
@@ -192,7 +192,7 @@ const PlayerItem = ({
               className={classes.marginRight}
               target="_blank"
               color="inherit"
-              href={`https://steamcommunity.com/profiles/${steamID64}`}
```
MarechJ, line 195: Was this one actually a problem?

Now that I think of it, why don't we put https hardcoded for both. There's no reason to downgrade is there?

Tanjitsu: No it wasn't causing an issue as it doesn't fetch anything but yeah :)

MarechJ: Can you hardcode https then? And let merge it :)

Tanjitsu: Oh ok, yeah ill push it tomorrow !

MarechJ (approved):

PR #52, opened 2020-10-01
Experimental

MarechJ:
Release

- Add auto-ban feature for VAC and game ban. With audit trail in DB and on discord
- Fix bug for old admin names staying in cache after been deleted from accounts
- Attempted fix to limit the frequency of failures in game logs calls
- Fix bug that would prevent the application from starting if discord is unreachable
- Fixed bug for permaban not being recorded with triggerd from Blacklist
- Audit log on discord for permas triggered by Blacklist (that gives you some statifaction ;))
- Added new endpoint to see the last N maps played and their start / end time `/api/get_map_history` 
- Quick reminder of the other hidden features: `/api/scoreboard` and `/api/upload_vips`


geekcubed (changes requested):
The logic in the the MAX_GAME_BAN_THRESHOLD is flawed.  How would you trigger the ban process for someone with a single game ban?  

e.g. 
`{"SteamId":"xxxxxxxxxx","CommunityBanned":false,"VACBanned":false,"NumberOfVACBans":0,"DaysSinceLastBan":26,"NumberOfGameBans":1,"EconomyBan":"none"}`

If you set the game threshold to 1, the rule won't pass (needs to be greater than).  If you set it to zero, it's disabled.

Suggest you change the logic to `>=`  - "N or more"

MarechJ:
Good point. I used strictly superior as I thought it was more intuitive with the 'maximum' wording but indeed now you can ban someone with a single game ban. Will change and add a test for it. Thanks for the review.

PR #61, opened 2021-02-08
Added in a remove_all_vips command

walgarch:
Added this extra command so we can mimic the upload_vips feature via command line

### DIFF
--- a/rcon/extended_commands.py
+++ b/rcon/extended_commands.py
@@ -218,19 +218,16 @@ class Rcon(ServerCtl):
         with invalidates(Rcon.get_vip_ids):
             return super().do_add_vip(steam_id_64, name)
 
+    def do_remove_all_vips(self):
+        vips = self.get_vip_ids()
+        for vip in vips:
+            try:
+                self.do_remove_vip(vip['steam_id_64'])
+            except (CommandFailedError, ValueError):
+                self._reconnect()
+                raise
 
-    @ttl_cache(ttl=60)
-    def remove_all_vips(self):
-        with invalidates(Rcon.get_vip_ids):
-            vips = self.get_vip_ids()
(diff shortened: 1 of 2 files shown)


MarechJ:
if you add it in `extended_commands` `Rcon` cl;ass  instead then we get it in the CLI automatically and also in the HTTP API for free

MarechJ (approved):
Approved, although if moved to extended_commands.py you get the same result but also the http api

walgarch:
Can you take a look at the updated commit @MarechJ? I believe everything should be right, but best get someone else's eye on it :)

rcon/extended_commands.py
```diff
@@ -218,6 +218,21 @@ def do_add_vip(self, name, steam_id_64):
         with invalidates(Rcon.get_vip_ids):
             return super().do_add_vip(steam_id_64, name)
 
+
+    @ttl_cache(ttl=60)
+    def remove_all_vips(self):
```
MarechJ, line 223: Just change to: `do_remove_all_vips` and you're good to got.
Only the methods starting with do_ set_ or get_ are picked up

rcon/extended_commands.py
```diff
@@ -218,6 +218,21 @@ def do_add_vip(self, name, steam_id_64):
...
+        with invalidates(Rcon.get_vip_ids):
+            vips = self.get_vip_ids()
+            for vip in vips:
+                try:
+                    self.do_remove_vip(vip['steam_id_64'])
+                except ValueError:
```
MarechJ, line 229: If there a failed command it will raise `CommandFailedError` after retrying once.
So here it's better to just catch this one, and `continue` instead. The manual reconnection won't help much here

MarechJ (approved):

PR #64, opened 2021-04-03
Additional punishments for the BAN_TK_ON_CONNECT module

timraay:
Allows server admins to choose different punishments for players that pass the TK threshold at the start of their session. See below file changes, should be pretty self-explanatory.

Particularly useful for server admins that don't want to accidently ban half of the server at the start of an event.

Note that I haven't been able to test these changes yet.


rcon/game_logs.py
```diff
@@ -386,6 +386,11 @@ def auto_ban_if_tks_right_after_connection(rcon: RecordedRcon, log):
     ignore_after_death = config.get("ignore_tk_after_n_death", 1)
     whitelist_players = config.get("whitelist_players", {})
     tk_tolerance_count = config.get("teamkill_tolerance_count", 1)
+    punishment = config.get("punishment", "ban").lower()
+    if punishment not in ["ban", "kick", "punish"]:
+        logger.error("Punishment %s not recognized, defaulting to \"ban\"" % punishment)
+        punishment = "ban"
+    ban_duration = config.get("ban_duration_in_hours", 0)
```
MarechJ, line 393: You probably want to try to cast that to an `int()` :)

MarechJ (changes requested):
Looks ok, except for the danger of the ban duration being a string.

If you motivation for adding that it competitive matches the best way would be to add a toggle in the UI though

timraay:
I would also have added a toggle in the UI, but I have no experience with JS whatsoever.

Anyway, I've made the requested changes. It should now handle invalid ban durations and default to 0.

MarechJ (approved):

MarechJ:
Please test and merge if it's fine :)

MarechJ:
Were you able to test?

PR #74, opened 2021-05-16
User profile css

charlesLoiseau:


rcongui/src/components/ChatWidget/ChatWidget.js
```diff
@@ -1,35 +1,26 @@
...
 import Typography from "@material-ui/core/Typography";
 import moment from "moment";
 
 const useStyles = makeStyles((theme) => ({
   message: {
     whiteSpace: "pre-wrap",
-    textAlign: "left"
+    textAlign: "left",
+    marginTop: '5px',
+    marginBottom: '5px',
```
MarechJ, line 14: Please us material UI's spacing. The whole framework is about getting the spacing consistant :)

rcongui/src/components/ChatWidget/ChatWidget.js
```diff
@@ -1,35 +1,26 @@
...
   },
   date: {
-    color: theme.palette.disabledTextColor,
+    color: theme.palette.disabledColor,
+  },
+  padding: {
+    padding: theme.spacing(1),
+    overflow: "auto",
+    maxHeight: '68vh'
```
MarechJ, line 22: Why do you need an fixed height?

rcongui/src/components/ChatWidget/ChatWidget.js
```diff
@@ -50,12 +41,12 @@ const ChatContent = ({ data, handleMessageSend }) => {
...
       alignItems="flex-start"
       direction="column"
-      spacing={1}
       className={classes.padding}
+      wrap="nowrap"
     >
       {data?.map((message, index) => {
         return (
-          <Grid item key={index}>
+          <Grid item key={index} xs={12}>
```
MarechJ, line 49: You won't be able to use the justify from flex box if you specify a size

rcongui/src/components/ChatWidget/ChatWidget.js
charlesLoiseau: so you just scroll on the comment and not on the page and the form to send comment is always visible

PR #76, opened 2021-06-05
Add a map_id parameter to the gamescoreboard URL

timraay:
Allows for having persistent links to directly select a specific match, for instance http://62.171.172.221:7041/?map_id=3793#/gamescoreboard

This also adds a bar displaying the link that can be shared to reference the currently selected map, as well as a button to directly copy it.

One flaw. When a map id is specified in the URL it has to fully redirect the user to a URL with the new ID.

### DIFF
--- a/rcongui/src/components/Scoreboard/GamesScore.js
+++ b/rcongui/src/components/Scoreboard/GamesScore.js
@@ -21,6 +21,7 @@ import Scores from "./Scores";
 import map_to_pict from "./utils";
 import { fade } from '@material-ui/core/styles/colorManipulator';
 import { CopyToClipboard } from "react-copy-to-clipboard";
+import { useParams } from "react-router-dom";
 
 const useStyles = makeStyles((theme) => ({
   singleLine: {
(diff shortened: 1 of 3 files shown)


rcongui/src/components/Scoreboard/GamesScore.js
```diff
@@ -104,10 +112,29 @@ const GamesScore = ({ classes }) => {
...
+  let { slug } = useParams();
+
   const lastRefresh = scores.get("snapshot_timestamp")
     ? moment.unix(scores.get("snapshot_timestamp")).format()
     : "N/A";
 
+  const getShareableLink = () => {
+    return window.location.href;
+  };
+  const doSelectMap = (map_id) => {
```
MarechJ, line 124: We don't even need a event handler. 
We just need to wrap the maps with this component: https://reactrouter.com/web/api/Link

rcongui/src/components/Scoreboard/GamesScore.js
```diff
@@ -138,6 +165,13 @@ const GamesScore = ({ classes }) => {
     if (!currentMapId) {
       return;
     }
+    if (!parseInt(slug)) {
+      window.location.hash = `#/gamescoreboard/${currentMapId}`
+      console.log(`No ID supplied, redirecting to ${window.location.hash}`);
+    } else {
+      setCurrentMapId(parseInt(slug));
```
MarechJ, line 172: You should add the slug in the list of dependancies on line 182, this will cause this function to re-excute if the slug changes. And therefore you won't need the doSelectMap at all. We just need links.

PR #80, opened 2021-06-11
Text Change for WatchList

Tanjitsu:
Altered the watch list info text to remove the reference to the player as the player is never informed.

### DIFF
--- a/rcongui/src/components/PlayerView/playerActions.js
+++ b/rcongui/src/components/PlayerView/playerActions.js
@@ -128,7 +128,7 @@ class ReasonDialog extends React.Component {
                 label="Reason"
                 variant="outlined"
                 margin="dense"
-                helperText="The message that will be displayed to the player. A message is mandatory"
+                helperText="The message that will be displayed when the `Watchlist` is triggered. A message is mandatory."
               />
             )}
           />
@@ -143,7 +143,7 @@ class ReasonDialog extends React.Component {
             label="Comment"
             variant="outlined"
             margin="dense"
-            helperText="A comment that will NOT be displayed to the player"
+            helperText="Extra info attached to the players profile regarding the `Watchlist`."
           />
           
           {open.actionType === "temp_ban" ? (


rcongui/src/components/PlayerView/playerActions.js
```diff
@@ -128,7 +128,7 @@ class ReasonDialog extends React.Component {
                 label="Reason"
                 variant="outlined"
                 margin="dense"
-                helperText="The message that will be displayed to the player. A message is mandatory"
+                helperText="The message that will be displayed when the `Watchlist` is triggered. A message is mandatory."
```
MarechJ, line 131: That's not right because this component is used for all the different actions. So this will be displayed for pusnih/kick/ban as well. You could make a condition based on the `{open.actionType}`

rcongui/src/components/PlayerView/playerActions.js
```diff
@@ -143,7 +143,7 @@ class ReasonDialog extends React.Component {
             label="Comment"
             variant="outlined"
             margin="dense"
-            helperText="A comment that will NOT be displayed to the player"
+            helperText="Extra info attached to the players profile regarding the `Watchlist`."
```
MarechJ, line 146: Should remove the whatchlist bit here or use 
```
helperText={`Extra info attached to the players profile regarding the ${open.actionType}.`}
```

MarechJ (changes requested):

rcongui/src/components/PlayerView/playerActions.js
Tanjitsu: Yeah i thought it might be the case when reading the code section.
Just based on the current text i would assume the player is notified.

rcongui/src/components/PlayerView/playerActions.js
Tanjitsu: Just opened the pull to keep it in mind, i was not sure this was even the correct text :D

PR #81, opened 2021-07-31
Adding minimum hours played condition

gvisniuc:
- Adding env var
- Adding on connect condition
- Adding model


.env
```diff
@@ -65,6 +66,10 @@ VOTE_MAP_INSTRUCTIONS=To vote write the map number in the chat
 # changes on your server. If you don't want to receive that info set it to 1
 SILENT_MAP_RECORDER=
 
+# If you would like to only allow players with a certain numbers of hours played
+# Requires STEAM_API_KEY to be set
+MIN_HLL_PLAYTIME_HOURS=
```
MarechJ, line 71: I would suggest using https://github.com/MarechJ/hll_rcon_tool/blob/master/rcon/config.py#L11  that grab the config from `confg/config.default.py` (or the user override in `config/config.yml`

Because the `.env` causes conflict everytime user pull the repo so I'm trying not to use  it anymore.
As users are not supposed to change the config.default they don''t have this problem anymore.

rcon/hooks.py
```diff
@@ -232,6 +241,9 @@ def handle_on_connect(rcon, struct_log):
             )
 
     timestamp = int(struct_log["timestamp_ms"]) / 1000
+
+    kick_if_low_playtime(rcon=rcon, struct_log=struct_log, steam_id_64=steam_id_64)
```
MarechJ, line 245: I'd suggest just adding the `@on_connected` on the `kick_if_low_playtime` function directly
It provided some isolation in case there a problem and won't break the rest of the features in that hook.

rcon/steam_utils.py
```diff
@@ -18,8 +19,12 @@
 @ttl_cache(60 * 60 * 24, cache_falsy=False, is_method=False)
 def get_steam_profile(steamd_id):
     profiles = get_steam_profiles([steamd_id])
+
     if not profiles or len(profiles) == 0:
         return None
+    
+    time_played = get_hll_playtime_in_hours(steam_id=steamd_id)
```
MarechJ, line 26: You could test if the profile is public here first, that'd avoid making an API call for nothing. We need to be mindful of the time delays we add to the even loop as it's been growing a lot already.

rcon/steam_utils.py
```diff
@@ -40,6 +45,38 @@ def get_steam_profiles(steam_ids):
...
+def get_hll_playtime_in_hours(steam_id):
+    if not STEAM_KEY:
+        return None
+    try:
+        api = WebAPI(key=STEAM_KEY)
+        # https://developer.valvesoftware.com/wiki/Steam_Web_API#GetOwnedGames_.28v0001.29
+
+        
+        response = api.IPlayerService.GetOwnedGames_v1(
+            steamid=steam_id,
```
MarechJ, line 58: It's not possible to query that in bulk? Just like we do in `get_steam_profiles` 

The `get_steam_profiles` function is used as a cron job once a day to refresh the steam profiles of the players in your DB btw, so if you'd just inject that info in the return value of `get_steam_profiles` itself then you get that for free.
The only down side is that it's a bit less nice when accessing the value when compared to the prime attribute that you added.

MarechJ (changes requested):

rcon/hooks.py
```diff
@@ -184,6 +185,13 @@ def ban_if_has_vac_bans(rcon: RecordedRcon, steam_id_64, name):
             except:
                 logger.exception("Unable to send vac ban to audit log")
 
+def kick_if_low_playtime(rcon, struct_log, steam_id_64):    
+    hll_timm_played = get_hll_playtime_in_hours(steam_id_64)
+    if hll_timm_played is None or hll_timm_played < MIN_HLL_PLAYTIME_HOURS:
+        logger.info(
+            f"Kicking player %s for having less than {MIN_HLL_PLAYTIME_HOURS} hours played ({hll_timm_played})"
+        )
+        rcon.do_kick(player=struct_log["player"], reason=f"Playtime < {MIN_HLL_PLAYTIME_HOURS} hours or Steam Profile not public", by="VLL Admin")
```
MarechJ, line 194: It's important to make that message configurable btw

rcon/hooks.py
```diff
@@ -184,6 +185,13 @@ def ban_if_has_vac_bans(rcon: RecordedRcon, steam_id_64, name):
             except:
                 logger.exception("Unable to send vac ban to audit log")
 
+def kick_if_low_playtime(rcon, struct_log, steam_id_64):    
+    hll_timm_played = get_hll_playtime_in_hours(steam_id_64)
+    if hll_timm_played is None or hll_timm_played < MIN_HLL_PLAYTIME_HOURS:
```
MarechJ, line 190: So that also kicks players that have a private profile?

For other features like this I add an immunity flag and I also make the VIPs safe example here: https://github.com/MarechJ/hll_rcon_tool/blob/master/rcon/auto_kick.py#L30  and there https://github.com/MarechJ/hll_rcon_tool/blob/master/rcon/game_logs.py#L377

PR #83, opened 2021-08-03
Use npm ci to install JS deps

royallthefourth:
This is faster and [recommended by NPM](https://docs.npmjs.com/cli/v7/commands/npm-ci) for automated build processes


MarechJ:
ok, i didn't know that, thanks for the patch

PR #84, opened 2021-08-03
Skip sourcemap generation on arm32

royallthefourth:
The React build process runs out of memory on small systems with the default settings. Skipping sourcemap generation saves a lot of memory, allowing the process to complete on these arm32 devices which typically have only 2 gigs of RAM.


MarechJ:
Any downsides? Nothing's for free in this world :)

royallthefourth:
The only downside is that arm32 users wouldn't be able to use their in-browser debuggers to trace code. Fortunately, nobody uses that type of machine for dev work anyway.

PR #85, opened 2021-08-22
Autosettings improvements - UI & API integration, more conditions, file structure rework & more

timraay:
- Implemented two new API endpoints, `get_auto_settings` and `set_auto_settings`
- Implemented a text field into the UI showing the current auto settings, which allows editing, saving, and forwarding it to all servers
- Added 4 new conditions, `time_of_day`, `online_mods`, `ingame_mods` and `current_map`
- Redone the file structure of auto settings
- Commands starting with `set_` called from the defaults are no longer ran if they are also called from a validated rule. This not only saves resources but also makes sure certain settings are not shortly switched to their "default" values when they shouldn't.
- Added a way to upgrade auto settings to the new format
- Recorded a [new video guide](https://youtu.be/2IKZwHj9PJw) :)

I've tested it all and it should work fine.
image


config/auto_settings.default.json
```diff
@@ -1,224 +1,136 @@
...
-              "num": 1
-            }
-          ]
-        ]
-      ]
-    ]
+      }
+    }
+  ],
+  "_available_commands": {
```
MarechJ, line 62: oh damn you typed all that.
We can generate that stuff automatically just like I do for the api endpoints

rcon/auto_settings.py
```diff
@@ -76,18 +150,28 @@ def run():
...
+                try:
+                    conditions.append(create_condition(c_name, **c_params))
+                except ValueError:
+                    logger.exception("Invalid condition %s %s, ignoring...", c_name, c_params)
+                except pytz.UnknownTimeZoneError:
+                    logger.exception("Invalid timezone for condition %s %s, ignoring...", c_name, c_params)
+            
+            if all([c.is_valid(rcon=rcon) for c in conditions]):
+                saved_commands = {**saved_commands, **{name: params for (name, params) in commands.items() if name.startswith('set_')}}
+                do_run_commands(rcon, {name: params for (name, params) in commands.items() if not name.startswith('set_')})
```
MarechJ, line 170: I don't understand the deal with the set_ or not set_ 🤔

rconweb/api/auto_settings.py
```diff
@@ -0,0 +1,75 @@
...
+    if not settings:
+        return api_response(error="No auto settings provided", failed=True, status_code=400)
+
+    try:
+        # Check if valid JSON and indent for readability
+        settings = json.loads(settings)
+    except JSONDecodeError:
+        return api_response(error="No valid JSON provided", failed=True, status_code=400)
+
+    with open(CONFIG_DIR+f'auto_settings_{server_number}.json', 'w+') as f:
```
MarechJ, line 57: I'm not sure if saving to a file is a good idea, might open doors for hacks. 
Usually for user config from the UI with save that in the DB

MarechJ:
Looks good. I'd just add Monaco editor for the GUI.
Also did you test that on your prod? We need to run that for a while before we release, that's usually what I do, I run the experimental branch on HLL France servers

timraay:
Like suggested, the settings are now stored in the DB, and Monaco editor is now used instead. Additionally, the auto settings are now consistently loaded into the UI, and there's [this flowchart I made](https://github.com/timraay/hll_rcon_tool/blob/autosettings-api/images/autosettings_flow.PNG).

The script to upgrade from the old format has also been updated accordingly. Users will have to run the below command in their shell for each server.
```
docker-compose exec backend_1 /code/upgrade_auto_settings.py
```

MarechJ (approved):

PR #115, opened 2022-07-16
Add support for trigger start words

Lelleck:
Adds support for trigger words that only trigger if they appear at the start of the message. This can be set via the `DISCORD_PING_TRIGGER_START_WORDS` in the `.env` file.


MarechJ:
Hi,

Thanks for the contribution. Sorry I didn't see that sooner. 

Could we use the yaml config though? Using the `get_config` call from python? I don't touch that `.env` file anymore that it create conflicts when people pull the git repo for release and as a good chunk of them are not technical it makes it hard for them.

PR #121, opened 2022-08-18
Team Balancing Feature

cemathey:
I think this is in a workable state for others to use or provide feedback on.

I did not test it rigorously (I don't have a discord to test the webhooks for instance) but I did sit around for a few hours with both of my HLL accounts (and a few community members) and tried the different methods.

I was mostly doing it through the REPL but if you set your environment variables you should be able to do it through the CLI.

I did not hook it up to the GUI or anything, with the way it works my intention would be for there to be a button admins could click on when they want to run.


cemathey:
I'm going to close this for now and add some team shuffling stuff and add it to the GUI.

PR #122, opened 2022-08-25
Force discord.py version to be 1.7.3

timraay:
The library discord.py released a new 2.0.0 version a couple of weeks ago, which includes some breaking changes. For the CRCON, this means that webhooks would break. Since the library is used for nothing else and the API is not being deprecated it is fine to keep using the old version.

Currently in `requirements.txt` the latest version is downloaded. We simply change that to 1.7.3, which is the most recent version prior to 2.0.0.


MarechJ:
I pushed the same thing on master so Im closing this one

PR #127, opened 2022-09-10
Feature/live logs filters

cemathey:
Add a slider to toggle live log actions between inclusivity and exclusivity.

The live logs view previously would only allow an admin to select specific actions to view. If an admin wanted to see everything except one or two things they had to select a large number of actions every time they used this view.

The slider allows an admin to change between which behavior they prefer, it defaults to the previous (expected) behavior so there should be minimal impact to the user base despite adding an extra control.

The commit history is a little messy because I didn't originally use any branches so I reverted those changes so I could have a clean branch that matches the base repository.

Some of these changes are formatting only because my set up auto formats on save with `black`.

I tested this on our event server and it appears to work fine with different combinations of users and actions with both toggle values. We don't have a ton of data on our event server so if this has negative performance implications I wouldn't see them.

If you would like to see it in action I can grant you temporary access to our event server RCON.


cemathey:
This shouldn't be approved yet, I made a [breaking change](https://github.com/cemathey/hll_rcon_tool/blob/feature/live_logs_filters/rconweb/api/logs.py#L64) here when I changed `player_name` to `player1_name` for consistency.

I either need to revert that change or update the corresponding field name [here](https://github.com/MarechJ/hll_rcon_tool/blob/master/rcon/models.py#L343).

rcon/game_logs.py
```diff
@@ -312,46 +319,69 @@ def get_recent_logs(
     min_timestamp=None,
     exact_player_match=False,
     exact_action=False,
+    actions_filter_in=True,
```
MarechJ, line 322: Why not calling this one "inclusive_filter' it's not great but maybe a bit more intutive than the current one?

rcongui/src/components/LogsView/logs.js
```diff
@@ -175,6 +183,7 @@ class Logs extends React.Component {
                 />
               )}
             />
+            <Padlock handleChange={() => this.setActionsFilterIn()} checked={actionsFilterInEnabled} label="Filter out/in"/>
```
MarechJ, line 186: The label might be confusing to users why not a Select with "inclusive/exclusive" as options, you can keep true/false as values so it's compatible with the rest

MarechJ (approved):

rcon/game_logs.py
cemathey: That would be much more clear! Changes made.

rcongui/src/components/LogsView/logs.js
cemathey: It does look much better as a select control, changes made.

cemathey:
This should be ready for merge with these latest commits. The UI has been updated so it is a select control instead of the toggle and it's been integrated into the grid layout (I adjusted the sizes on other controls to fit it into the 12 column MUI layout).  I tried the layout with different window sizes and I think it matches the appearance/sizing behavior of the other controls.

I manually tested it on our event server and the behavior is working correctly, your credentials are still active if you'd like to see it working live.

There are some unrelated formatting changes from my editors code auto formatters, I hope that's not an issue but if it is I can go through and rework it to get rid of them.

PR #136, opened 2022-09-19
Fix inconsistent map naming

zMoooooritz:
Should be relatively straight forward.

Fixing the current inconsistencies in the naming scheme.

Hope the changes to the way SME and SMDM are named are okay.


rcon/utils.py
```diff
@@ -102,19 +102,19 @@ def get_map_side(map_):
     "foy_warfare": "Foy",
     "foy_offensive_ger": "Foy Offensive (GER)",
     "foy_offensive_us": "Foy Offensive (US)",
-    "stmariedumont_warfare": "St Marie du Mont",
-    'stmariedumont_off_us': "St Marie Du Mont Offensive (US)",
-    'stmariedumont_off_ger': "St Marie Du Mont Offensive (GER)",
+    "stmariedumont_warfare": "Sainte-Marie-du-Mont",
```
MarechJ, line 105: That's too long. Character real estate is expensive. St. is a proper abreviation for Saint / Sainte, that's what you'd see on road signs in Fance, we also use the shortcut for post and such. To be proper you coud add an e, sot "Ste"
dash are not required either and would prevent word / line breaks in some places so it's not ideal either.And finally to be totally pedantic "du" should not be capitalised as it's just a connecting word like "of the" 

See the image below. And no not all french dress like that ahaha only farmers from my grandpa's generation ;)
image

rcon/utils.py
```diff
@@ -102,19 +102,19 @@ def get_map_side(map_):
...
     "hurtgenforest_offensive_US": "Hurtgen Forest Offensive (US)",
-    "utahbeach_warfare": "Utah beach",
+    "utahbeach_warfare": "Utah Beach",
     "utahbeach_offensive_us": "Utah Beach Offensive (US)",
     "utahbeach_offensive_ger": "Utah Beach Offensive (GER)",
     "omahabeach_offensive_us": "Omaha Beach Offensive (US)",
-    "stmereeglise_warfare": "St Mere Eglise",
-    "stmereeglise_offensive_ger": "St Mere Eglise Offensive (GER)",
-    "stmereeglise_offensive_us": "St Mere Eglise Offensive (US)",
+    "stmereeglise_warfare": "Sainte-Mère-Église",
```
MarechJ, line 115: Same here. Ste and no - please.

MarechJ (changes requested):
I'm fine to merge that. Thank's for the contrib. However please change the Ste back and remove the -, the rest looks better.

zMoooooritz:
Alright, hank you for your detailed response and the development of this great tool.

Changed "Sainte" to "Ste" and removed the dashes. Good point concerning line breaks and representation, did not think this one through.
The capitalized "Du" was included in the previous version in two instances and I changed them to "du".

In this case I will be pedantic as well change "Hurtgen" to "Hürtgen" to feature the umlaut "ü" as it is done in Germany ;)

PR #137, opened 2022-10-09
Feature/team balance

cemathey:
Adds a module and API commands for balancing (evening team sizes) and shuffling teams by different criteria/methods. I personally do not think that a shuffle feature would be very popular (outside of events run by servers perhaps) but it is a feature that people ask for, so I've added it.

There is a a very crude UI added to the game view but this was only so that I could test it on a real server. **It is not suitable for production** and needs to be cleaned up/refined quite a bit, but frankly I burned out on doing this part because my React/Javascript knowledge is quite limited.

Especially important on the current temporary/testing UI, it has no confirmation dialog or any other way to prevent accidental clicks from shuffling teams which would be **disastrous** in a real game.

I am confident the backend works correctly, I did a combination of manual testing on our event server with real players and unit tests, to include some property based testing with hypothesis to make sure shuffled teams would always end up with equitable team sizes.

This also updates `USERGUIDE.md` to document how the feature works.

Logging is performed in a separate file to make it easier for admins to see what is happening without having to filter through `api.log` but of course RCON actions are also still logged in `api.log`.

This also adds an optional new Discord Webhook so that server owners can if they choose record all shuffle/balancing actions and expose it to their communities without having to also expose all of their other audited RCON actions.

At the moment it makes no consideration whatsoever for trying to try to maintain squads while shuffling, I may circle back and work on this in the future, but it's a tough problem if you also want to actually shuffle teams up, and not everyone in a squad together is grouped up out of game. If the party feature that is on the roadmap is released and this information is available to RCON then I will definitely come back to this.

Of note, this pulls a single team view for both balancing/shuffling and works with that team view from start to finish, this means that the information may be stale if players change roles, teams, leave, etc. from when it was grabbed, but the data we're working with is at least always internally consistent. This also means that it takes a few seconds for a shuffle or balance to complete.

Currently there is a hard coded **5 second limit** between running either command that is checked on the backend (the time stamp is stored in Redis) to prevent doubling up on commands on accident.


cemathey:
Closing this for now because I need to fix merge conflicts and I have to finish the UI, will get back to it at some point.

PR #138, opened 2022-10-18
Add remagen image

zMoooooritz:


MarechJ:
Thanks for the addition.
I believe it's also broken for night map no? But I think that requires a bit of a backend change too

PR #140, opened 2022-11-08
Feature/expiring vip

cemathey:
Adds the ability to set an expiration date for VIP status and automatically remove expired VIPs.

I tried to do this is the most optional/backwards compatible way possible, users can't opt out of the GUI changes in the front end but they don't need to enable the removal service and they can set players to have indefinite VIP.

We've been running this on our main server without issue for several days now and our event server prior to that without issue.

Expiration timestamps are tracked locally in RCONs database as opposed to including the data in the name field that is sent to the game and can be optionally downloaded with the VIP list and parsed on upload. This supports transferring VIP lists to older versions of CRCON, potentially to different RCON implementations and also allows users to create VIP lists from other sources (Excel, whatever) and process them as long as they use a date [dateutil parser](https://dateutil.readthedocs.io/en/stable/parser.html) can parse.

Updates `USERGUIDE.md` to document how the feature works for users.

Adds an optional webhook for the removal of expired VIPs.

Screenshot 2022-11-08 113330

**Backend changes:**

Adds `player_vip` table and an alembic migration.

Adds an `expiring_vips` service and updates `default_config.yml` to  allow enabling/disabling the service, set a Discord webhook and configure the frequency of removing expired VIPs.

Updates `rcon/extended_commands.py` to augment the results from `get_vip_ids()` to include expiration dates.
Updates `rcon/models.py` to add the SQLAlchemy model for the `player_vip` table and update `PlayerSteamID` with the relationship to it.
Updates `rcon/player_history.py` to augment `get_players_by_appearance()` with the players VIP expiration timestamp.
Updates `rcon/recorded_commands.py` to override `do_remove_vip` and `do_add_vip` to create/update/delete expiration records.
Updates `rcon/workers.py` to include expiration timestamps in `bulk_vip` and fix a typo (`bluk_vip`).
Updates `rconweb/api/vips.py` to include expiration dates in `async_upload_vips`.

**UI changes.**

Updates the Player Tile to reflect a users VIP expiration (null, never, in the past, or in the future) with humanized dates, and the actual timestamp on mouse over.

userguide_vip_player

Changes the behavior of the player tile star from a binary toggle to a dialog with a date picker and some shortcuts for adding/subtracting days.

userguide_vip_dialog

Changes the `Manage VIP` portion of the settings to support optionally including expiration dates on upload/download and add a date picker when adding a VIP, defaulting to indefinite VIP.

Screenshot 2022-11-08 104509


rcon/recorded_commands.py
```diff
@@ -61,20 +85,91 @@ def do_switch_player_now(self, player, by):
...
-                    'profile': profiles.get(p.get(STEAMID)),
-                    'is_vip': p.get(STEAMID) in vips
+                    "profile": profiles.get(p.get(STEAMID)),
+                    "is_vip": p.get(STEAMID) in vips,
                 }
             )
 
         return players
+
+    def do_remove_vip(self, steam_id_64):
```
MarechJ, line 106: Why adding that in the recorded commands and not the extended one?

MarechJ (approved):

PR #144, opened 2022-11-13
Add user subscribable webhook events for log events

cemathey:
Opening this for discussion.

A user requested the ability to be pinged on match end, rather than make a specific hook for that I've added a generic way to subscribe to any kind of log event.

Things I like:
- Users can opt into any log event they want (for instance vote kicks, match ending, etc.)
- Users can subscribe more than one web hook per event
- Role/user mentions can be chosen per web hook
- For users that run more than one server they can manage their hooks by server

Things I don't like:
- There's considerable overlap with some of the other discord web hooks, for instance the chat one, which might confuse users
- The yaml configuration is a bit strange looking due to the nested map/sequences which isn't the most user friendly
- Requires an RCON restart to make any changes take effect
- The embed and log format isn't user customizable

This also formats a few files with black and removes unused imports.

Built a freshly pulled repo and tested it on our event server/Discord successfully.


rcon/game_logs.py
```diff
@@ -4,42 +4,50 @@
...
+    "CONNECTED": [],
+    "DISCONNECTED": [],
     "KILL": [],
-    "CAMERA": [],
-    "TK": [],
-    "MATCH": [],
     "MATCH START": [],
     "MATCH ENDED": [],
+    "MATCH": [],
+    "TEAM KILL": [],
```
MarechJ, line 42: You should move the teamkill / connect / disconnect back up. We want those to be executed quickly

config/default_config.yml
```diff
@@ -275,3 +275,57 @@ SCOREBOT:
...
+  TEAMSWITCH:
+  # Automatic kicks for team kills
+  TK:
+  TK AUTO:
+  TK AUTO BANNED:
+  TK AUTO KICKED:
+  # Vote kicks
+  VOTE:
+  VOTE STARTED:
+  VOTE COMPLETED:
```
MarechJ, line 331: I like it. However it's not really fool proof. I betcha that a few user will use them all and that will end up being rate limited on discord or (both) slow the event loop so much due to IOs that the actual important critical features will start to pile up a delay. 

I suppose we could start circumventing that by having 2 loops, 1 that does the critcal time sensitive stuff. And the other that does all the discord IO stuff (both the ones configured by user and the ones we already had builtin if we still have those).

I had concerns about those delays already a few times, everytime I added hooks actually.

config/default_config.yml
```diff
@@ -275,3 +275,57 @@ SCOREBOT:
       - WHAT_IS_A_BREAK
       - SURVIVORS
       - U_R_STILL_A_MAN
+
+
+# This allows you to subscribe any number of Discord webhooks and optional roles and/or users to mention
```
MarechJ, line 280: We have a generic piece in the UI for that btw. Not sure how much work it'd be to reuse

rcon/game_logs.py
cemathey: Whoops, I didn't even think about that, I sorted it so I had an easier time making sure I had each type in both files.

config/default_config.yml
cemathey: I think I'd prefer to leave it in `config.yml` for now as the settings UI is a bit cluttered, if we break the settings in the frontend out into multiple views in the future we could add it.

config/default_config.yml
cemathey: I tweaked the documentation slightly to mention rate limiting.

We could remove some of the more redundant events, for instance only offer `CHAT` and `TK` instead of each sub type, or we could do that and remove any of the events that are handled in other places like `CAMERA`.

Performance wise what if we start using asynchronous web hooks? Not every hook would benefit but for those that do we could get back to the event loop more quickly.

MarechJ: ye I guess that's a way

MarechJ (approved):
We can merge this but I'd add a strong warning about not overloading it in the config

PR #146, opened 2022-11-14
Make automod use message to directly contact each player

FlorianSW:
Of course this is not tested in real life, yet 🙈


cemathey:
Builds for me on a fresh repository clone, and the tests pass. I'm not familiar with squad automod at all, we don't use it for our server.

The only thoughts I have on this are would it be a bad idea to turn the `default_config.yml` strings into placeholders that we could then str.format() to make them internally consistent?

For instance users can change `warning_interval_seconds` to something besides 60, but `warning_message: "Warning! Your squad does not have an officer. Players of squads without an officer will be punished, then kicked.\nNext check will happen automatically in 60s."` always says 60 and requires users to be aware of and able to change that to be consistent.

We wouldn't be accepting arbitrary input from the internet/random so I don't think it's an inherently terrible idea, but it would provide more opportunity to confuse users.

rcon/squad_automod/automod.py
```diff
@@ -219,80 +222,59 @@ def get_punitions_to_apply(rcon, config: NoLeaderConfig) -> PunitionsToApply:
...
-                if state == PunishStepState.wait:
-                    punitions_to_apply.pending_warnings[team].append(squad_name)
-                    punitions_to_apply.add_squad_state(squad_name, squad)
-                if state != PunishStepState.go_to_next_step and state != PunishStepState.disabled:
-                    continue
+                    state = should_warn_player(watch_status, config, squad_name, player)
+
+                    if state == PunishStepState.apply:
+                        punitions_to_apply.warning.append(a_player)
+                        if not state_added:
```
MarechJ, line 240: You could also just add it every time and handle duplicates in the add_squad_state to make this bit a bit simple

rcon/squad_automod/automod.py
```diff
@@ -219,80 +222,59 @@ def get_punitions_to_apply(rcon, config: NoLeaderConfig) -> PunitionsToApply:
...
+                            punitions_to_apply.add_squad_state(squad_name, squad)
+                            state_added = True
+                        continue
+
+                    if state == PunishStepState.wait:
+                        punitions_to_apply.pending_warnings.append(a_player)
```
MarechJ, line 246: We could remove that now. I only added the pending warning to have the full picture everytime there's a broadcast. But if we're not using broadcasts anymore it's irrelevant

rcon/squad_automod/automod.py
```diff
@@ -219,80 +222,59 @@ def get_punitions_to_apply(rcon, config: NoLeaderConfig) -> PunitionsToApply:
...
-{allies}
-{axis}
-{config.warn_message_footer}
-""", f"{allies} {axis}"
+def get_warning_message(config: NoLeaderConfig) -> str:
+    if config.warn_message_header != "" or config.warn_message_footer != "":
+        return f"""{config.warn_message_header}
+{config.warn_message_footer}"""
+    else:
+        return config.warning_message
```
MarechJ, line 277: I think this is a bit confusing. Why don't we just keep the old parameter? We should actually show how many warning a player has left before he's going to get slapped, that's a good motivation. We can insert it between the header and footer

rcon/squad_automod/automod.py
```diff
@@ -308,12 +290,15 @@ def _do_punitions(
         return
     for aplayer in players:
         try:
-            if method == "punish":
+            if method == "message":
+                audit(config, f"Warnings: {message}")
+                rcon.do_message_player(aplayer.player, aplayer.steam_id_64, message, by=AutomodUsername)
```
MarechJ, line 295: I'm worried that with all those calls we're going to add quite some latency and we may start to see more false positives if ppl take the officer role while a punition loop is applying. 

We could either use the thread pool I added a while back. or as a simpler solution, start with the punish which should be timely. Then the messages and kicks. What do you think?

rcon/squad_automod/models.py
```diff
@@ -80,8 +82,8 @@ class ASquad:
 
 @dataclass
 class PunitionsToApply:
-    warning: Mapping[str, List[str]] = field(default_factory=lambda: {"allies": [], "axis": []})
-    pending_warnings: Mapping[str, List[str]] = field(default_factory=lambda: {"allies": [], "axis": []})
+    warning: List[APlayer] = field(default_factory=list)
+    pending_warnings: List[APlayer] = field(default_factory=list)
```
MarechJ, line 86: we can remove that one

cemathey:
> > Builds for me on a fresh repository clone, and the tests pass. I'm not familiar with squad automod at all, we don't use it for our server.
> 
> Do they? At least one test failed for me, but it fails on a clean repo as well 🙈 (I also had to change one test, as it was failing on a clean repo as well) :) But good to know. (It was the config test, however, as I did not change anything there, I "dismissed" that for now).
> 

Sorry I meant the tests for squad automod passed for me, I also got failing ones from a different test file.

rcon/squad_automod/automod.py
FlorianSW: I thought so to and removed it. At some point it made sense to me to still have it. I'll revisit that part and either give you a response or remove it 😅

rcon/squad_automod/automod.py
FlorianSW: 👍

rcon/squad_automod/automod.py
FlorianSW: I liked the idea of @cemathey. Instead of building this on our own, we could provide the user the possibility to write the whole message on their own. We could then provide them with variables they can use as placeholders, like the players name, the number of warnings the player got already, the warnings in total that are required to get to the next stage, and so on.

rcon/squad_automod/automod.py
FlorianSW: I think with the threadpool we would be still constrained if the server where the tool is running is not powerful enough to get the work done. Changing the order would probably the best thing (I would also do the kicks before the messages, isn't it? It's pretty destructive as well)

rcon/squad_automod/automod.py
FlorianSW: I added the formatting option now. I think this way we give the community a much more flexible solution. What do you think? :) If this is something we like, we could also add this to the other messages as well (however, there are probably much more restrictions with regards to the length 🙈

MarechJ:
Looks good. We should way for the update before merging though I guess. Or we make a beta branch and I add a docker build on top

PR #149, opened 2022-11-16
Clean up and adapt public_info for new data

zMoooooritz:
Include the newly available information (via the `gamestate` command) into the data returned by public_info.

Unfortunately the `time_remaining` does not really contain useful information (not even sure what the game-server will return for this value on offensive maps).
Therefore the cumbersome way of loading the start time of the current map via the MapsHistory is still included.

Furthermore redundant fields are removed and additional information for the `next_map` is included.

Removed the `nb_players` field and instead added the `max_player_count` field.

### DIFF
--- a/rconweb/api/views.py
+++ b/rconweb/api/views.py
@@ -82,30 +82,44 @@ def get_version(request):
 
 @csrf_exempt
 def public_info(request):
-    status = ctl.get_status()
+    gamestate = ctl.get_gamestate()
+    curr_players, max_players = tuple(map(int, ctl.get_slots().split("/")))
     try:
-        current_map = MapsHistory()[0]
+        current_map_start = MapsHistory(max_len=1)[0]["start"]
     except IndexError:
         logger.error("Can't get current map time, map_recorder is probably offline")
-        current_map = {"name": status["map"], "start": None, "end": None}
-    current_map = dict(
-        just_name=map_name(current_map["name"]),
-        human_name=LONG_HUMAN_MAP_NAMES.get(current_map["name"], current_map["name"]),
-        **current_map,
-    )
-    vote_status = get_votes_status(none_on_fail=True)
-    next_map = ctl.get_next_map()
+        current_map_start = None
+
+    def explode_map_info(game_map: str, start) -> dict:
+        return dict(
+            just_name = map_name(game_map),
+            human_name = LONG_HUMAN_MAP_NAMES.get(game_map, game_map),
+            name = game_map,
+            start = start
+        )
+
     return api_response(
-        result=dict(
-            current_map=current_map,
-            **status,
-            vote_status=vote_status,
-            next_map=next_map,
+        result = dict(
+            current_map = explode_map_info(gamestate["current_map"], current_map_start),
+            next_map = explode_map_info(ctl.get_next_map(), None),
+            player_count = curr_players,
+            max_player_count = max_players,
+            players = dict(
+                allied = gamestate["num_allied_players"],
+                axis = gamestate["num_axis_players"]
+            ),
+            score = dict(
+                allied = gamestate["allied_score"],
+                axis = gamestate["axis_score"]
+            ),
+            vote_status = get_votes_status(none_on_fail=True),
+            name = ctl.get_name(),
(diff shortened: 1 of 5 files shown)


rconweb/api/views.py
```diff
@@ -82,30 +82,44 @@ def get_version(request):
...
+            vote_status = get_votes_status(none_on_fail=True),
+            name = ctl.get_name(),
+            short_name = os.getenv("SERVER_SHORT_NAME", "HLL RCON"),
             public_stats_port=os.getenv('PUBLIC_STATS_PORT', "Not defined"),
             public_stats_port_https=os.getenv('PUBLIC_STATS_PORT_HTTPS', "Not defined")
         ),
-        failed=False,
-        command="public_info",
+        failed = False,
+        command = "public_info",
```
MarechJ, line 122: The norm is no space for keyword arguments. 
I think we should run `black` on the whole codebase.

rconweb/api/views.py
```diff
@@ -82,30 +82,44 @@ def get_version(request):
...
+            start = start
+        )
+
     return api_response(
-        result=dict(
-            current_map=current_map,
-            **status,
-            vote_status=vote_status,
-            next_map=next_map,
+        result = dict(
```
MarechJ, line 102: Is this backward compatible? I think it's used in the public gamestate view and in the scorebot.

MarechJ (commented):
Looks good but want to make sure it's backward compatible

rconweb/api/views.py
zMoooooritz: Done, yes this sounds like a good idea!

rconweb/api/views.py
zMoooooritz: Good catch! Should have checked that beforehand but thought all the other things would use the "internal structure".
That's a dependency I did not think of, that does look somewhat cumbersome.
But I (hopefully) fixed those dependencies, which should make it "backward compatible" again.

rconweb/api/views.py
```diff
@@ -82,25 +82,39 @@ def get_version(request):
 
 @csrf_exempt
 def public_info(request):
-    status = ctl.get_status()
+    gamestate = ctl.get_gamestate()
+    curr_players, max_players = tuple(map(int, ctl.get_slots().split("/")))
```
cemathey, line 86: We're pulling the total number of players on the server out of `get_slots` here but we also have that information in `get_gamestate`.

I think there's a chance that sometimes we could return inconsistent information since we're getting the server population through two different calls that might return different numbers if users have connected/disconnected between the calls?

Maybe just add together the total number of allied and axis players that we're getting through `get_gamestate`?

`get_slots` is on a 20 second cache timer also, right now `get_gamestate` isn't cached.

zMoooooritz: This is indeed a good note and the cache timing may lead to inconsistencies.
However the number calculated from `get_gamestate` won't be accurate either since it does not include 'unassigned' players (players that have connected but not picked a team yet).

cemathey: Dang I can't believe I didn't think of that. Excellent point. Maybe we should have `get_gamestate` invalidate the cache for `get_slots`?

zMoooooritz: Mhh this would add an obscure relationship between those two functions, which is not desirable.
Maybe we can just reduce the ttl *thinking* 5 seconds, and ignore the remaining edge cases. Since those will always be present at one point or another

rconweb/api/views.py
MarechJ: I'm not sure if it is or not. If it's just extra keys it probably is fine

zMoooooritz: No the 'signature' of the dict had to change a bit. For example the inclusion of the `human_name` for the `next_map` (to have a format analog to the `current_map` format) does mean the 'internal' name of `next_map` is not anymore reachable via `next_map` but instead via `next_map.name`. But I did search for all occurrences of `public_info` and adapted the code accordingly.

MarechJ:
LGTM you can merge

cemathey:
Builds for me on a fresh pull and aside from the server (obviously) failing on get_gamestate I don't think it breaks anything where else in RCON.

PR #151, opened 2022-11-16
Catch and log failed ban removals

### LINKED ISSUE #150: Orphaned RCON Temporary Bans Preventing Removal of Permanent Bans/Blacklist
The game server appears to retain expired temporary bans under certain (unknown) conditions. Removing these temporary bans fail (not a Community RCON bug, it fails with the official RCON client as well).

[ServerCTL.do_unban()](https://github.com/MarechJ/hll_rcon_tool/blob/master/rcon/extended_commands.py#L536) first removes a temporary ban and then removes a permanent ban from a player, however if the player has one of these expired and irremovable temporary bans it is not possible to remove the permanent ban through CRCON.

The permanent ban can be removed through direct CRCON API use, through the official RCON client or (presumably) through the Battlemetrics RCON but I have not attempted that.

I think this is an easy fix and I will knock out a quick patch that will allow permanent bans to be removed correctly when these expired temporary bans exist.

I am also filing a bug with the developers for the lingering temporary ban bug.

Unfortunately since this appears to be a bug with the game server the `/player/<steam ID>` view will incorrectly report temporary bans, I think we could probably work around that by parsing the ban string and explicitly filtering out expired temporary bans.  However the timestamp saved on the game server (the `banlist` string required to unban a player) appears to use the local time zone of the game server itself and this seems impractical to be able to deduce and requiring CRCON users to manually set it also seems impractical.

We could likely deduce the game servers time zone by temporarily banning a fake player, inquiring and then unbanning said player, but it is unknown if the game server is using a simple UTC offset, or if it's using an actual time zone aware timestamp [dates and times are really complicated](https://codeblog.jonskeet.uk/2015/05/05/common-mistakes-in-datetime-formatting-and-parsing/) so I don't think that is practical.

cemathey:
Only lines 590-598 are relevant, the rest is formatting from `black`.

Resolves #150 by catching and logging failed ban removals.

This builds successfully on a fresh repository and I tested this successfully with a user that currently has one of these expired temporary bans on our game server.


FlorianSW:
This will catch all errors unconditionally, right? Means: If there is a legitimate error when removing a temp ban, this will silently be ignored to the user (except the log line)?
Might not be a bit issue, we simply remove the perma ban then, and the user still sees the temp ban being applied in the UI, though (?).

FlorianSW (approved):

MarechJ:
I don't understand why we need that

MarechJ (approved):

cemathey:
Okay round two, this will no longer silently swallow exceptions caused by failed ban removals.

It will now return the reason(s) for failed bans which are then checked for in `rconweb.api.views` which will return them in the response to the API request and display them in the frontend UI like other errors.

Unfortunately when the command fails because of a bugged temp ban the UI on the `History > Players` view ends up in an inconsistent state, the permanent ban and/or blacklist is removed but the UI doesn't refresh, I don't think there's an easy way to change that unless we were to totally rework the error display logic.

I don't think that's worth it for something that is (hopefully) a fairly rare problem, I was able to find another user with a bugged temporary ban (our game server had been reset since I originally found this issue) and used them to successfully test these changes.

Builds on a fresh repo clone without issue and I tested this against a real game server with a real user with a bugged temporary ban.

PR #156, opened 2022-11-25
Format Python/Javascript

cemathey:
Adds a nox file so the entire project can be formatted with `black` and imports sorted with `isort` in one line: `nox` from the root directory

Adds prettier so the entire project can be formatted in one line: `npx prettier --write src/` from `./rcongui/`

Nothing should break since it's formatting changes only, I did do a build and connect it to our event server, seems to work fine.

I think the only potential for breaking anything is the inclusion of new packages with `pip` but I specifically fixed `black` at the highest version we could that doesn't cause `click` to upgrade past our dependency on `Click==7.1.2`


MarechJ (approved):
tl;dr

PR #161, opened 2022-11-30
Feature/u13 updates

cemathey:
Adds the message history sent to a player to the detailed player view.

Adds a parameter to `do_message_player` that defaults to messages not being saved in the database to avoid spam from any service that messages players automatically.

Updated the collapse card to add a parameter to allow controlling whether the card starts open or closed.


rcongui/src/components/PlayerInfo/PlayerInfo.js
```diff
@@ -358,15 +368,23 @@ const PlayerInfoFunc = ({ classes }) => {
                 </Grid>
               </Grid>
             </Grid>
-            <Grid item xl={3} lg={3} md={3} sm={4} xs={12}>
+            <Grid item xl={3} lg={3} md={3} sm={4} xs={12} container>
```
MarechJ, line 371: item AND container, that's not correct, it's either or. If you need a container you should remove the sizing parameters also

rcongui/src/components/collapseCard.js
```diff
@@ -1,40 +1,54 @@
...
-          })}
-          onClick={handleExpandClick}
-          aria-expanded={expanded}
-          aria-label="show more"
-        >
-          { expanded ? <ExpandLessIcon/> : <ExpandMoreIcon /> }
-        </IconButton>
-      } />
+  React.useEffect(() => {
+    setExpanded(startOpen);
```
MarechJ, line 27: setting state from a prop could lead to weird bugs I think, i remember that the react docs are talking about that

rconweb/api/history.py
```diff
@@ -175,6 +179,28 @@ def players_history(request):
     )
 
 
+@csrf_exempt
+@login_required
+def get_player_message(request):
```
MarechJ, line 184: should be plural no? messages

MarechJ (approved):

PR #163, opened 2022-11-30
Re-use add VIP dialog for settings page add vip

### LINKED ISSUE #162: VIP-Management: Not able to change name and VIP to expire never
Good day all,
thank you for this amazing piece of software.

Since the last update, we've got the following issue with VIP-management:
We usually give users VIP through the settings-panel to be able to manually input the name of the player. We need to manually change the name in the list to document the origin of the player (like Clan, Group etc.), so we do not use the History -> Player GUI. Since the update (which introduced expiring VIPs), it is not possible to set the expiry to never in the settings menu. If I give a Player VIP through History -> Player, i can set the expiry to never, but I can't change the nickname.

Can you please make it possible to set the expiry to never and change the name?
If possible, it would be cool and user-friendly to be able to do both in both GUIs (History -> Player and Settings).

Thank you!

LordofAgents

TLDR: Give VIP through History > Player, not able to change name. Give VIP through Settings, not able to set Expiry to never.

FlorianSW:
Fixes #162


cemathey:
This looks good to me, it builds successfully.

VIP uploading/downloading works.

Works for both players that do and don't have a `PlayerSteamID` record.

rcongui/src/components/SettingsView/vips.js
```diff
@@ -182,9 +173,7 @@ const VipEditableList = ({
 }) => {
   const [name, setName] = React.useState("");
   const [steamID64, setSteamID64] = React.useState("");
-  const [expirationTimestamp, setExpirationTimestamp] = React.useState(
-    moment().add(30, "days").format()
-  );
+  const [VIPPlayer, setVIPPlayer] = React.useState(false);
```
MarechJ, line 176: I would have set an empty object or undefined as the default here. Using false as the default made me though this variable was meant to be a boolean flag

MarechJ (approved):

rcongui/src/components/SettingsView/vips.js
FlorianSW: I thought about this when I did the changes as well. However, I then took what was done in the Playerhistory for now. It seems I totally forgot revisiting this part and re-considering is a boolean is really a good idea.

PR #164, opened 2022-12-02
Implement adaptive theme for auto-settings editor

zMoooooritz:
image
image

### DIFF
--- a/rcongui/src/App.js
+++ b/rcongui/src/App.js
@@ -152,7 +152,7 @@ const RedTheme = createMuiTheme({
   },
 });
 
-const GreyBlueTheme = createMuiTheme({
+const GreyBlueDarkTheme = createMuiTheme({
   palette: {
     primary: {
       light: "#8eacbb",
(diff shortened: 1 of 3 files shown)


rcongui/src/App.js
```diff
@@ -152,7 +152,7 @@ const RedTheme = createMuiTheme({
   },
 });
 
-const GreyBlueTheme = createMuiTheme({
```
FlorianSW, line 155: Wouldn't it be possible to have a property in the theme that specifies which editor style should be used? I would feel much better with such an explicit approach (even given it is a bit more verbose in all themes), instead of relying on that a theme which should use a dark editor theme has the string "dark" in the name :)

zMoooooritz: On the one hand yes although with my limited understanding of React this would require some extra code and passing down an additional parameter.
On the other hand it would be good to explicitly name dark themes as such. I thought `GreyBlue` would act like `Red` and `PurplePink` and not be a dark-theme equivalent with a slightly changed navbar-color.
Does anyone else have thoughts concerning this?

FlorianSW: isn't the theme available as a property to a component when wrapped with the `withTheme` helper?
https://v4.mui.com/styles/api/#withtheme-component-component

zMoooooritz: Nice, this is indeed a neat feature, as said absolute novice in React ^^.
Adding extra `editor` field to the `MUITheme` which is then passed to the monaco-editor. Using `vs` (normal mode) as fallback if nothing is specified.
In case monaco supports this in the future we could include custom themes this way as well.

FlorianSW: Yeah, sounds awesome! 👍

FlorianSW (approved):
Looks awesome, thanks!  =)

PR #175, opened 2022-12-09
Clean up and improve automod

zMoooooritz:
Does build successfully and looked as if it is still running as it is supposed to.

If anyone has the server and time capacity to test it properly I would be thankful

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -116,35 +116,32 @@ GTX:
 
 
 NOLEADER_AUTO_MOD:
-  # Set to true to enable the auto moderation of squads without officers
-  # This auto-moderator has 3 steps, 1. warning via broadcast (N times), 2. Punish (N times) 3. kick
+  # This auto-moderator has 3 steps, 1. warning via direct message (X times), 2. punish (Y times) 3. kick
   # Note that the state cycle for a given squad only resets once it finally has an officer. So for example if you
   # disabled the kick and have 3 punishes configured, once the 3 punishes are 
   # through that's it, nothing will happen anymore for that squad even if they still dont have an officer. 
   # 
   # Note 2: punishes and kicks are tracked individual per player, not at the squad level.
   # So if a player joins a squad that already received 3 punishes and will be kicked next, the new player will 
-  # start at the first punish regardless (so he won't be kick), howver he won't have a warning broadcast message as those are issued and tracked per squad)
-  enabled: false
+  # start at the first punish regardless (so he won't be kick)
+  # Set to true to enable the auto moderation of squads without officers
+  enabled: true
   # If this is set to true no warning / punish / kick will be applied for real 
   # It turns the code into a "simulation" mode and only send what it would do to your discord audit log webhook
   dry_run: true
   # Discord Webhook URL where audit logs should be sent. If not specified, the envrionment variable DISCORD_WEBHOOK_AUDIT_LOG
   # will be used, instead.
   discord_webhook_url: ''
-  # The first step of the auto-moderation is to send a broadcast warning the squads that don't have an officer
-  # set to 0 to disable (this it will move to punish directly), -1 for infinite warnings (will never go to punishes)
+  # The first step of the auto-moderation is to send a direct warning message to the squads that don't have an officer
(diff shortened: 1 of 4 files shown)


config/default_config.yml
```diff
@@ -116,35 +116,32 @@ GTX:
...
   # disabled the kick and have 3 punishes configured, once the 3 punishes are 
   # through that's it, nothing will happen anymore for that squad even if they still dont have an officer. 
   # 
   # Note 2: punishes and kicks are tracked individual per player, not at the squad level.
   # So if a player joins a squad that already received 3 punishes and will be kicked next, the new player will 
-  # start at the first punish regardless (so he won't be kick), howver he won't have a warning broadcast message as those are issued and tracked per squad)
-  enabled: false
+  # start at the first punish regardless (so he won't be kick)
+  # Set to true to enable the auto moderation of squads without officers
+  enabled: true
```
MarechJ, line 128: hmm I'd not change that default just yet

rcon/squad_automod/automod.py
```diff
@@ -63,154 +65,151 @@ def is_time(times: List[datetime], interval_seconds: int):
...
+    aplayer: APlayer
 ):
+    warnings = watch_status.warned.setdefault(aplayer.name, [])
+
+    if len(warnings) == 0:
+        logger.debug("Noted but won't warn: %s in %s", aplayer.short_repr(), squad_name)
+        warnings.append(datetime.now())
+        return PunishStepState.WAIT
+
     if config.number_of_warning == 0:
```
MarechJ, line 87: You should move that before the previous block

rcon/squad_automod/automod.py
```diff
@@ -63,154 +65,151 @@ def is_time(times: List[datetime], interval_seconds: int):
...
 
     if len(squad["players"]) < config.min_squad_players_for_punish:
         logger.debug("Squad %s below min player count for punish", squad_name)
-        return PunishStepState.wait
+        return PunishStepState.WAIT
 
     if (
-        int(player["level"]) <= config.immuned_level_up_to
-        or player["role"] in config.immuned_roles
+        aplayer.lvl <= config.immuned_level_up_to
```
MarechJ, line 138: I wouldn't remove the type case here. I'm not sure the playerinfo method is casting it and APlayer doesn't either

rcon/squad_automod/automod.py
```diff
@@ -63,154 +65,151 @@ def is_time(times: List[datetime], interval_seconds: int):
...
 
     if len(squad["players"]) < config.min_squad_players_for_kick:
         logger.debug("Squad %s below min player count for punish", squad_name)
-        return PunishStepState.wait
+        return PunishStepState.WAIT
 
     if (
-        int(player["level"]) <= config.immuned_level_up_to
-        or player["role"] in config.immuned_roles
+        aplayer.lvl <= config.immuned_level_up_to
```
MarechJ, line 194: Same here, typecast for safety

rcon/squad_automod/models.py
```diff
@@ -65,56 +78,31 @@ class NoLeaderConfig:
...
+            f"(name={self.name}, lvl={self.lvl}, role={self.role})"
+        )
 
 
 @dataclass
 class PunitionsToApply:
     warning: List[APlayer] = field(default_factory=list)
     punish: List[APlayer] = field(default_factory=list)
     kick: List[APlayer] = field(default_factory=list)
-    squads_state: List[ASquad] = field(default_factory=list)
```
MarechJ, line 87: I think we should keep that for debug purpose until it's stable

MarechJ (commented):
Do the unittest pass?

config/default_config.yml
zMoooooritz: Yes good catch, residue of testing

rcon/squad_automod/automod.py
zMoooooritz: Did reorder those blocks to have the grace period for all (warn, punish, kick) without implementing it in each method separately.
If there is a more elegant way let me know

zMoooooritz:
> Do the unittest pass?

Yes now they do.
The new solution does add some computational overhead but I think this is pretty clean and nicely configurable

MarechJ (approved):
lgtm you can merge

PR #181, opened 2022-12-13
Hotfix/expiring vip timestamps

cemathey:
* Move thread_pool stuff to `RecordedRcon` to allow threaded VIP bulk uploads
* Fix the expiring VIP feature to handle multiple game server CRCON setups with independent expiration timestamps on each server
* Merge alembic heads again
* Add `get_server_number` to `rcon.utils` to abstract getting the server number since it's used all over the place
* Adds the `ForwardCheckBox` to the VIP dialog boxes so VIP can be forwarded from the `Player > History` view

I set up a CRCON pointed at our event server and main server to test in a multi-server CRCON environment. I can't test with with more servers than that but it should work with any number.

This will leave some orphaned records in peoples databases (missing server numbers) but that should have a very small impact.

Uploading VIPs is now dramatically faster because it's threaded. I diff'd a downloaded file after uploading and they were identical (after sorting, order is not preserved).

I am not very familiar with threading so that might be worth a close look but I tested it and used the rest of the project as reference.


MarechJ (approved):
lgtm

PR #182, opened 2022-12-13
do not overwrite map rotation with vote map when vote map is not enabled

FlorianSW:
On a match start event, vote map is initialised and applied. Applying vote map settings should, however, only be done when the vote map feature is actually enable.d

### DIFF
--- a/rcon/hooks.py
+++ b/rcon/hooks.py
@@ -59,11 +59,11 @@ def count_vote(rcon: RecordedRcon, struct_log: StructuredLogLine):
 
 
 def initialise_vote_map(rcon: RecordedRcon, struct_log):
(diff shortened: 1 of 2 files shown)


timraay (approved):

rcon/hooks.py
```diff
@@ -59,6 +59,10 @@ def count_vote(rcon: RecordedRcon, struct_log: StructuredLogLine):
 
 
 def initialise_vote_map(rcon: RecordedRcon, struct_log):
+    config = VoteMapConfig()
+    if not config.get_vote_enabled():
```
MarechJ, line 63: I'd have not put one there though, we'd still want to generate the suggestion, otherwise if you enable the votemap mid map it will not work right away I think. The one in apply_results is enough

PR #183, opened 2022-12-14
Prevent invalid vote messages when vote map is disabled

cemathey:
Invalid vote messages were appearing for any chat message that started with a number even with map voting disabled.

Prevents this and caches the map vote settings since it is now checked for every chat message.

Fix a spelling mistake.

Built and tested on our event server and it still registers votes, we don't use the map vote feature so I am unfamiliar with it but I do not think I broke it.


cemathey:
Closing this because of branch shenanigans on my end, will re-open from a different branch.

PR #185, opened 2022-12-15
New AutoMod: Seeding rules

FlorianSW:
A simple automod that checks for violations of seeding rules. There is one configurable rule for now, only: Disallowed roles.
The main use case is to prohibit the use of tanks during seeding of a server.


rcon/squad_automod/seeding_rules.py
```diff
@@ -104,10 +107,10 @@ def punitions_to_apply(self, _, squad_name: str, team: str, squad: dict) -> Puni
                 )
 
                 violations = []
-                print(self.config.disallowed_roles.roles)
-                if aplayer.role in self.config.disallowed_roles.roles:
-                    violations.append(self.config.disallowed_roles.message.format(
-                        role=self.config.disallowed_roles.roles.get(aplayer.role)))
+                if self.config.disallowed_roles.min_players < server_player_count < self.config.disallowed_roles.max_players:
```
zMoooooritz, line 110: This is not an inclusive lower bound as described in the config.

zMoooooritz (commented):
Isn't it somewhat confusing to have seeding rule enforcement in the `squad_automod` module?
Should probably rename this module and maybe even split up the logic in different submodules?

FlorianSW:
> Isn't it somewhat confusing to have seeding rule enforcement in the squad_automod module?
> Should probably rename this module.

I thought about the "Squad automod" more about like a module moderating sqauds, so felt the name was okish :P But we can change it ofc, maybe something like just "automods"?

rcon/squad_automod/seeding_rules.py
FlorianSW: Good catch! I'll look into this with the next commits 👍

zMoooooritz:
Nice looks good, just out of interest:
What exactly is the effect of `merge()` in `models.py`? If we for example have a squad which has no leader and does contravene the seeding rules, how do the punishes stack (if they do)?
What are the respective messages that are displayed?

MarechJ (approved):

MarechJ (approved):

FlorianSW:
Let's not merge this right now, I would at least give it a bit of time to test on a real-life server. Wasn't able to to that during the christmas days 🙈

FlorianSW:
Sorry, saw this comment just, yet 🙈 

> Nice looks good, just out of interest: What exactly is the effect of `merge()` in `models.py`? If we for example have a squad which has no leader and does contravene the seeding rules, how do the punishes stack (if they do)? What are the respective messages that are displayed?

Depends on :P There are tests which may explain it a bit:
https://github.com/MarechJ/hll_rcon_tool/pull/185/files#diff-0f0d85d47ee3a35df3138204ddef9c0c683c2f14e8ee796dc5ed623f8f087906

For warnings, e.g., if multiple automods want to warn the same player, the player will get two separate messages (might not be ideal, but I'm not sure what better option we have right now :/). For punishes and kicks, only the _first_ one will apply. Means: if the squad has no leader and a seeding rule is broken, then only the no sl automod will kick (while the no seeding automod will still _count_ the punish). Means: Even if the No SL issue is fixed, then the seeding rule punish will still continue (or go straight to kick, considering, that the warning message is ignored).

rcon/config.py
```diff
@@ -21,7 +21,7 @@ def get_config():
         user_config_path = Path(root) / Path("config.yml")
     user_config = {}
     try:
-        with default_config_path.open() as f:
+        with default_config_path.open(encoding='utf8') as f:
```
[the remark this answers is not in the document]
FlorianSW: Not on windows :P (hate me for that statement 🤣). On windows it seems to be their shitty cp1252 (https://docs.python.org/3/library/functions.html#open).

We _could_ set the env `PYTHONUTF8=1` in the docker files, but for the dev environment it does not work really. As we always expect UTF-8, I think this is the better solution :)

FlorianSW:
Found some minor things and edge cases. Will run the seeding rules automod for one or two more days on our main server now and see how it performs when server is actually seeding (tested that on our second server only, which was empty except me).

zMoooooritz:
> For warnings, e.g., if multiple automods want to warn the same player, the player will get two separate messages (might not be ideal, but I'm not sure what better option we have right now :/). For punishes and kicks, only the _first_ one will apply. Means: if the squad has no leader and a seeding rule is broken, then only the no sl automod will kick (while the no seeding automod will still _count_ the punish). Means: Even if the No SL issue is fixed, then the seeding rule punish will still continue (or go straight to kick, considering, that the warning message is ignored).

Ok nice, this sounds like a reasonable approach. If we add more automods this may lead to an excessive amount of warning messages but should be fine for now.

FlorianSW:
> > For warnings, e.g., if multiple automods want to warn the same player, the player will get two separate messages (might not be ideal, but I'm not sure what better option we have right now :/). For punishes and kicks, only the _first_ one will apply. Means: if the squad has no leader and a seeding rule is broken, then only the no sl automod will kick (while the no seeding automod will still _count_ the punish). Means: Even if the No SL issue is fixed, then the seeding rule punish will still continue (or go straight to kick, considering, that the warning message is ignored).
> 
> Ok nice, this sounds like a reasonable approach. If we add more automods this may lead to an excessive amount of warning messages but should be fine for now.

Yeah, I was also thinking about a possible solution for that, but didn't come up with one, yet. And as there are "only" two possible automods so far, I left this issue to solve for future-Florian 🤣

FlorianSW:
Ok, it was running for quite some time now and did a reasonable good job. I think this is ready to be merged and put into the next release, what do you guys think?

We should then think about other seeding rules that we could potentially check automatically in this mod to work on afterwards.

PR #187, opened 2022-12-16
Expose thread_pool size as config option and decrease CRCON startup time

cemathey:
* Add `ADVANCED_CRCON_SETTINGS` as a `default_config.yml` option and add a thread pool size for users to set their pool size as desired
* Skip automatically connecting to the game server in `rcon.cli` since `inspect.getmembers` causes the property to be initialized even if we skip it while exposing endpoints as CLI commands, so CRCON starts faster

Built the images after catching back up with the `master` branch and connected it to our server and everything appears to work correctly. You weren't kidding about `pandas` taking a long time to build from scratch (=

I think we're okay to just use `get_config` since it's on a 10 minute cache so I don't think it's slow to use it to get the thread pool size.

If we can't modify `.env` (understandable) or tweak any of the compose files I think this is the best way to do it.


config/default_config.yml
```diff
@@ -11,6 +11,11 @@
 ######################################################################################
 ######################################################################################
 
+# Advanced settings, you should only touch these if you really know what you're doing
+ADVANCED_CRCON_SETTINGS:
+  # The number of concurrent connections to the game server CRCON will open
```
MarechJ, line 16: Bit misleading maybe, as it's the number of connection per process, so the total is much higher than that.
Maybe also add that the sweet spot is 10-20 and that below 4 we don't recommend using the automod as there might be significant delays in the Game view.

rcon/extended_commands.py
```diff
@@ -117,9 +117,12 @@ class Rcon(ServerCtl):
     MAX_SERV_NAME_LEN = 1024  # I totally made up that number. Unable to test
     log_time_regexp = re.compile(r".*\((\d+)\).*")
 
-    def __init__(self, *args, pool_size=10, **kwargs):
+    def __init__(self, *args, pool_size=None, **kwargs):
         super().__init__(*args, **kwargs)
-        self.pool_size = pool_size
+        if pool_size:
```
MarechJ, line 122: I'd test for None explicitly.

rcon/commands.py
```diff
@@ -77,11 +79,17 @@ class ServerCtl:
...
+        # .env fed config from rcon.SERVER_INFO
         self.config = config
         self.conn = None
         # self._connect()
         self.auto_retry = auto_retry
 
+        # config/default_config.yml config.yml, etc.
+        config = get_config()
+        # logger.debug(f"{config=}")
+        self.advanced_settings = AdvancedConfigOptions(**config['ADVANCED_CRCON_SETTINGS'])
```
MarechJ, line 91: Since this layer doesn't use it I'd have put that in the init of the extended one. Now this class can't be used with a config it's a bit of a shame.

rcon/models.py
```diff
@@ -678,3 +678,9 @@ def valid_role(cls, values):
                 raise ValueError(f"Invalid Discord role {role_or_user}")
 
         return values
+
+
+class AdvancedConfigOptions(pydantic.BaseModel):
+    """ADVANCED_CRCON_SETTINGS in config.yml"""
+
+    thread_pool_size: pydantic.conint(ge=1, le=100)
```
MarechJ, line 686: I'd catch and log the error that this is throwing in the init where you initialise that config.

MarechJ (changes requested):

cemathey:
I believe I addressed all of the comments above.

Includes some miscellaneous changes as well:
* Updated .gitignore to ignore auto generated Django html/css/etc., the `certs/` folder and `supervisor_data/` for scorebot because it was all cluttering up my VS Code source control view and annoying me.
* `isort`ed and `black`ed the files I touched and removed some unused imports
* Moved some types out of `rcon.extended_commands` into `rcon.types` since that exists now
* Moved `get_team_view_fast` out of `rcon.extended_commands` into `rcon.recorded_commands` because it won't work from there with the thread pool stuff being in `RecordedRcon`, and we can't move it back without threading bulk VIP uploads in some other fashion (which is an actual issue for people with large VIP lists because they can time out)
* The auto exposed API uses `RecordedRcon` so this shouldn't hurt anything, and squad auto mod uses it as well
* Updated `rcon.cli` to use `RecordedRcon` because of the above change
* Prevents `rcon.steam_utils.get_players_ban` from logging warnings if there are no players on the server and a steam API key is not set

I built the images and hooked it up to our server and everything seems to be working fine.

MarechJ:
lgtm

MarechJ (approved):

PR #188, opened 2022-12-17
Hotfix/django migrations

cemathey:
Removes the unused `rconweb/entrypoint.sh` and adds creation of Django migrations to the root `entrypoint.sh`

The `api_rconuser` relation which correlates to `RconUser` in `rconweb.api.models` migration was not being generated which was blocking the deletion of user accounts from the Admin panel.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -14,6 +14,7 @@ then
   ./manage.py init_db
   ./manage.py register_api
   cd rconweb 
+  ./manage.py makemigrations --no-input
   ./manage.py migrate --noinput
   ./manage.py collectstatic --noinput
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -14,6 +14,7 @@ then
   ./manage.py init_db
   ./manage.py register_api
   cd rconweb 
+  ./manage.py makemigrations --no-input
```
FlorianSW, line 17: I might be wrong here, as I didn't work with that that often in the past: But isn't the database schema handled by alembic? We have an upgrade statement in L13 already?

FlorianSW (approved):

PR #192, opened 2022-12-21
Implement whitelist logic for votemap via autosettings

zMoooooritz:
This allows to restrict the available maps for votemap via the autosettings.

The logic for storing and accessing the whitelist has been tested, but I am currently unable to test the whole package against a server. Although I think it should be fine™

This is only a "backend" implementation and does not contain a rotation manager like GUI to modify the whitelist. I am just not fluent enough in React to implement this in a reasonable time. Not even sure if it would be used by anyone.

Furthermore the code does not perform tests in order to avoid an empty selection.
We could maybe fallback on `pick_default_next_map` in case this happens (and ignore the whitelist as is currently)?

### DIFF
--- a/rcon/auto_settings.py
+++ b/rcon/auto_settings.py
@@ -7,6 +7,7 @@ import pytz
 
 from rcon.audit import ingame_mods, online_mods
 from rcon.extended_commands import Rcon
+from rcon.vote_map import VoteMap
 from rcon.settings import SERVER_INFO
 from rcon.user_config import AutoSettingsConfig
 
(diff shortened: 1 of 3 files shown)


rcon/auto_settings.py
```diff
@@ -181,18 +182,30 @@ def create_condition(name, **kwargs):
...
         try:
             logger.info("Applying %s %s", command, params)
-            rcon.__getattribute__(command)(**params)
+            vm_commands = {
+                "do_add_map_to_whitelist": votemap.do_add_map_to_whitelist,
+                "do_add_maps_to_whitelist": votemap.do_add_maps_to_whitelist,
+                "do_remove_map_from_whitelist": votemap.do_remove_map_from_whitelist,
+                "do_remove_maps_from_whitelist": votemap.do_remove_maps_from_whitelist,
+                "do_reset_map_whitelist": votemap.do_reset_map_whitelist,
+                "do_set_map_whitelist": votemap.do_set_map_whitelist,
```
MarechJ, line 195: Oh wow you went all the way. 
I was really just thinking of a list in the config.yml and we call it a day 😅

MarechJ: Now we have to add a UI

MarechJ: If we have set_whitelist(maos) I can just reuse the UI from the maprotation I think

MarechJ (approved):

PR #196, opened 2022-12-22
AutoSettings: Allow invocation of multiple rules

zMoooooritz:


config/auto_settings.default.json
```diff
@@ -1,5 +1,6 @@
 {
   "always_apply_defaults": false,
+  "can_invoke_multiple_rules": false,
```
MarechJ, line 3: Can we add an explanation somewhere / somehow?

MarechJ (approved):

PR #197, opened 2022-12-23
Fix commander bug in team view

zMoooooritz:
image
GameView of this branch (left side) vs prod (right side), seems to be working fine


MarechJ (approved):
Fine to merge if you're sure this is compatible with automod and game view

zMoooooritz:
yes, this change does only alter the representation that is returned by `get_detailed_player_info` (for the commander).
The `get_game_view(_fast)` does insert the commander and eventual "No Unit" squad at the respective positions.

The current iteration of automod does only look at the "squads" in the `team_view` which does not include the commander and ignores the "empty"/"No Unit" squad

PR #203, opened 2023-01-07
Fix rcon mod wrongfully considering kill as "during seeding"

FlorianSW:
In rare cases, the rcon tool might not yet recognized that the server reached enough population to disable seeding rules (something around 2-7 seconds, depending on how long it takes for the first call to get_team_view).

Kills during this time are handled by an event, while the seeding automod assumes the server did not reach the seeding threshold yet. This might result in kills being punished as they were made with a seeding-disallowed-weapon, despite the server having enough players.

This commit fixes this by giving automods generally a bit of time after restaret of the rcon tool to warmup. This might defer or miss enforcements on some kills, but at least no-one is punished wrongfully.

### DIFF
--- a/rcon/automods/automod.py
+++ b/rcon/automods/automod.py
@@ -119,6 +119,7 @@ def punish_squads(rcon: RecordedRcon):
     punitions_to_apply = get_punitions_to_apply(rcon, mods)
 
     do_punitions(rcon, punitions_to_apply)
+    global first_run_done
     first_run_done = True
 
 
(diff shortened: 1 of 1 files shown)


rcon/automods/automod.py
```diff
@@ -118,6 +119,7 @@ def punish_squads(rcon: RecordedRcon):
     punitions_to_apply = get_punitions_to_apply(rcon, mods)
 
     do_punitions(rcon, punitions_to_apply)
+    first_run_done = True
```
MarechJ, line 122: might need to add 

```
global first_run_done
```
above that line

MarechJ (approved):

PR #205, opened 2023-01-13
Do not record duplicated player sessions

FlorianSW:
Hooks might be called twice (when the log event loop crashes or gets restarted for another reason while processing hooks). There should never be a reason to save the same player session twice, anyway.

### DIFF
--- a/rcon/player_history.py
+++ b/rcon/player_history.py
@@ -352,10 +352,22 @@ def save_start_player_session(
             )
             return
 
+        start_time = datetime.datetime.fromtimestamp(timestamp)
+        already_saved = (
+            sess.query(PlayerSession)
+            .filter(PlayerSession.steamid == player)
+            .filter(PlayerSession.start == start_time)
+            .first()
+        )
+
+        if already_saved is not None:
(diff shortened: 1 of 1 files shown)


rcon/player_history.py
```diff
@@ -352,10 +352,22 @@ def save_start_player_session(
             )
             return
 
+        start_time = datetime.datetime.fromtimestamp(timestamp)
+        already_saved = (
+            sess.query(PlayerSession)
+            .filter(PlayerSession.steamid == player)
+            .filter(PlayerSession.start == start_time)
+            .first()
```
MarechJ, line 360: first() doesn't raise if there's no result? I tend to use `one_or_none()`

MarechJ (approved):

rcon/player_history.py
FlorianSW: Docs say it returns none:
https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.first
> Return the first result of this Query or None if the result doesn’t contain any row.

PR #207, opened 2023-01-21
Reduce memory footprint of backend container

FlorianSW:


rcon/commands.py
```diff
@@ -73,78 +89,119 @@ class ServerCtl:
...
+    @contextmanager
+    def with_connection(self) -> HLLConnection:
+        logger.debug("Waiting to acquire lock")
+        with self.mu:
+            if len(self.idles) != 0:
+                logger.debug("acquiring connection from idle pool")
+                conn = self.idles.pop()
+
+            elif self.numOpen >= self.maxOpen:
+                logger.debug("Max connections already open, waiting for connection returned to pool")
```
MarechJ, line 111: so the pool size should always be equal or higher than the number of threads used by gunicor so this doesn't happen

rcon/commands.py
```diff
@@ -73,78 +89,119 @@ class ServerCtl:
...
+                c = 0
+                while len(self.idles) == 0:
+                    if c >= 30:
+                        logger.error("waiting for connection returned to pool timed out after " + str(c) + " seconds")
```
MarechJ, line 115: Best to use the string format builtin in the logger.

` logger.error("waiting for connection returned to pool timed out after %s seconds", c)`

rcon/commands.py
```diff
@@ -73,78 +89,119 @@ class ServerCtl:
...
+            yield conn
+        finally:
+            logger.debug("return connection")
+            if len(self.idles) >= self.maxIdle:
+                logger.debug("Enough connections in pool, closing")
+                self.numOpen -= 1
+                conn.close()
+            else:
+                logger.debug("Returning connection to pool")
+                self.idles.append(conn)
```
MarechJ, line 137: hmm I'm not sure it's the best move to add the connection in the "finally" cause what if that connection just raised cause it was a bad password, or just somehow it got disconnected. We could shove a broken connection in the pool

rcon/commands.py
```diff
@@ -73,78 +89,119 @@ class ServerCtl:
...
+
+    def _connect(self, conn: HLLConnection):
+        try:
+            conn.connect(self.config["host"], int(self.config["port"]), self.config["password"])
         except (TypeError, ValueError) as e:
-            logger.critical("Invalid connection information")
+            logger.critical("Invalid connection information", e)
```
MarechJ, line 143: you need to add the %s in the format string also. but if you use `logger.exception` the stacktrace should contain the exception name an its content anyways. Dunno why I used cirtical in there, not sure how it behaves actually

rcon/commands.py
```diff
@@ -163,7 +220,8 @@ def _timed_request(self, command: str, can_fail=True, log_info=False):
             result=result,
         )
 
-    def _read_list(self, raw):
+    def _read_list(self, raw, conn: HLLConnection):
+        assert conn is not None
```
MarechJ, line 224: We don't run python with the optimize flag but if we would then that assert statement would be removed. So basically for production code assert should never be used as a mean to raise an exception outside of tests or debug code.

rcon/commands.py
```diff
@@ -209,13 +266,14 @@ def _read_list(self, raw):
         return [l.decode() for l in res[1:]]
 
     @_auto_retry
-    def _get(self, item, is_list=False, can_fail=True):
-        res = self._request(f"get {item}", can_fail, decode=not is_list)
+    def _get(self, item, is_list=False, can_fail=True, conn: HLLConnection = None):
+        assert conn is not None
```
MarechJ, line 270: same here.

rcon/commands.py
```diff
@@ -290,8 +348,22 @@ def get_map_rotation(self):
...
+                try:
+                    steam_id_64, name = item.split(" ", 1)
+                    name = name.replace('"', "")
+                    name = name.replace("\n", "")
+                    name = name.strip()
+                    vip_ids.append(dict(steam_id_64=steam_id_64, name=name))
+                except ValueError:
+                    self._reconnect(conn)
+                    raise
+            return vip_ids
```
MarechJ, line 366: Why the big change here? Isn't the extended_commands.Rcon doing that already?

rcon/commands.py
```diff
@@ -300,37 +372,39 @@ def get_autobalance_enabled(self):
         return self._get("autobalanceenabled", can_fail=False)
 
     @_auto_retry
-    def get_logs(self, since_min_ago, filter_=""):
+    def get_logs(self, since_min_ago, filter_="", conn: HLLConnection = None):
+        assert conn is not None
```
MarechJ, line 376: .

rcon/commands.py
```diff
@@ -300,37 +372,39 @@ def get_autobalance_enabled(self):
...
-                break
-            res["result"] += self.conn.receive().decode()
-        return res
+        with self.with_connection() as conn:
+            res = self._timed_request(f"showlog {since_min_ago}", conn=conn)
+            for i in range(30):
+                if res["result"][-1] == "\n":
+                    break
+                res["result"] += conn.receive().decode()
+            return res
```
MarechJ, line 407: I don't think this method is used but it's not locking / unlocking its connection

MarechJ (approved):
Some comments but I think it's viable. thanks for the contrib!

rcon/commands.py
```diff
@@ -73,78 +89,119 @@ class ServerCtl:
     set password not implemented on purpose
     """
 
-    def __init__(self, config, auto_retry=1):
+    def __init__(self, config, auto_retry=1, pool_size=20):
```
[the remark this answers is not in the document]
FlorianSW: It should do so already when RecordedRcon is used, but might be worth moving this logic out there and into here 👍

rcon/commands.py
```diff
@@ -73,78 +89,119 @@ class ServerCtl:
...
         # .env fed config from rcon.SERVER_INFO
         self.config = config
-        self.conn = None
-        # self._connect()
         self.auto_retry = auto_retry
+        self.mu = threading.RLock()
+        self.idles: list[HLLConnection] = []
+        self.numOpen = 0
+        self.maxOpen = 20
```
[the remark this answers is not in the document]
FlorianSW: lol, it actually should 🤣 (it's even the more important one than maxIdle 🙈)

rcon/commands.py
FlorianSW: Not quite, I mean, there might be _some_ overhead on certain actions that require the RCon actions to be performed. In this case, the rcon connection would be returned to the pool much sooner than the gunicorn process is done.

One thing were this would not be true would be the team view: It will utilize as much connections as possible (up to the number of threads in the RecordedRcon ThreadPool), while it is still just one gunicorn process. In this case, the pool _might_ hit the max open connections limit and new actions that wait for a connection will be kept on hold once a connection is available.

Maybe we can reduce the wait time from 1 second to some lower value. But I assume that people will not see this very often based on the default pool size as well as the caching going around for most of the data.

rcon/commands.py
FlorianSW: I was thinking about his as well.

For the wrong password case, this should already be handled when the connection is setup in the _connect method. So this error should raise all the way up to the caller.
For the errors that can happen when an action si executed: Most of the commands have a handling of that already (and either call reconnect or do some other handling). I'm not sure, if we should change that in this commit, tbh. Right now, if a connection is broken, we keep using it in the same thread as well, with the same consequences.
Overhauling error handling of such basic RCon errors should probably be done in a separate PR? Or what do you think? :)

rcon/commands.py
FlorianSW: lol, you might be right, didn't really look into that :D I'll might simply change that to logger.exception 😅

rcon/commands.py
FlorianSW: good hint, didn't know that for python 🙈 I'll check other ways on how to do that easily :)

rcon/commands.py
FlorianSW: it did, however, it required the connection being exposed to handle the reconnect when reading the result of the vipids command. With this code being in extended_commands (while ServerCtl actually does the reequest to the server) there was not context for what connection was used to reconnect it. To keep the same functionality, I moved reading the list into here (assuming that no-one needs the raw list anyways, at least not in the current version of the code)

rcon/commands.py
FlorianSW: It did not do that before, either 🤔 Locking/Unlocking on the connection _itself_ should anyway be pretty useless here. The pool ensures that a connection is not used for another RCon command before the current user of the connection finishes somehow.

rcon/commands.py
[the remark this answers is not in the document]
MarechJ: ah ok. Not all the code uses the RecordedVersion, if you do that in the extended one then it's safe though

PR #208, opened 2023-01-21
[supervisord] Merge routine services into one program

FlorianSW:
These services are run frequently in a scheduled fashion, but are mostly bound to I/O. They do not necessarily need to run in their own, separate environment (process) but can run in an async fashion. This saves memory consumption for their own, separate python runtime.


rcon/routines.py
```diff
@@ -56,21 +41,56 @@ def toggle_votekick(rcon: RecordedRcon):
...
+
+async def run():
+    from rcon.settings import SERVER_INFO
+    rcon = RecordedRcon(SERVER_INFO)
+    server_number = os.getenv("SERVER_NUMBER")
+
+    tasks = asyncio.gather(
+        scheduled_routines(rcon),
+        auto_broadcast(component_logger("broadcasts", server_number), rcon),
+        expire_vips(component_logger("expiring_vips", server_number), rcon),
```
MarechJ, line 89: Hmm is that going to work? I'm afraid the broadcast will be quite out of sync especially if ppl use it with 1 second messages. Since those are not really async function, when they do they blocking IO they won't be releasing their time to other function, only when they reach an async sleep. so the whole timing will be highly dependent on how long the preceding function take to run. Could be quite messy no?

FlorianSW: When there is quite some traffic in these things (routines, auto-broadcasts and expire_vips), the execution might be delayed a bit, yes. However, is 1-second delayed messages in auto-broadcast really a thing that we see in the wild or even want to really support? 🤔 
If so, this whole thing would benefit from the RCon connections being truly async, yes.

PR #209, opened 2023-01-23
Bugfix/vips

cemathey:
* Fix unnecessary and spammy gamestate logging.
* Fix the duplicate record insertion for expiring VIPs and allow the service to be run from the command line for debugging purposes.
* Change the call to `get_vip_ids()` in `rcon.recorded_rcon` to reference `super` even though it works to make it more explicit.
* Some minor formatting changes from `black` and some minor tweaks to make pylance happy.
* Fix a bug I introduced in `rcon.steam_utils` because I was misunderstanding the logged messages, my 'fix' was causing errors.


rcon/commands.py
```diff
@@ -492,7 +492,7 @@ def get_gamestate(self) -> List[str]:
         # Has no trailing "\n"
 
         result = self._get("gamestate", can_fail=False)
-        logger.info("Gamestate results:\n|%s|", result)
+        logger.debug("Gamestate results:\n|%s|", result)
```
FlorianSW, line 495: I think this can be removed completely. If I'm not mistaken, this should already be removed in master.

FlorianSW (approved):
I think this looks good. Could you please rebase first to see if my comment is out of date? :)

PR #215, opened 2023-02-02
always release lock, even when connection is not established

FlorianSW:
If connecting to the HLL server for a new connection fails for whatever reason, the lock should also be released. Otherwise, the exception will raise up the call chain, leaving the lock in an unrecoverable state acquired, which prevents any subsequent request to the pool to indefinetly wait for the lock to be available.

Also: Never wait indefinetly for a lock, which will never throw any error in such cases, which makes it especially hard to debug such things with production logs only.


MarechJ (approved):

rcon/commands.py
```diff
@@ -138,7 +138,8 @@ def __init__(self, config, auto_retry=1, max_open=None, max_idle=None):
     @contextmanager
     def with_connection(self) -> HLLConnection:
         logger.debug("Waiting to acquire lock %s", threading.get_ident())
-        self.mu.acquire()
+        if not self.mu.acquire(timeout=30):
```
MarechJ, line 141: You could make it more agressive. If the nb of connection in the pool matches the number of gunicorn threads per worker then a thread should never have to wait to get a connection in theory

rcon/commands.py
```diff
@@ -156,19 +157,26 @@ def with_connection(self) -> HLLConnection:
                     raise TimeoutError()
                 c += 1
                 time.sleep(1)
-                with self.mu:
-                    idle_len = len(self.idles)
 
-            with self.mu:
-                conn = self.idles.pop()
-                logger.debug("connection appeared in pool: %s, acquiring it", conn.id)
+                if not self.mu.acquire(timeout=30):
```
MarechJ, line 161: why not putting this timeout in the context manager directly?

MarechJ (approved):

rcon/commands.py
FlorianSW: That would, however, only work for the backend. The pool is the same for background tasks (like automod, routines, log_event and so on). In these case, the thread count might be different from the number of connections in the pool. And I would love to have the pool Independent from configuration in gunicorn 😅

rcon/commands.py
FlorianSW: I don't know a way of passing a timeout to acquire when using the context manager 🤔 I mean, without writing my own contextmanager for that :D

PR #220, opened 2023-02-05
Remove server stat background jobs

FlorianSW:


rcon/server_stats.py
```diff
@@ -67,14 +67,6 @@ def save_server_stats_for_last_hours(hours=24, skip_last_hours=1):
     return save_server_stats_for_range(start, end)
 
 
-def save_server_stats_since_inception():
```
cemathey, line 70: Why not just leave this one so it doesn't need to get rewritten later? With the CLI disabled and it not being called in `entrypoint.sh` it shouldn't hurt anything.

rcon/cli.py
```diff
@@ -45,11 +45,6 @@ def run_stats_loop():
         sys.exit(1)
 
 
-@cli.command(name="record_server_stats_inception")
```
cemathey, line 48: Maybe just comment this out? This stuff is getting re-enabled at some point I assume.

entrypoint.sh
```diff
@@ -32,8 +32,6 @@ then
     exit 0
 fi
   sleep 10
-  LOGGING_FILENAME=historical_server_stats_$SERVER_NUMBER.log python -m rcon.cli record_server_stats_inception &
```
cemathey, line 35: Maybe just comment this out? This stuff is getting re-enabled at some point I assume.

MarechJ: it's fine to remove this. However why remove the code also?

cemathey (approved):

PR #221, opened 2023-02-05
Hotfix/multi-line log parsing

cemathey:
* Update log parsing to parse multi-line content returned from the game server `showlog` command (thanks Abu)
* Update log parsing to account for vote kick format changes
* Refactor log parsing a bit to make testing easier
* Rewrite `tests/test_logs.py` to reflect log parsing changes
* Flatten multi-line messages sent to the discord audit webhook
* Add some typing/update types to reflect log parsing changes
* Update the regex pattern for kills/teamkills to account for player name exploit (thanks Fachi for pointing it out)
* Move all the regex's to be pre-compiled for performance reasons
* Update `Rcon.extract_time` to explicitly return a UTC `datetime` even though it makes no difference in production, makes testing complicated because otherwise you will get a local `datetime` and tests will fail in different timezones.
* Update `LogRecorder._save_logs` to include the log version (we should now start incrementing this when we make changes) and to include the `weapon` for kills/team kills.
* Sort the actions returned from `rcon.game_logs.get_recent_logs` for QoL improvement on the Live View
* Bump `black` version to the latest since we bumped `click` in a previous commit and bump `isort` version to the latest
* Some miscellaneous format changes from `black`, `isort` and remove some unused imports

I tested this against two separate game servers for awhile and did some `diffs` between my version and our real CRCON and reviewed log files for errors, I believe this is working correctly and maintains the parsed log format.

The only log parsing error it generated was what looked like it came from an incomplete `showlog` returned from the game server.


rcon/extended_commands.py
```diff
@@ -439,6 +467,7 @@ def do_unban(self, steam_id_64) -> List[str]:
         failed_ban_removals: List[str] = []
         for b in bans:
             if b.get("steam_id_64") == steam_id_64:
+                # TODO: This is no longer true as of U13
```
FlorianSW, line 470: did you verify that? :)

FlorianSW (commented):
Did you run the logs test against the new implementation? :) Does it contain all the new additions btw:? Like host connections and stuff? 🤔

MarechJ (commented):
This is hard to review with all the reformatting

tests/test_logs.py
```diff
@@ -1,380 +1,711 @@
-from unittest import mock
+from datetime import datetime
 
-from rcon.extended_commands import Rcon
+import pytest
 
-#  Unable to parse line: [4:48 min (1659278057)] KICK: [|||™ Marcraƒt] has been kicked. [Host closed the connection.]
-# TODO: Unable to parse line: [4:54 min (1659283019)] KICK: [[BC] Gimrrin] has been kicked. [Host closed the connection.]
```
MarechJ, line 6: You removed that todo but is this one handled now?

cemathey: That is correct, not sure how to link it directly in here but line 515 on the tests file we check for the `Host closed the connection.` style of disconnects.

rcon/discord.py
```diff
@@ -45,6 +45,8 @@ def dict_to_discord(d):
 
 def send_to_discord_audit(message, by=None, silent=True, webhookurl=None):
     webhookurl = webhookurl or os.getenv("DISCORD_WEBHOOK_AUDIT_LOG", None)
+    # Flatten messages with newlines
```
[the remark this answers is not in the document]
FlorianSW: Yeah, I mean, that is how the message was written to the player. I would not really want to change that tbh 😅 but that is subjective, I think.

rcon/extended_commands.py
```diff
@@ -80,15 +86,37 @@ class Rcon(ServerCtl):
...
     chat_regexp = re.compile(
-        r"CHAT\[((Team)|(Unit))\]\[(.*)\(((Allies)|(Axis))/(\d+)\)\]: (.*)"
+        r"CHAT\[(Team|Unit)\]\[(.*)\((Allies|Axis)/(\d+)\)\]: (.*)"
     )
     player_info_pattern = r"(.*)\(((Allies)|(Axis))/(\d+)\)"
     player_info_regexp = re.compile(r"(.*)\(((Allies)|(Axis))/(\d+)\)")
-    MAX_SERV_NAME_LEN = 1024  # I totally made up that number. Unable to test
     log_time_regexp = re.compile(r".*\((\d+)\).*")
+    connect_disconnect_pattern = re.compile(r"(.+) \((\d+)\)")
+    # Checking for steam ID length so people can't exploit it with a name like: short(Axis/123) ->
```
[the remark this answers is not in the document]
FlorianSW: ah, got it 👍 That makes sense. Thanks for the explanation ❤️

FlorianSW (approved):
From a brief look through the code and the tests, this looks good. However, another one might want to check it as well.

PR #225, opened 2023-02-16
Update .env documentation and remove deprecated env variables

cemathey:
* Update the `.env` documentation slightly
* Remove deprecated `VOTE_MAP_INSTRUCTIONS` and `SILENT_MAP_RECORDER` options from the `.env`
* Update the compose files to reflect the `.env` changes

I know we try very hard to avoid to avoid changing the `.env` file to make it easier for people to upgrade, but the next release contains a change to support the backend changes @FlorianSW made (Thanks!) so now would not be a bad time to make any other updates we want to make.

I did a build and fired it up and nothing was broken.


.env
```diff
@@ -2,19 +2,27 @@
...
+
 # -----------------------------
 # Database Details
 # -----------------------------
 # This is a database password you have to choose, It does not have to be memorable 
-#as you will probably never access it directly but make it strong.
-HLL_DB_PASSWORD=
+# as you will probably never access it directly but make it strong.
+# Prohibited characters for HLL_DB_PASSWORD: %#
+HLL_DB_PASSWORD=asdf
```
zMoooooritz, line 18: I think we don't want to include "asdf" here as value and keep it empty as is

zMoooooritz (approved):
All the other stuff looks fine except one thing.
The "VAC & Game Bans" section seems totally out of place. Probably a relic of the time when all configuration was done inside this file.
I think it would be good to move this setting section into the `config.yaml`, which should be pretty easy to do?!
Unfortunately I do not have time to do it right now

PR #227, opened 2023-02-21
Do not throw away connection on FAIL

FlorianSW:
When the game server returns a FAIL, this usually does not mean that the connection itself is broken, hence it should not be discarded.

This also fixes the set_map logic that ensures that the map is in the rotation if changing the map fails.


zMoooooritz (commented):
lgtm

zMoooooritz (approved):

PR #228, opened 2023-02-25
Fix announcing seeding rules on connect

FlorianSW:
A global variable set during an automod run is not available when en event from the log event loop is processed. Save this shared state in redis instead.

Also make announcement work when only disallowed roles is configured.

### DIFF
--- a/rcon/automods/automod.py
+++ b/rcon/automods/automod.py
@@ -23,7 +23,8 @@ from rcon.settings import SERVER_INFO
 from rcon.types import StructuredLogLineType
 
 logger = logging.getLogger(__name__)
-first_run_done = False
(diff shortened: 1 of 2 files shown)


cemathey (approved):
lgtm but I'm not very familiar with the automod portion.

Might be preferable to use dependency injection in `is_first_run_done` and `set_first_run_done` to make it easier to mock during testing though.

FlorianSW:
> Might be preferable to use dependency injection in is_first_run_done and set_first_run_done to make it easier to mock during testing though.

There is not much level we can move this up, but I just go ahead and do it as far as possible (the methods are used in class-less function entrypoints anyway).

PR #230, opened 2023-03-03
Fix log timestamp parsing

cemathey:
The Unix epoch timestamp sent from the game server was already milliseconds and we were double converting it which would then break anything that then converted it to a Python datetime because it would exceed the maximum allowed year.

For instance: `[299 ms (1677864556)] TEAMSWITCH Solid Cake (None > Allies` would be stored as `1677864556000` (the year 55,139)  instead of `1677864556`

Bumps the log version.

I did build this and test it against our event server, it is working correctly and the TK auto punish feature now works again.

### DIFF
--- a/rcon/extended_commands.py
+++ b/rcon/extended_commands.py
@@ -1133,7 +1133,6 @@ class Rcon(ServerCtl):
             else:
                 raise ValueError(f"Unable to parse line: {raw_line}")
         elif raw_line.startswith("KICK") or raw_line.startswith("BAN"):
-
             if "FOR TEAM KILLING" in raw_line:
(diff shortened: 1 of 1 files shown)


FlorianSW:
Ok, I don't get it. What the gameserver sends looks pretty much like a "normal" POSIX timestamp (seconds), right? So the value should be multiplied with 1000 when we want to have a timestamp_ms? What do I miss? 😅

cemathey:
I am dumb, the real problem is here https://github.com/MarechJ/hll_rcon_tool/blob/master/rcon/game_logs.py#L540

PR #232, opened 2023-03-03
Print the YAML validation error for easier troubleshooting

cemathey:
Print the actual YAML validation error when loading invalid YAML so users can troubleshoot easier.

For example:

```
[2023-03-03 19:26:43,343][ERROR] rcon.config config.py:get_config:39 | User config at '/config/config.yml' is invalid YAML
[2023-03-03 19:26:43,343][ERROR] rcon.config config.py:get_config:40 | while scanning for the next token
found character '\t' that cannot start any token
  in "/config/config.yml", line 329, column 38
```

### DIFF
--- a/rcon/config.py
+++ b/rcon/config.py
@@ -35,8 +35,9 @@ def get_config():
             user_config = yaml.safe_load(stream=f)
     except FileNotFoundError:
         logger.warning("No user config found, defaults only are loaded")
-    except yaml.YAMLError:
+    except yaml.YAMLError as e:
         logger.error("User config at '%s' is invalid YAML", str(user_config_path))
+        logger.error(e)
 
     config.update(user_config)
     return config


rcon/config.py
```diff
@@ -35,8 +35,9 @@ def get_config():
             user_config = yaml.safe_load(stream=f)
     except FileNotFoundError:
         logger.warning("No user config found, defaults only are loaded")
-    except yaml.YAMLError:
+    except yaml.YAMLError as e:
         logger.error("User config at '%s' is invalid YAML", str(user_config_path))
```
FlorianSW, line 39: when using logger.exception (iirc), the stack trace should already be there :)

cemathey: So previously you would only get error messages like:

```
[2023-03-03 19:26:43,343][ERROR] rcon.config config.py:get_config:39 | User config at '/config/config.yml' is invalid YAML
```

The change makes the extra line show up:

```
[2023-03-03 19:26:43,343][ERROR] rcon.config config.py:get_config:40 | while scanning for the next token
found character '\t' that cannot start any token
  in "/config/config.yml", line 329, column 38
```

Which I would say generally isn't a big deal since YAML validators exist, but in this case I had to troubleshoot like this because online YAML validators (maybe different schema versions or something?) were reporting a users YAML file as valid, but CRCON was saying it was invalid.

Might as well include the reason why Pythons YAML validator thinks a file is invalid, only downside is some extra info in the logs, which as long as people haven't screwed up their log rotation isn't a big deal.

FlorianSW: I'm totally with you here. I just wanted to point out, that we might be able to get rid of these two log statements in the code, when using logger.exception 😅 it's just an idea, we do not need to do so.

FlorianSW (approved):

rcon/config.py
cemathey: Ah yeah I understand now. I think this might be easier for users, stack traces confuse people without Python/programming experience in general and we don't need them to trouble shoot in this case I think since it prints out the validation error.

PR #233, opened 2023-03-05
Enforce Cap Fight during seeding

FlorianSW:
A new seeding rule that, when configured, warns/punishes players who gain offensive points when their team is "winning" (has the amount of caps configured in the config), which is considered "attacking the enemy sector".

### DIFF
--- a/rcon/automods/seeding_rules.py
+++ b/rcon/automods/seeding_rules.py
@@ -2,6 +2,7 @@ import logging
 import pickle
 from contextlib import contextmanager
 from datetime import datetime, timedelta
+from typing import Literal
 
 import redis
 
@@ -21,10 +22,11 @@ from rcon.automods.num_or_inf import num_or_inf
 from rcon.cache_utils import get_redis_client
 from rcon.extended_commands import StructuredLogLineType
 from rcon.game_logs import on_match_start
+from rcon.types import GameState
 
 SEEDING_RULES_RESET_SECS = 120
 AUTOMOD_USERNAME = "SeedingRulesAutomod"
-SEEDING_RULE_NAMES = ["disallowed_roles", "disallowed_weapons"]
+SEEDING_RULE_NAMES = ["disallowed_roles", "disallowed_weapons", "enforce_cap_fight"]
 
 
 def disabled_rule_key(rule: str) -> str:
@@ -82,10 +84,10 @@ class SeedingRulesAutomod:
         disallowed_weapons = set(self.config.disallowed_weapons.weapons.values())
 
         if self.config.announce_seeding_active.enabled and (
-            len(disallowed_roles) != 0 or len(disallowed_weapons) != 0
+                len(disallowed_roles) != 0 or len(disallowed_weapons) != 0
         ):
             if all(
-                [self._is_seeding_rule_disabled(r) for r in SEEDING_RULE_NAMES]
+                    [self._is_seeding_rule_disabled(r) for r in SEEDING_RULE_NAMES]
             ):
                 return p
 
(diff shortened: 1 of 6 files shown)


rcon/automods/seeding_rules.py
```diff
@@ -257,16 +259,40 @@ def punitions_to_apply(
...
+                        and drc.min_players <= server_player_count < drc.max_players
                 ):
                     if aplayer.role in drc.roles:
                         violations.append(
                             drc.message.format(role=drc.roles.get(aplayer.role))
                         )
 
+                if not self._is_seeding_rule_disabled("enforce_cap_fight") and \
+                        (team == "axis" and game_state["axis_score"]) >= ecf.max_caps or \
+                        (team == "allies" and game_state["allied_score"] >= ecf.max_caps):
```
timraay, line 278: Enforcing cap fight will cause problems on Offensive gamemodes. There should be a statement here to make sure this is only active on Warfare.

rcon/automods/seeding_rules.py
```diff
@@ -257,16 +259,40 @@ def punitions_to_apply(
...
+                    self.logger.debug("Player is on " + team + " side and winning")
+                    op = player['offense']
+                    oop = watch_status.offensive_points.setdefault(aplayer.name, 0)
+
+                    self.logger.debug(
+                        "Player had " + str(oop) + " offensive points last and now " + str(op))
+                    if oop != 0 and oop < op:
```
timraay, line 285: Correct me if I'm wrong, but whenever this rule becomes active you're not populating the watch status with the current offense score of players and instead default it to 0. If any players on the server already have an offense score before the seeding automod is enabled, either because the min/max player requirements are met or the service was just started, it may falsely flag them as attacking and warn or punish them.

rcon/automods/seeding_rules.py
```diff
@@ -257,16 +259,40 @@ def punitions_to_apply(
...
+                        violations.append(ecf.message)
+
+                        if ecf.skip_warning:
+                            warnings = watch_status.warned.setdefault(aplayer.name, [])
+                            for _ in range(self.config.number_of_warning):
+                                warnings.append(datetime.now() - timedelta(seconds=self.config.warning_interval_seconds + 1))
+                    watch_status.offensive_points[aplayer.name] = op
```
timraay, line 292: Why are we using player names and not steam IDs as key here?

rcon/automods/seeding_rules.py
```diff
@@ -257,16 +259,40 @@ def punitions_to_apply(
...
+                else:
+                    watch_status.offensive_points.clear()
```
timraay, line 294: If a player's team is `None` (either because they are on team select screen or have a space as 20th char in their name) this will end up triggering this `else`-clause.

timraay (changes requested):

rcon/automods/seeding_rules.py
FlorianSW: Good point! 👍

rcon/automods/seeding_rules.py
FlorianSW: I think that is kind of right.

The seeding automod cannot go into "active" once it was deactivated once for the round (that prevents situations, especially with other rules, that a tank crew is punished because they used a tank when the player requirement allowed it and then it drops below the requirement).
However, when the automod is started _during_ a round which matches the seeding requirements, I think you're right, that it will punish players with a pre-existing Offensive point. Let me take a look how to work that in, so that this does not happen. I think changing the default value of `oop` (old offensive points) from 0 (which is a valid in-game value) to -1 might be the easiest solution (while ignoring an oop of -1 in the "should punish check").

rcon/automods/seeding_rules.py
FlorianSW: good question, I actually don't know, it was that case when I started working with the existing automod.
However, I think it might be worth revisiting this and use steam IDs instead. But, I would not do that in this PR and instead in a separate one, what do you think?

rcon/automods/seeding_rules.py
FlorianSW: When the player does not have a team, then they should not be included in this for loop at all (we iterate through the squads of each team in the team_view, ignoring squads which do not have a name (where players in the team selection or squad selection screen appear). See L234. Does this resolve your point? :)

rcon/automods/seeding_rules.py
timraay: Sure thing

timraay (approved):

PR #235, opened 2023-03-15
Add missing positional argument

FlorianSW:
Fixes a missing argument for the no leader automod (from integrating a new seeding automod rule)


cemathey (approved):
lgtm

PR #238, opened 2023-03-17
really ignore commander

FlorianSW:


FlorianSW:
Will need to be tested on a real server. I will do that today in the afternoon.

timraay (approved):

PR #240, opened 2023-03-27
Show with what weapon a player was killed

FlorianSW:
This was a feature request I recently got multiple times (and found myself as well). Instead of just showing with what weapon a player made their kills, it is also of interest with what weapon they were killed the most (e.g. artillery, MGs, etc).

As this is pretty cheap (from a calculation point of view) to implement, let's simply add it :D


cemathey (approved):
lgtm, I built it and ran it against our game server for a minute and the scoreboard looks fine to me.

PR #247, opened 2023-04-02
Show and allow to filter by steam ID on scoreboards

FlorianSW:
This commit alsoc hanges the default visibility of the steam ID column to false. This info is mostly not needed, except people really want to show it. On the other hand, the column is pretty wide and takes huge amount of space in the table, which is already pretty crowded.

Filtering by steam ID is also possible when the column is not displayed, so this should be fine.


cemathey (approved):
lgtm if you tested it on a live one and works

PR #250, opened 2023-04-11
Create endpoint permissions / remove unused permissions

cemathey:
Creates a permission for each endpoint except as detailed below. It removes the unused permissions that are auto-generated (see migration 4) and creates a data migration to create the new permissions in existing CRCON installations (see migration 2).

Default groups are created in migration 3, (`admin` has full access, `moderator` mirrors the old style `@mod_users_allowed` permission except it has access to all the view permissions and `read_only` only allows viewing information).

`/admin/logs` is from the `django-directory` app, it is only viewable by users with staff accounts already so I left it alone.

Since we're using Django migrations to create the default groups, this will only modify the groups the first time the migration is run (when people first upgrade to it, or create a fresh CRCON install) and they can then modify them or create new groups, etc. without their changes being over written, so they can tailor their users exactly how they want, but we provide useful defaults.

I think ideally we release this on its own, give people a week or so to handle upgrading, and then release the other (I still have to do a PR for it on a separate branch) part that actually applies the permissions to end points, that way people can get their account access handled without having to immediately fix CRCON user accounts that can't do anything anymore.

No permissions required
*admin/ (this is handled by Django's existing permissions)
*api/ login
*api/ logout
*api/ is_logged_in

I didn't touch these endpoints that use `@stats_login_required` so that we don't break existing things since it ties into the config file and scorebot and who knows what else.
*api/ live_scoreboard
*api/ get_scoreboard_maps
*api/ get_map_scoreboard
*api/ get_live_game_stats
*api/ get_map_history


MarechJ (approved):
fine to merge but there's no permission enforcement as far as I can see

PR #251, opened 2023-04-11
Apply permissions created in PR #250

cemathey:
This is dependent on #250   I did some manual testing and I am *fairly* certain that this is all working correctly but it's a big change and I had to look up a bunch of Django stuff to do this, so a thorough review is necessary.

If #250  and this one is approved, we should do a release of PR #250 prior to this one so that users have time to create/modify/apply groups and permissions.

* Removes a lot of unused imports
* Applies Django's `permission_required` to each endpoint so permissions created in #250 take effect
* `RecordedRcon`'s endpoints are still being dynamically created.
* The portions of `Rcon` that were using thread pool stuff moved to `RecordedRcon` as `RecordedRcon` is where the API is exposed, and they wouldn't work anyway if you instantiate `Rcon` instead of `RecordedRcon` anyway.
* Removes the `mod_users_allowed` decorator and permission

Moving forward, any new endpoint would require the creation of a new permission (or it would be viewable by default, subject to being logged in or not if `@login_required` is applied) and labeling the endpoint with `@permission_required` unless it's a `RecordedRcon` endpoint that is being automatically exposed (it's applied as part of the endpoint creation)

I found a few endpoints that weren't being exposed, I labeled them with `TODO` so we can come back to them later.


rcon/extended_commands.py
```diff
@@ -586,27 +528,26 @@ def get_gamestate(self) -> GameState:
         }
 
     @ttl_cache(ttl=2, cache_falsy=False)
-    def team_sizes(self) -> Tuple[int, int]:
+    def get_team_sizes(self) -> Tuple[int, int]:
```
MarechJ, line 531: by changing the name to get_ you're exposing those to the API layer too

rcon/recorded_commands.py
```diff
@@ -53,7 +54,25 @@ def thread_pool(self):
     def run_in_pool(self, function_name: str, *args, **kwargs):
         return self.thread_pool.submit(getattr(self, function_name), *args, **kwargs)
 
-    @mod_users_allowed
+    @ttl_cache(ttl=2)
```
MarechJ, line 57: not sure if it's wise to cache that, the calls inside it are already cached no? and now you should invalidate on kick/ban/move player

rconweb/api/audit_log.py
```diff
@@ -65,6 +61,7 @@ def wrapper(func):
 
 @csrf_exempt
 @login_required()
+@permission_required("api.can_view_audit_logs_autocomplete", raise_exception=True)
```
MarechJ, line 64: oh wow you bothered typing it all? I'd have just done that automatically by inspecting the function name that is being decorated

rconweb/api/views.py
```diff
@@ -71,27 +72,31 @@ def set_name(request):
...
 def set_temp_broadcast(request):
     return set_temp_msg(request, temporary_broadcast, "set_temp_broadcast")
 
 
+# TODO: this is not an exposed endpoint
 @csrf_exempt
-@login_required(True)
+@login_required()
 @record_audit
 def set_temp_welcome(request):
```
MarechJ, line 87: no permission required here, intended?

rconweb/api/views.py
```diff
@@ -71,27 +72,31 @@ def set_name(request):
     )
 
 
+# TODO: this is not an exposed endpoint
 @csrf_exempt
-@login_required(True)
+@login_required()
 @record_audit
 def set_temp_broadcast(request):
```
MarechJ, line 79: no permission required here, intended?

rconweb/api/views.py
```diff
@@ -71,27 +72,31 @@ def set_name(request):
...
 @csrf_exempt
-@login_required(True)
+@login_required()
 @record_audit
 def set_temp_welcome(request):
     return set_temp_msg(request, temporary_welcome, "set_temp_welcome")
 
 
 @csrf_exempt
+@permission_required("api.can_view_crcon_version", raise_exception=True)
```
MarechJ, line 92: hmm this is called without requiring a login though

rconweb/api/views.py
```diff
@@ -71,27 +72,31 @@ def set_name(request):
...
 def get_version(request):
     res = run(["git", "describe", "--tags"], stdout=PIPE, stderr=PIPE)
     return api_response(res.stdout.decode(), failed=False, command="get_version")
 
 
 @csrf_exempt
+@permission_required("api.can_view_public_info", raise_exception=True)
 def public_info(request):
```
MarechJ, line 100: This is public info, you don't need a login. The stats page and the score bot use that. You'll break stuff I think

rconweb/api/views.py
```diff
@@ -291,20 +302,23 @@ def _do_watch(request, add: bool):
 
 @csrf_exempt
 @login_required()
+@permission_required("api.can_watch_players", raise_exception=True)
```
MarechJ, line 305: that permision name is missleading

rconweb/api/views.py
```diff
@@ -645,7 +679,8 @@ def get_connection_info(request):
 
 
 @csrf_exempt
-@login_required(True)
+@login_required()
+@permission_required("api.can_run_raw_commands", raise_exception=True)
```
MarechJ, line 683: maybe you should do a super_admin group for that one

MarechJ (approved):

rconweb/api/views.py
cemathey: Yeah, it isn't actually exposed at all so people couldn't use it, so I skipped it to come back to later.

rconweb/api/views.py
cemathey: Yeah, it isn't actually exposed at all so people couldn't use it, so I skipped it to come back to later.

rcon/extended_commands.py
cemathey: Whoops, that was intentional but in retrospect I should not I think since we already have that information available elsewhere and in a nicer format.

rconweb/api/audit_log.py
cemathey: I thought about that, and it was a lot of copy/pasting around, but I wanted it to be searchable for people who might be not very familiar with the code base but want to mess with their permissions groups.

rconweb/api/views.py
cemathey: Whoops, fixed!

rconweb/api/views.py
cemathey: Whoops, fixed.

rconweb/api/views.py
cemathey: Added an owner group in PR 250 that has that permission and removed it from the admin group.

PR #252, opened 2023-04-11
Add missing alembic db migration for death_by_weapon

FlorianSW:


cemathey:
I built a version of the last good release (v6.1.4), let it collect some games, then switched to master and pulled in these changes, built new images and let it collect some more games.

Both historical and new games worked fine.

cemathey (approved):

PR #255, opened 2023-04-30
Feature/watchlist improvements

cemathey:
Update the watchlist feature to track who added it and the number of times a player has connected since they've been watched.

* Drop the `comment` column in `player_watchlist` as it is unused and superseded by the `player_comments` table
* Add a modified timestamp to `player_watchlist`
* Add an author field to `player_watchlist`
* Add a session count to `player_watchlist`
* Add some miscellaneous typing
* Update the discord embed for watches to include the new information
* Fix the reason for a watch not being reset after the very first time it was set
* Include the player name on the API calls to watch/unwatch players so it shows in the Discord audit trail
* Remove some old unused console logging in the frontend

I tested this on a clean build (built the current version, added/removed a watch, and then built with these changes) to make sure the migration was applied and it all seemed to work fine to me.


FlorianSW (approved):
Looks good 👍

PR #257, opened 2023-05-07
Add CE, Off, Def and support points

FlorianSW:


rcon/utils.py
```diff
@@ -382,22 +390,37 @@ def __len__(self):
...
+    support: int
+
+
+class MapInfo(TypedDict):
+    name: str
+    start: None | float
+    end: None | float
+    guessed: bool
+    player_stats: dict[str, PlayerStat]
+
```
cemathey, line 406: Think you could add these to `rcon/types` so we have all (or most of) our typedefs in one spot?

rconweb/api/scoreboards.py
```diff
@@ -1,3 +1,4 @@
+import json
```
cemathey, line 1: imported but not used

cemathey:
I built a version from the current state of `master`, let it record a game, swapped to this branch and built a version and let it record a game and it upgraded and recorded games fine, so lgtm.

Could you `black` and `isort` this before merging though? There's some extra spaces and stuff in weird places.

cemathey (approved):

PR #261, opened 2023-05-18
remove useless logging

FlorianSW:


rcon/game_logs.py
```diff
@@ -151,6 +151,8 @@ def run(self, loop_frequency_secs=2, cleanup_frequency_minutes=10):
                 last_cleanup_time = datetime.datetime.now()
 
             dp = self.rcon.get_detailed_players()
+            if dp['fail_count'] > 0:
```
cemathey, line 154: do we want to do a .get() here in case the key is missing?

FlorianSW: not really necessary. I think this one is always set :)

cemathey (approved):
lgtm

PR #263, opened 2023-05-21
use vite for faster and supported way to build react frontend

FlorianSW:
Create React App is unsupported and mainly unmaintained. It also does not build stuff very fast and efficiently. Vite is one alternative, which fixes all these points.


FlorianSW:
I run the thing locally as well as when building it with docker. However, at never with a running backend so far :D

rcongui/.env
```diff
@@ -1 +1 @@
-REACT_APP_API_URL=http://127.0.0.1:8000/api/
+REACT_APP_API_URL=http://localhost:3000/api/
```
cemathey, line 1: Do we want to change this?

FlorianSW: Yeah, the backend runs on 8000, however, there is a proxy in vite.config.js, which proxies the api requests to the backend running on 8000 :)
I usually do it like that, as it (from my point of view) is more natural to look for the API url in the config there. And I use that proxy mechanism, when I want to hook up my local frontend with a live backend (without running that locally). Doing it with the vite proxy allows me to go around CORS 🤣

cemathey (approved):
Dev server worked for me this time and it built, so lgtm but I know very little about Javascript and its tooling.

PR #265, opened 2023-05-23
U14 updates

cemathey:
Format changes because people refuse to use `black` 😉 

*Add the new endpoints
`get_current_map_sequence`
`get_map_shuffle_enabled`
`set_map_shuffle_enabled`
*Add the new map names (still need to do `LOG_MAP_NAMES_TO_MAP`)
*Add the new map pictures

New RCON commands are reported to be:
```
QueryMapShuffle
ToggleMapShuffle
ListCurrentMapSequence
```

I am not sure if `set_map_shuffle_enabled` actually takes an argument or if it is a true toggle but it is callable without a parameter just in case.


FlorianSW (approved):
lgtm

PR #266, opened 2023-05-26
Fixes for U14 fuckups

FlorianSW:
This is all untested right now, until I've found a server, which's MapRotation.ini is fixed already.

### DIFF
--- a/rcon/commands.py
+++ b/rcon/commands.py
@@ -526,7 +526,7 @@ class ServerCtl:
         return self._request("listcurrentmapsequence").split("\n")[:-1]
 
     def get_map_shuffle_enabled(self):
-        return self._request("querymapshuffle")
+        return self._request("querymapshuffle").endswith('TRUE')
 
     def set_map_shuffle_enabled(self, enabled: str | None = None):
         if enabled is None:
@@ -580,19 +580,17 @@ class ServerCtl:
     def do_add_map_to_rotation(
         self,
         map_name: str,
-        after_map_name: str = None,
+        after_map_name: str,
         after_map_name_number: str = None,
     ):
-        cmd = f"rotadd {map_name}"
-        if after_map_name:
-            cmd = f"{cmd} {after_map_name}"
-            if after_map_name_number:
-                cmd = f"{cmd} {after_map_name_number}"
+        cmd = f"rotadd /Game/Maps/{map_name} /Game/Maps/{after_map_name}"
+        if after_map_name_number:
+            cmd = f"{cmd} {after_map_name_number}"
 
         return self._request(cmd, can_fail=False, log_info=True)
 
     def do_remove_map_from_rotation(self, map_name, map_number: str = None):
-        cmd = f"rotdel {map_name}"
+        cmd = f"rotdel /Game/Maps/{map_name}"
         if map_number:
             cmd = f"{cmd} {map_number}"
 
--- a/rcon/commands.py
+++ b/rcon/commands.py
@@ -528,11 +528,10 @@ class ServerCtl:
     def get_map_shuffle_enabled(self):
         return self._request("querymapshuffle").endswith('TRUE')
 
-    def set_map_shuffle_enabled(self, enabled: str | None = None):
-        if enabled is None:
-            return self._request(f"togglemapshuffle")
-        else:
-            return self._request(f"togglemapshuffle {escape_string(enabled)}")
+    def set_map_shuffle_enabled(self, should_be_enabled: str | None = None):
+        enabled = self.get_map_shuffle_enabled()
+        if enabled != should_be_enabled:
+            self._request(f"togglemapshuffle")
 
     def set_idle_autokick_time(self, minutes):
         return self._request(f"setkickidletime {minutes}", log_info=True)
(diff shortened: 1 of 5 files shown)


FlorianSW:
Got a server to test that shit:
- ✅ Toggling map shuffle
- ✅ Replacing the entire map rotation
- ✅ re-ordering the current map rotation
- ✅ adding a map to the map rotation (at various positions)
- ✅ removing a map from rotation

I think that should be it.

rcon/extended_commands.py
```diff
@@ -998,7 +1004,7 @@ def set_maprotation(self, rotation):
         rotation = list(rotation)
         logger.info("Apply map rotation %s", rotation)
 
-        current = self.get_map_rotation()
+        current = super().get_map_rotation()
```
FlorianSW, line 1007: as we rely on the current map rotation to rebuild the current one, I thought it would be wise to bypass any cached version we might have in case the server shuffled the map rotation or someone changed it with another rcon tool already.

rcon/extended_commands.py
```diff
@@ -968,7 +969,7 @@ def get_map_rotation(self):
         return l
 
     def do_add_map_to_rotation(
-        self, map_name, after_map_name: str = None, after_map_name_number: str = None
+        self, map_name, after_map_name: str, after_map_name_number: str = None
```
timraay, line 972: You really should keep `after_map_name` optional. When not provided, you may assume that the user intends to append the map to the end of the rotation. That way we preserve the function's signature and behavior.

timraay (changes requested):
Still some things that need changing :)

rcon/commands.py
```diff
@@ -526,7 +526,7 @@ def get_current_map_sequence(self):
         return self._request("listcurrentmapsequence").split("\n")[:-1]
 
     def get_map_shuffle_enabled(self):
-        return self._request("querymapshuffle")
+        return self._request("querymapshuffle").endswith('TRUE')
 
     def set_map_shuffle_enabled(self, enabled: str | None = None):
```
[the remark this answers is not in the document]
FlorianSW: It is already cached (caching seems to be only done in extended_commands, not in commands). The other stuff I change 👍

rcon/commands.py
```diff
@@ -580,19 +580,17 @@ def do_switch_player_now(self, player):
     def do_add_map_to_rotation(
         self,
         map_name: str,
-        after_map_name: str = None,
+        after_map_name: str,
         after_map_name_number: str = None,
```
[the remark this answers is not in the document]
FlorianSW: I was wondering that as well, but didn't want to make too much changes in the api 🙈

rcon/extended_commands.py
```diff
@@ -633,7 +633,8 @@ def set_map(self, map_name):
                 if res != "SUCCESS":
                     raise CommandFailedError(res)
             except CommandFailedError:
-                self.do_add_map_to_rotation(map_name)
+                maps = self.get_map_rotation()
+                self.do_add_map_to_rotation(map_name, maps[len(maps) - 1], str(len(maps)))
```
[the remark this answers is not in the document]
FlorianSW: Good catch, one leftover of my misunderstanding of the ordinal numbere 😅

rcon/extended_commands.py
```diff
@@ -987,8 +988,13 @@ def do_remove_maps_from_rotation(self, maps):
 
     def do_add_maps_to_rotation(self, maps):
         with invalidates(Rcon.get_map_rotation):
+            existing = self.get_map_rotation()
+            last = existing[len(existing) - 1]
+            map_index = {last: len(existing)}
```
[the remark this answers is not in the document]
FlorianSW: I think this is just a wording issue with the variable name, right? I've trouble to find an accurate name here, tbh. I'll change that to map_number, but even that does not feel entirely correct.

From the working of the variable, it should be correct, isn't it? 🤔

rcon/extended_commands.py
FlorianSW: lol, `invalidates` invalidates the cache _before_ and _after_ executing, I didn't expect that :( Then I can move querying the current map rotation a bit down and it should be save as well.

rcon/extended_commands.py
FlorianSW: Then we would need to query the current map rotation before, isn't it? As the second argument is now not optional anymore.

rcon/extended_commands.py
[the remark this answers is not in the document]
FlorianSW: Ah, no, got it, changed! 👍

rcon/commands.py
```diff
@@ -526,13 +526,12 @@ def get_current_map_sequence(self):
...
+        return self._request("querymapshuffle").endswith('TRUE')
 
-    def set_map_shuffle_enabled(self, enabled: str | None = None):
-        if enabled is None:
-            return self._request(f"togglemapshuffle")
-        else:
-            return self._request(f"togglemapshuffle {escape_string(enabled)}")
+    def set_map_shuffle_enabled(self, should_be_enabled: str | None = None):
+        enabled = self.get_map_shuffle_enabled()
+        if enabled != should_be_enabled:
```
timraay, line 533: You're comparing a bool to a string

rcon/commands.py
```diff
@@ -581,18 +580,20 @@ def do_add_map_to_rotation(
...
-            cmd = f"{cmd} {after_map_name}"
-            if after_map_name_number:
-                cmd = f"{cmd} {after_map_name_number}"
+        if after_map_name is None:
+            current = self.get_map_rotation()
+            after_map_name = current[len(current) - 1]
+            after_map_name_number = current.count(after_map_name)
+        cmd = f"rotadd /Game/Maps/{map_name} /Game/Maps/{after_map_name}"
+        if after_map_name_number:
+            cmd = f"{cmd} {after_map_name_number}"
```
timraay, line 591: Only in extended_commands should `after_map_name` remain optional. `commands` is only a direct wrapper of the protocol's commands. And more importantly, you can not invalidate the response of `extended_commands`'s `get_map_rotation` like this.

timraay (changes requested):
a

rcon/extended_commands.py
timraay: Yes.

rcon/commands.py
FlorianSW: 🙈

rcon/commands.py
FlorianSW: Good point 👍

timraay (approved):
woooo!

PR #268, opened 2023-05-30
Show Steam ID in stats when steam API key is not given

FlorianSW:
Instead of retrieving the steam ID from the steam info (which is fetched from the Steam API), we should take it from our own records. While having a steam API key set is recommended, it is not enforced. Not having one would show the stats page without steam IDs, which is inconsistent depending on the setup. Instead, always show the Steam ID as we have that in our DB anyways.

### DIFF
--- a/rcon/models.py
+++ b/rcon/models.py
@@ -490,6 +490,7 @@ class PlayerStats(Base):
         nullable=False,
         index=True,
     )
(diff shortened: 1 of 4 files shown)


cemathey (approved):
Build runs and works fine on my end, lgtm.

PR #272, opened 2023-06-03
Feature/level kicks - Add auto kick by minimum level at player's connection feature

OlolBzH:


zMoooooritz:
Thank you for your contribution :)

Looks good but has one major flaw that isn't your fault...
The game servers are still somewhat inconsistent and return at certain events the wrong level of a player.
This is known as the "level 1" bug, which can also be observed ingame.
The problem is caused by connection issues to some servers in the background, but resolves itself most of the time after 5-10 minutes.

Therefore two changes are needed:
1. Don't kick players whose level is reported as 1
2. We need to check the level restriction periodically and kick if required

If you are willing to do those changes have a look at the `rcon/automods` directory and implement the logic as an automod.

Additionally this feature could be extended by adding level caps to specific roles

OlolBzH:
Thanks for your review.

> The game servers are still somewhat inconsistent and return at certain events the wrong level of a player. This is known as the "level 1" bug, which can also be observed ingame. The problem is caused by connection issues to some servers in the background, but resolves itself most of the time after 5-10 minutes.
> 
> Therefore two changes are needed:
> 
> 1. Don't kick players whose level is reported as 1
> 2. We need to check the level restriction periodically and kick if required
> 
> If you are willing to do those changes have a look at the `rcon/automods` directory and implement the logic as an automod.

About this problem, it looks like and edge issue that occurs randomly. Maybe a simple reconnexion can fix it and adding a false positive warning in the message can be enough. 
It could be more problematic with max level feature but a tool is not perfect and we already need moderators to handle edge cases. 
I think this feature does not imp)lies the setup of a new automod feature wich seems to me a bit overkill regarding the need. 

> Additionally this feature could be extended by adding level caps to specific roles
Added feature in last commits

zMoooooritz:
> Thanks for your review.

You're welcome :) 
> About this problem, it looks like and edge issue that occurs randomly. Maybe a simple reconnexion can fix it and adding a false positive warning in the message can be enough. It could be more problematic with max level feature but a tool is not perfect and we already need moderators to handle edge cases. I think this feature does not imp)lies the setup of a new automod feature wich seems to me a bit overkill regarding the need.

To be honest it is not a rare occurrence to see this bug especially on Tuesday (Steam maintenance day).
Obviously one can just reconnect, but the player needs to wait in the queue again (in case the server is filled up) which in turn may result in the feature not being used at all (while having false positives), that is at least my experience with automod features that are faulty at times (squad leader automod).

> > Additionally this feature could be extended by adding level caps to specific roles
> 
> Added feature in last commits

There may have been a communication issue here :/
What I meant was adding the ability to restrict players from taking specific roles by their level i.e. sl can only be taken by players above level 25 while command is restricted to players above level 50.
Since there are no hooks in crcon to check for role changes currently, it would not be sufficient to only check the player level `on_connected`.
Furthermore this feature should include warnings to not directly kick players from the server.
All in all this qualifies for a full fledged automod imo.

OlolBzH:
Understood, I did not know this "level 1" problem occurs so frequently.
I reworked the feature to use the automod and added level thresholds by role.

OlolBzH:
Added kick enforcement for global level thresholds (min/max) violations.

zMoooooritz (approved):
lgtm

Paulfyar:
Hello,
Install it but can't get turn on, the other script work well but not level_threholds.py
- i edited my files with "nano"
- download your git and copy/paste it on my existing one (PNSR server FR)
- can't figure it out what it's wrong.

cemathey:
I built this and tested it against a live server, however it immediately kicked me despite being configured to issue warnings.

Paulfyar:
ODE LOCAL:

docker-compose -f docker-compose.yml -f docker-compose.stage.yml build
docker-compose -f docker-compose.yml -f docker-compose.stage.yml up -d --force-recreate

SI MODIF Config.yml faire ON/OFF -Automod
											-Log_event_loop

OlolBzH:
Review config's comment to explain global vs by role behavior.

Added min_players option to level thresholds by role configuration (not tested I do not have access to a running server for now).

cemathey:
I built and tested it against our event server and everything I tried seemed to work fine to me.

PR #278, opened 2023-06-18
Add map voting whitelist endpoints

cemathey:
Add the vote map whitelist stuff as endpoints.

I did build this, I only explicitly tested `get_map_whitelist`, `do_add_map_to_whitelist` and `do_remove_map_from_whitelist` endpoints.

`reset_votemap_state` wasn't instantiating an instance of `VoteMap` so apparently it's been broken for a bit? Or I'm just confused, but I added it so it should work now if it wasn't.

I am not very familiar with map voting, we don't use it on our server so someone should take a close look at this.


rconweb/api/votemap.py
```diff
@@ -109,7 +115,6 @@ def reset_votemap_state(request):
             command="reset_votemap_state",
         )
     audit("reset_votemap_state", request, {})
-    v = VoteMap()
```
timraay, line 112: You said it wasn't instantiating a VoteMap instance here. Clearly it did, but you must've accidentally removed it. Not a big deal, though preferably you only instantiate it after you check if the request method is indeed POST.

rconweb/api/votemap.py
```diff
@@ -122,3 +127,158 @@ def reset_votemap_state(request):
...
+@login_required(True)
+@record_audit
+def do_add_map_to_whitelist(request):
+    data = _get_data(request)
+    try:
+        map_name = data["map_name"]
+    except KeyError:
+        return api_response(
+            failed=True,
+            result=None,
```
timraay, line 153: The result should be a very simple error message ideally.

rconweb/api/votemap.py
```diff
@@ -122,3 +127,158 @@ def reset_votemap_state(request):
...
+@record_audit
+def do_add_maps_to_whitelist(request):
+    data = _get_data(request)
+
+    try:
+        map_names = data["map_names"]
+    except KeyError:
+        return api_response(
+            failed=True,
+            result=None,
```
timraay, line 179: Same here, would need basic error message. This applies to more functions further down it seems.

timraay (approved):

rconweb/api/votemap.py
cemathey: Now that I understand better, it was instantiating it below but we were trying to use it in the response for non-POST requests.

I changed it to an error message instead.

PR #279, opened 2023-06-18
Move VAC ban settings to default_config.yml and allow whitelisted fla…

cemathey:
* Move the VAC ban settings from `.env` to `default_config.yml`
* Allow whitelisted player flags to skip the VAC ban check
* Update unit tests to reflect changes

I did run the tests and add a couple more to reflect the changes and added some typing.

This should be mergeable when we're ready to do a release, I was able to test it against a live server and it works, and I tested it after modification to make sure the rest of the steam API stuff works (looking up bans for the Live view, which works).

I did not deprecate the environment variables from `.env` or any of the Docker files yet, if we do merge this we can do that in a later release once people have had time to update their settings.

Preferably we do the deprecation after we do PR 277 (if we merge/release that) so that we can change the `.env` file easier in future releases.


config/default_config.yml
```diff
@@ -47,6 +47,15 @@ BANS:
   # Same as above but the other way around, remove a player from the blacklist will unban him
   unblacklist_does_unban: yes
 
+VAC_GAME_BANS:
```
timraay, line 50: You should add some more comments here.

config/default_config.yml
```diff
@@ -47,6 +47,15 @@ BANS:
...
+  steam_api_key: ""
```
timraay, line 51: Steam API key config key value should not be under the VAC game ban config. It should either be entirely separate or stay inside of the .env

timraay (changes requested):

cemathey:
The latest commits should cover the requested changes, I also added some better logging for steam API errors.

I accidentally rate limited myself while working on this with my steam API key, but we were swallowing those errors, now it should log a stack trace and people can see it happening.

timraay (approved):

PR #285, opened 2023-07-07
Populate DB URI from .env vars to allow ? in passwords and add healthchecks

cemathey:
`?` and `#` were incompatible DB passwords because the `DB_URL` was being parsed as a URL to get its component pieces.

* Add `.env` variables for the components of the postgres DB URI
* Pulls DB URI components from the environment instead of parsing the URL to allow the `?` character in passwords
* Adds Docker health checks for `postgres` and `redis` and delays starting the backend containers until they report healthy
* Add `.env` variable for the `redis` hostname and update the associated compose/etc. files to match
* Renames the `redis` environment variables to prefix with `HLL_` to match the postgres naming convention

I did build and connect this to two servers and I believe it works fine.

I don't think this is that useful (outside of expanding the usable characters for the database password) but there's really no reason not to let people pass env variables to point at a different database/redis if they want without having to hand edit compose files.


cemathey:
After a little bit of experimentation I had to tweak this slightly (our postgres connection URL wasn't including the database name) to get it to work properly with non-default database names.

With fresh eyes I also found a couple of spots I hadn't updated the redis env variables correctly so I fixed that as well.

PR #287, opened 2023-07-18
Hotfix/permissions

cemathey:
* Start auto settings by default (not sure why it was ever off to begin with)
* Fix some mislabeled permissions for `RecordedRcon` (missing app label)
* Move `get_online_console_admins` to `RecordedRcon` since it uses `get_players()`
* Move the cache invalidation for things that use `get_players` to `RecordedRcon`

I did build this and test it against the event server and kick/temp/perma ban and blacklist myself and it all works.

I'm not sure I entirely understand the cache invalidation stuff though if anyone sees a problem with this.


config/supervisord.conf
```diff
@@ -38,7 +38,7 @@ command=/code/manage.py auto_settings
 environment=LOGGING_FILENAME=auto_settings_%(ENV_SERVER_NUMBER)s.log
 startretries=100
 startsecs=1
-autostart=false
+autostart=true
```
FlorianSW, line 41: ehhh, is tthis really required? Meaning: Why would we change that default and why in a hotfix?

FlorianSW (changes requested):
The cache invalidation changes are out of scope of the permissions hotfix as well, isn't it? Or do I miss something here?

PR #288, opened 2023-07-28
Start auto-settings by default

cemathey:
Every other service starts by default, I don't think there's a downside to starting auto-settings by default also.

Individual installs can easily override their supervisord.conf if they want


FlorianSW:
Uhm, I think there is a default auto-settings config, which also _does_ things. I don't think we should magically enable auto-settings by default which might change peoples configuration to undesired settings.

PR #289, opened 2023-07-30
Feat/api documentation

cemathey:
Adds the `get_api_documentation` endpoint to automatically generate some minimal API documentation (endpoints, parameter names, default values and doc strings) for each endpoint.

Adds the new permission to the default groups.

The return type isn't particularly useful right now, but could be if we add more typing.

An example auto-documented endpoint:

```json
{
        "endpoint": "do_add_vip",
        "arguments": {
          "name": {
            "default": null,
            "annotation": null
          },
          "steam_id_64": {
            "default": null,
            "annotation": null
          },
          "expiration": {
            "default": "",
            "annotation": "<class 'str'>"
          }
        },
        "return_type": null,
        "doc_string": "Adds VIP status on the game server and adds or updates their PlayerVIP record."
}
```

### DIFF
--- a/rconweb/api/urls.py
+++ b/rconweb/api/urls.py
@@ -1,4 +1,8 @@
+from inspect import _empty, getdoc, signature, unwrap
+from typing import Callable
+
 from django.urls import path
+from django.views.decorators.csrf import csrf_exempt
 
 from . import (
     audit_log,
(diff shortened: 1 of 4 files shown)


rconweb/api/urls.py
```diff
@@ -14,51 +20,102 @@
...
+def _get_empty(value):
+    """Return None for inspect._empty values or the value as a string"""
+    if value is _empty:
+        return None
+    else:
+        return str(value)
+
+
+@csrf_exempt
+@login_required()
```
FlorianSW, line 35: does this really need a login, btw.? 🤔

cemathey: Probably not now that I think about it, we already expose the CRCON version publicly even when someone isn't logged in, and if you know the version then you know the API.

The only time it wouldn't match is if someone was building their own images on an altered build.

I'll remove the permissions.

FlorianSW (approved):

PR #291, opened 2023-08-03
Maint/upgrade django

cemathey:
We can jump from Django `3.1` to `4.2` with almost no changes, but Django started implementing CSRF origin checking in `4.0`.

This means that (the CRCON frontend works fine) to use the admin site users would have to configure their `RCONWEB_SERVER_URL` environment variable to match their originating host (url, http://<vps ip>:<port>, etc.) or they get a CSRF error on any page on the admin site.

For example: `RCONWEB_SERVER_URL=http://localhost:8010`

`django-directory` was incompatible with Django `4.0`+ but I forked it/patched it (the `url` method was deprecated) and updated `requirements.txt` to point to the working version.

`CORS_ORIGIN_WHITELIST` is renamed to `CORS_ALLOWED_ORIGINS`  to match `django-cors-headers` documentation (the old name is an alias for now, but I changed it in case they deprecate it in the future)

Django `3.1` ends LTS in April 2024 so it's still supported, but prior to Django `4.1` they used `jQuery` to add Javascript to manage formset events on the admin page and I'm working on that on a separate branch and would rather just do it once, in the modern style with the latest version of Django.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,12 @@ Per server settings
     HLL_HOST
     HLL_PORT
     HLL_PASSWORD
+    RCONWEB_SERVER_URL
+
+**You must configure `RCONWEB_SERVER_URL` for each server you're setting up to match the URL you're hosting CRCON on, or you will be unable to access the admin site due to CSRF errors**
+
+For example if you are hosting using `HTTPS` on `example.com` you would set `RCONWEB_SERVER_URL=https://example.com`
+For example if you are hosting using `HTTP` on `127.0.0.1` on port `8010` you would set `RCONWEB_SERVER_URL=http://127.0.0.1:8010`
 
 There are other optional (but beneficial) settings in your `.env` such as Discord integration.
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 Click~=8.0.0
 cachetools==4.0.0
-Django~=3.2.0
-django-cors-headers==3.5.0
+Django~=4.2.0
+django-cors-headers==4.2.0
 redis==3.5.3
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -138,6 +138,12 @@ Per server settings
     HLL_HOST
     HLL_PORT
     HLL_PASSWORD
+    RCONWEB_SERVER_URL
+
+**You must configure `RCONWEB_SERVER_URL` for each server you're setting up to match the URL you're hosting CRCON on, or you will be unable to access the admin site due to CSRF errors**
```
FlorianSW, line 143: I'm fine with that, but we should definitely point that out in the release announcement :D

requirements.txt
```diff
@@ -24,5 +24,7 @@ pandas>=1.4.1,<2.0.0
 pydantic==1.9.1
 discord.py==1.7.3
 pytest==7.2.0
-django-directory
+# django-directory
+# django-dirctory is currently incompatible with Django 4.0+, PR is pending to fix django-directory upstream
+git+https://github.com/cemathey/django-directory@f51ee1e8dc50edf453fee4a0d9631c0e46fe9433
```
FlorianSW, line 29: do we really need this? And if so, for what 😅

FlorianSW (approved):
Looks good so far (just one comment). However, didn't test it, yet :D

requirements.txt
cemathey: It's used for the api/logs/ endpoint that lets people browse logs through the UI.

README.md
cemathey: Definitely.

PR #292, opened 2023-08-07
Increase TCP socket buffer size

cemathey:
Increase the buffer size from 8,192 to 32,768 bytes. This aligns on a power of 2 (as the docs recommend) and should reduce the number of incomplete requests we get due to HLLs RCON protocol limitations.

### DIFF
--- a/rcon/connection.py
+++ b/rcon/connection.py
@@ -5,7 +5,7 @@ import time
 import uuid
 from threading import Lock, get_ident
 
-MSGLEN = 8196
+MSGLEN = 32_768
 TIMEOUT_SEC = 20
 
 logger = logging.getLogger(__name__)


FlorianSW:
What docs do you refer to? I think we barely read more than 8 kb (or even 32 kb) in one receive, except for logs and other types of lists. How would a higher buffer size protect against incomplete requests?

cemathey:
> What docs do you refer to? I think we barely read more than 8 kb (or even 32 kb) in one receive, except for logs and other types of lists. How would a higher buffer size protect against incomplete requests?

https://docs.python.org/3/library/socket.html#socket.socket.recv This recommends the power of 2 socket size, on master right now we are doing 2^13+4=8196

FlorianSW:
> https://docs.python.org/3/library/socket.html#socket.socket.recv This recommends the power of 2 socket size, on master right now we are doing 2^13+4=8196

Hmm, I read the docs like I would expect it: The buffer size should be a relatively small power of 2. 8kb is already quite big, 32kb is 4 times as big. There is _probably_ no issue with that, _however_ we will allocate much more memory for the responses while we mostly do not read that amount of data at all.

The 20 connections are per worker, iirc. Depending on active services you will have much more connections to the server.

cemathey:
> > https://docs.python.org/3/library/socket.html#socket.socket.recv This recommends the power of 2 socket size, on master right now we are doing 2^13+4=8196
> 
> Hmm, I read the docs like I would expect it: The buffer size should be a relatively small power of 2. 8kb is already quite big, 32kb is 4 times as big. There is _probably_ no issue with that, _however_ we will allocate much more memory for the responses while we mostly do not read that amount of data at all.
> 
> The 20 connections are per worker, iirc. Depending on active services you will have much more connections to the server.

Well I'm not married to the change, it would help people with more exotic character ranges in their game logs, CRCON as is works for most users though so it won't hurt my feelings to leave it as is.

Even if we leave it at its current size though we should probably change it to the actual power of 2 instead of 2**13 with 4 extra bytes allocated.

I did some minor digging into the C code for the socket module Python uses (haven't written any C in forever, so maybe I'm reading it wrong) and as far as I can tell it's only allocating that memory when it actually tries to receive data over that buffer and then it's accessible to the caller until it's garbage collected on the Python side.

It also resizes the buffer if the received data is smaller than the asked for length.

I think this means that we aren't actually using any memory until we both open a socket *and* try to read over it, and any open but unused sockets aren't allocating that memory.

I'm about as far from an expert on C or socket programming as you can get, but I think the end result here is we're talking about at most a megabyte or two of extra RAM usage in a program that uses a couple of gigabytes under normal usage.

FlorianSW:
I mean, if it actually improves things for _some_ and doesn't hurt anyone else, then I'm fine with merging it and see what people say in real setups (if they even see any consequences :D).

FlorianSW (approved):

PR #295, opened 2023-08-14
Feat/api keys

cemathey:
Adds the ability to authenticate with an API key and the HTTP authorization header instead of having to include a valid `sessionid` cookie with every request.

This should change nothing for anyone using the standard browser based version of CRCON, but allow easier usage of the API.

API keys are hashed (so we aren't storing them in plaintext in the database) using Django's default password hasher, and uses the `SECRET_KEY` as defined in `.env` as the salt to make sure we get deterministic values, otherwise the password hasher uses its own salt and the same value hashes to different values every time.

API keys are constrained to a minimum of 32 characters to try to help users not pick stupid defaults that will compromise them.

I'm using `uuidv4` from https://jspm.dev/uuid to generate default API keys in the admin view, I tested it without a network connection and nothing breaks, just throws an error in the console if that URL is unavailable. Users can just type whatever value they want in the UI if they want different keys, but that should provide a safe and useful default for most users.

Authentication wise it just looks for a matching API key and looks up the corresponding user, and then uses the normal user based Django authentication (to include permissions, etc.). API keys are constrained to be unique on the database side because of this.

Duplicate API keys are checked for during form validation, which does allow someone to determine if an API key is already in existence which is less than ideal, but this is part of the admin site and only accessible to trusted users.  AFAIK there is no way around this unless we were to also include a username as part of the HTTP authorization header.

I did all my testing through both the admin site to check for usability/errors, and Postman to send requests to the API and as far as I can tell it all works correctly.


timraay (approved):
If there's two things I'm bad at it's authentication and Django. Take my approval and pretend like this was a serious review.

PR #296, opened 2023-08-20
rcongui: Fix stats player URL when steaminfo is not available

MartijnHols:
The stats player name URL now falls back to a community URL with the steamID so the player name now always has a link to the user's profile.


cemathey (approved):
Looks good to me, I built a version with PR 296 and 298 in them and they seem to work fine, holding merge in case anyone else wants to comment.

PR #305, opened 2023-09-10
Autosettings: Add "set_map_shuffle_enabled" to default config

LordofAgents:
Hey everyone,

first, thank you for this great software!
I have noticed that the command "set_map_shuffle_enabled" is missing in the standard configuration for autosettings. Since the standard configuration serves as documentation, I think it makes sense to include this command in the file. In this pull request I have done exactly that. This should make it easier for new users (including me) to include it. 

Thank you very much!


cemathey (approved):
lgtm, I couldn't find any references to `config/auto_settings.default.json` though, I think that's a legacy file that isn't used anywhere.

Will hold on merging for awhile in case anyone has any comments, thanks for your contribution!

LordofAgents:
I used this file, to recover the lost default file on our deployment. So i think, if we update the default file, we should update it to

cemathey:
> I used this file, to recover the lost default file on our deployment. So i think, if we update the default file, we should update it to

Great point, didn't consider that, that's a lot easier for most people than trying to look inside the actual code itself for the defaults.

PR #308, opened 2023-09-19
Add server number to get_connection_info endpoint

cemathey:
* Add server number to get_connection_info endpoint
* Login is already required, and the server number doesn't seem like particularly privileged information
* Had a user request this

Output would look like:

```json
{
  "result": {
    "name": "Saucymuffin's [BEER] Haus & Squidd Cafe Part Deux",
    "port": "8010",
    "link": "http://localhost:8010",
    "server_number": "1"
  },
  "command": "get_connection_info",
  "arguments": null,
  "failed": false,
  "error": null,
  "forwards_results": null
}
```

### DIFF
--- a/rcon/utils.py
+++ b/rcon/utils.py
@@ -481,7 +481,7 @@ class ApiKey:
         }
 
 
-def get_server_number() -> int:
+def get_server_number() -> str:
     """Get the CRCON server number"""
     server_number = os.getenv("SERVER_NUMBER")
     if not server_number:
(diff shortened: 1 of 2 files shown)


timraay (approved):
Not sure if it would be smarter to cast server number to an int since that would make more sense, but it's kind of whatever.

cemathey:
> Not sure if it would be smarter to cast server number to an int since that would make more sense, but it's kind of whatever.

I changed it but only on the endpoint result, `get_server_number` is used all over the place internally and wasn't sure without a bunch of looking if I'd accidentally break something.

PR #310, opened 2023-09-25
Remove gendered references

cemathey:
Change some pronouns to be gender neutral


Noblezim711 (approved):
Code reviewed. It makes sense to change these, it makes it more readable, and conforms to the English language writing rules.

ElGuillermo (approved):

ElGuillermo (commented):
We'll definitely have to regroup all these sentences in a "language" file, thus we will be able to translate/change them from a single place.

PR #319, opened 2023-10-20
Feat/UI settings

cemathey:
# User Settings

## Summary

Essentially this moves almost all of the settings from `.env` and `config.yml` into the database and allows them to be set/viewed via the command line, or through endpoints.

All of the tests pass except with these changes except there are two TK ban on connect tests that fail (these also fail on master though).

All models (except auto settings, which remains the same at the moment) are Pydantic models which handles all validation and serialization.

All settings inherit from `rcon.user_config.utils.BaseUserConfig` and must implement `load_from_db()` and `save_to_db()`. Database keys are in the format `{server_number}_{class_name}` i.e. `1_ScorebotUserConfig`.

All models must be instantiable without needing to explicitly passing values, so all fields must have defaults. Seeding the database (except auto settings) is no longer required, if a model is not saved in the database it simply returns one with default settings.

All models must be serializable to JSON.

All models are stored in one complete JSON blob, they're no longer split (for some of the old user settings) across multiple key, value pairs in the database.

It *should* be impossible to throw an exception when loading user configs from the database, by default if there is an error it will return defaults. If the database isn't running, well nothing else works anyway so it doesn't matter.

It's also a lot easier to refactor, manage type checking and use built in linting stuff like pylance with Pydantic models versus loading stuff up as a dictionary from a yaml file.

All settings are now per server, there was a few things (scorebot, the generic log lines feature, etc.) that used to allow you to configure things across multiple servers inside one config file.

Most, but not all user settings support having multiple webhooks. (Even though it's not a great idea to use more than one or two, since it will block for each one)

Discord URLs are now `HttpUrl`'s across the board, which makes consuming them a little annoying (have to cast them to a string before passing them to `requests` or whatever), but the validation is worth it I think.

All user settings have a `get_`, `set_`, `validate_` and `describe_` endpoint. Validation endpoints do not actually save the model, just checks for validation errors

Describe endpoints return the Pydantic schema for the model.

All endpoints are commutative (or whatever the term is, I'm dumb), output from a `get_` is/should always be valid when fed to a `set_` endpoint.

Set endpoints can reset a config to default by including `reset_to_default` in the JSON body with a truthy bool and API errrors can be returned as JSON by including `errors_as_json` with a truthy bool.

## Miscellaneous Changes

* The settings page now has a `Reconnect To Gameserver` button which restarts `gunicorn` which causes the Rcon instance used to register the API to reconnect to the game server.
    * Individual services (log event loop, etc.) would still have their connections unless they're toggled off/on.
    * This was the only way I could think of to get the to get the API to reconnect to the game server without actually restarting all the containers which you can't do from inside of itself
* Punitions changed to punishments everywhere I saw it (both for translation purposes, and to support the standard messages endpoint changes)
* The API will return errors when parsing invalid JSON instead of defaulting to an empty parameter body
  * POST requests without a valid JSON body to decode will :warning: **fail** :warning: with a JSON decode error, so empty POST requests should be changed to GETs in tools that use the CRCON API
* The trailing / for server URLs for Django CSRF checking is now stripped so users can enter it either way

## Miscellaneous Bug Fixes

* `get_welcome_message` endpoint would fail if the welcome message was an empty (`b""`) value

## Backwards Incompatible Changes

* Admin pings no longer go to the chat webhook if an admin ping webhook isn't set.
* No automatic conversion of settings happens, when people upgrade they have to manually set all of their settings again
  * Settings aren't discarded, `config.yml`, and their database records are untouched, but not accessible through the UI in any way.
* `default_config.yml` exists, but all the settings are removed. Some users probably have edited this and will lose their settings, but they had to reconfigure anyway and they shouldn't have been editing that file, it's full of warnings not to.

## requirements.txt

* Pydantic version bumped from `1.9.1` to `2.3.0`
  * There was only a few miscellaneous method renamings that had to be changed to get onto the 2.x.x version
* discord-webhook switched from Marech's fork to my fork
  * He forked a long time ago to support timeouts, that is now in the main library
  * I forked to add `__len__` to embeds, I have a PR open but it doesn't look like it will be merged

## New Permissions

* `"can_view_auto_mod_level_config"`
* `"can_change_auto_mod_level_config"`
* `"can_view_auto_mod_no_leader_config"`
* `"can_change_auto_mod_no_leader_config"`
* `"can_view_auto_mod_seeding_config"`
* `"can_change_auto_mod_seeding_config"`
* `"can_view_tk_ban_on_connect_config"`
* `"can_change_tk_ban_on_connect_config"`
* `"can_view_expired_vip_config"`
* `"can_change_expired_vip_config"`
* `"can_view_server_name_change_config"`
* `"can_change_server_name_change_config"`
* `"can_view_log_line_discord_webhook_config"`
* `"can_change_log_line_discord_webhook_config"`
* `"can_view_name_kick_config"`
* `"can_view_rcon_connection_settings_config"`
* `"can_change_rcon_connection_settings_config"`
* `"can_view_rcon_server_settings_config"`
* `"can_change_rcon_server_settings_config"`
* `"can_view_scorebot_config"`
* `"can_view_standard_broadcast_messages"`
* `"can_change_standard_broadcast_messages"`
* `"can_view_standard_punishment_messages"`
* `"can_change_standard_punishment_messages"`
* `"can_view_standard_welcome_messages"`
* `"can_change_standard_welcome_messages"`
* `"can_view_steam_config"`
* `"can_change_steam_config"`
* `"can_view_vac_game_bans_config"`
* `"can_change_vac_game_bans_config"`
* `"can_view_admin_pings_discord_webhooks_config"`
* `"can_change_admin_pings_discord_webhooks_config"`
* `"can_view_audit_discord_webhooks_config"`
* `"can_change_audit_discord_webhooks_config"`
* `"can_view_camera_discord_webhooks_config"`
* `"can_change_camera_discord_webhooks_config"`
* `"can_view_chat_discord_webhooks_config"`
* `"can_change_chat_discord_webhooks_config"`
* `"can_view_kills_discord_webhooks_config"`
* `"can_change_kills_discord_webhooks_config"`
* `"can_view_watchlist_discord_webhooks_config"`
* `"can_change_watchlist_discord_webhooks_config"`
* `"can_restart_webserver"`

# Command Line Interface

This adds three new commands, `get_user_settings`, `set_user_settings`, and `reset_user_settings` which are pretty self explanatory.

Only JSON is supported for input/output, I had issues getting YAML to produce usable output because of tags, or something weird. JSON just works.

Settings can only be set/retrieved one server at a time.

The CLI can only discover subclasses through introspection if the module is loaded, so `__init__.py` imports all the user config modules, if we add more it needs to be updated.

# User Interface

This user interface fucking sucks, and I will not defend it. I am insufficiently skilled/knowledgeable with HTML/CSS/Javascript/React to do a good job at it.

The UI is essentially the same as auto settings, but it won't let you save to the database if the model doesn't validate.

Error messages are not *great*, but it tells you if you are missing mandatory fields, and it gives back Pydantic error messages, it's at least no worse (and better because of validation) than setting these through config files.

Each setting has a manual example with comments similar to `config.yml` and `.env`, and the Pydantic schema is (although not very readable) shown at the bottom.

# File renaming

Some files were renamed (modules that mirrored the package name) because it was conflicting with my ability to use `rcon/cli.py` due to local packages masking other ones.

I'm not sure if there is a better way to handle this: https://discord.com/channels/685692524442026020/698857000804548629/1153317949181014047

# Caching

It's tough to handle redis caching/invalidation because of circular imports. Hitting the database shouldn't be much of a performance impact, but generally speaking saving `rcon.user_config.utils.get_user_config` is cached, and setting any user config clears this.

# Refreshing Settings

To the greatest extent possible I tried to make it so changing settings would impact things immediately:

* Auto broadcasts refresh after a complete message cycle
* Auto vote kick refreshes after every loop (30 seconds)
* Auto mods refresh after every loop (5 seconds)
* Ban TK on connect refreshes every TK
* Camera notification refreshes every camera enter
* Expired VIPs refreshes every loop (hourly by default)
* GTX name change refreshes every request
* Log line webhooks refresh on service startup (log event loop) and every 5 minutes
* Name kicks refreshes every player connection
* RCON connection settings, only on restart?
* RCON server settings, used all over the place, best to restart to be sure
* Real VIP, refreshes every player connect/disconnect
* Scorebot refreshes every loop (1 second)
* Standard messages refreshes every request
* Steam refreshed each use, but the individual requests are * cached
* VAC/game bans, refreshes every player connect
* Vote map refreshes throughout
* Web hooks are cached (5 minute) and refreshed on log event loop start

## Discord Auditing

Because some of the models are really verbose, only changed values are sent to the discord audit webhook.

It's not *strictly* only changed keys, the whole nested object is included when a nested object key is changed.

Sensitive stuff (it has to be manually updated as new models are added) is masked (webhook URLs, etc) when it's sent to Discord.

## .env changes

The following environment variables no longer exist and were moved to various different user settings.

Leaving them in is meaningless (and does allow users to downgrade), they just won't affect anything if people leave them in.

```
SERVER_SHORT_NAME
RCONWEB_SERVER_URL
DISCORD_WEBHOOK_AUDIT_LOG
DISCORD_CHAT_WEBHOOK
DISCORD_CHAT_WEBHOOK_ALLOW_MENTIONS
DISCORD_PING_TRIGGER_WORDS
DISCORD_PING_TRIGGER_ROLES
DISCORD_PING_TRIGGER_WEBHOOK
DISCORD_KILLS_WEBHOOK
DISCORD_SEND_KILL_UPDATES
DISCORD_SEND_TEAM_KILL_UPDATES
```

## default_config.yml

I left all of the code to use the file (but removed all the values from it) so we can use it to rapidly protoype new features.

## Upgrading

* Backup your `.env`
* Backup your `config.yml` file(s)
* Backup your broadcasts
  * Without manually querying your database this will be unaccessible after upgrading
* Backup your standard messages (punitions, welcome, broadcasts)
  * Without manually querying your database this will be unaccessible after upgrading
* Backup your UI set webhooks (camera/watchlist)
  * Without manually querying your database this will be unaccessible after upgrading
* Perform a standard upgrade
* Set all of the settings, either through the UI, CLI, manual POST requests to the API, whatever.

# Downgrading
* If there's an issue, people should be able to downgrade without issue if they leave their old settings in `.env` and `config.yml`, extra stuff in `.env` is meaningless, it just isn't used.
  * Of course if they have set *new* settings, those wouldn't reflect if they downgraded

# General Concerns / TODO
* Auditing can leak secrets (such as your GSP username/password for the server rename feature) or webhooks since the audit trail view is behind a single permission
  * Sensitive stuff is masked from the discord audit trail
* You can no longer configure any of this stuff without the database container running like you could with the text files.
  * You can start just the redis/postgres containers and set it through the CLI of course, but it's less convenient.
\* A better UI would help end users considerably
* The endpoints support reseting models to default, but it should also be in the UI but guarded behind a confirm dialog


docker-compose.yml
```diff
@@ -187,31 +176,20 @@ services:
       HLL_DB_NAME: ${HLL_DB_NAME}
       HLL_DB_HOST: ${HLL_DB_HOST}
       HLL_DB_HOST_PORT: ${HLL_DB_HOST_PORT}
-      HLL_DB_URL: 'postgres://${HLL_DB_USER}:${HLL_DB_PASSWORD}@${HLL_DB_HOST}:${HLL_DB_HOST_PORT}/${HLL_DB_NAME}'
-      DISCORD_WEBHOOK_AUDIT_LOG: ${DISCORD_WEBHOOK_AUDIT_LOG_3}
+      HLL_DB_URL: "postgres://${HLL_DB_USER}:${HLL_DB_PASSWORD}@${HLL_DB_HOST}:${HLL_DB_HOST_PORT}/${HLL_DB_NAME}"
```
timraay, line 179: And here

timraay (approved):
I very quickly gave up on reviewing all of this. Seems much easier to just release it as a beta or something.

PR #321, opened 2023-10-21
Update docker-compose-common-components.yml

cemathey:
Remove single quotes around `HLL_DB_URL` so Docker will substitute the env variables properly.


timraay (approved):

timraay (commented):
What about these?
https://github.com/MarechJ/hll_rcon_tool/blob/master/docker-compose.yml#L115

PR #322, opened 2023-10-21
Update default.env

cemathey:
Remove single quotes, it doesn't play well with different versions of docker compose


default.env
```diff
@@ -32,27 +32,27 @@ TAGGED_VERSION=latest
 # Prohibited characters for HLL_DB_PASSWORD: %
 # You should wrap the password in single quotes: 'likethis' to avoid
```
timraay, line 33: Remove this :)

timraay (changes requested):

default.env
cemathey: Done

timraay (approved):

PR #323, opened 2023-10-24
Automod logs less verbose

ElGuillermo:
Do not log "no action" messages, as they tend to fill the logs quite quickly.


cemathey (approved):
Sounds fine to me, I'll probably make this user configurable on the UI settings branch.

PR #324, opened 2023-10-24
Always use local time zone in frontend

FlorianSW:
Instead of using mixed UTC and local times, the frontend should use one timezone consistently. As local seems to be the more natural thing, I choosed to change all time usages I could find to the local timezone.
This includes the log view, which might be the most visible change with this commit (the time in the log view was UTC so far).

### DIFF
--- a/rcongui/src/components/AuditLog/index.jsx
+++ b/rcongui/src/components/AuditLog/index.jsx
@@ -33,7 +33,7 @@ const AuditLogsTable = ({ auditLogs }) => {
       label: "Time",
       options: {
         customBodyRenderLite: (dataIndex) =>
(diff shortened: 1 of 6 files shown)


cemathey (approved):
lgtm, oddly on my end some (all?) of this stuff was already reflecting local time. Maybe some weird browser setting, much better to force it for consistency.

FlorianSW:
> lgtm, oddly on my end some (all?) of this stuff was already reflecting local time. Maybe some weird browser setting, much better to force it for consistency.

I mean, if you're in UTC, you wouldn't see a difference 😜

ElGuillermo (approved):

PR #333, opened 2023-10-27
Short votemap chat command

ElGuillermo:
I've added a "!vm" command, considering a lot of players seem to miss votes, either because they're too lazy to type in chat, either because they make typos in the "!votemap" command (ie : "!votmap", "!vote map", "votemap", etc).

Now you can either use the regular "!votemap" or the shortened "!vm" command.
Tested today on a full server with no apparent prob.

### DIFF
--- a/rcon/vote_map.py
+++ b/rcon/vote_map.py
@@ -349,7 +349,7 @@ class VoteMap:
         message = struct_log.get("sub_content", "").strip()
         config = VoteMapConfig()
         enabled = config.get_vote_enabled()
-        if not message.startswith(("!votemap", "!vm", "!VOTEMAP", "!VM")):
+        if not message.lower().startswith(("!votemap", "!vm")):
             return enabled
 
         steam_id_64_1 = struct_log["steam_id_64_1"]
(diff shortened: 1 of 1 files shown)


rcon/vote_map.py
```diff
@@ -349,7 +349,7 @@ def handle_vote_command(self, rcon, struct_log: StructuredLogLineType) -> bool:
         message = struct_log.get("sub_content", "").strip()
         config = VoteMapConfig()
         enabled = config.get_vote_enabled()
-        if not message.startswith("!votemap"):
+        if not message.startswith(("!votemap", "!vm", "!VOTEMAP", "!VM")):
```
timraay, line 352: Use `message.lower().startswith(..)` to avoid having to excplicitly state all cases

timraay (changes requested):

rcon/vote_map.py
ElGuillermo: Nice one. Changing it now !

timraay (approved):

PR #338, opened 2023-10-29
Create dependabot.yml

cemathey:
Let dependabot create PRs for bumping Python package versions.

Might need to enable this in the repo settings in addition to this file?

Should also hold on merging this until we've merged all the other version PRs so it doesn't spam us with irrelevant nonsense.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


ElGuillermo (approved):

cemathey:
> This one is quite handy, but : isn't it there some risk of compatibility issues where some new version of a script isn't backward compatible ? Does it apply updates automatically or does it just give advice about upgrading ?

It just creates pull requests when it sees new versions of things, but yes we should not blindly merge them without reading the release notes and checking for backwards incompatible things, and ideally running the tests and making a real build and testing it against a server.

PR #339, opened 2023-10-30
Bump redis and discord.py versions

cemathey:
* Bump `redis` to the newest version
* Remove `redistimeseries` as it is now a part of `redis`
* Bump `discord.py` to the newest version
  * Took a few minor changes because of differences between 1.x and 2.x
  * Removed a duplicate `requirements.txt` specification for the discord package

I tested a live build of this and verified that scorebot and (I only tested chat, but the rest should be identical) the webhook changes worked correctly.

I also tested the stats calculation off live builds, and while I did not get identical results, I think that's a larger issue.

I tried two identical (`v8.1.1`) builds connected to the same game server and saw differences there as well, I haven't looked into it but I think the difference is intermittent playerinfo commands failing.

Testing methodology was connecting both RCONs, allowing a complete match to cycle where the match start was *after* RCON connection and then downloading the `.csv` stats from both, sorting them and then `diff`ing the results.


ElGuillermo:
Playerinfo commands failing... Do we know why ?
Maybe we can find a way to circumvent it ? Or force a "retry until you got it" ?

cemathey:
> Playerinfo commands failing... Do we know why ? Maybe we can find a way to circumvent it ? Or force a "retry until you got it" ?

I forget off the top of my head, I think it is a combination of whitespace/pineapple names, and then sometimes the game server just telling us to fuck off for whatever reason.

ElGuillermo (approved):

PR #363, opened 2023-11-19
Pseudo with ending space warn or kick

ElGuillermo:
Gives the option to kick the "bad pseudo" players (20th char is a space), to avoid unjustified punitions of their squad mates bu noleader_automod.


cemathey:
I'll update this to work with UI settings soon.

We'll also want to update it to account for pineapple names (multi byte code point in the last position) https://discord.com/channels/685692524442026020/698857000804548629/1098181704595820554

rcon/hooks.py
```diff
@@ -359,61 +368,188 @@ def update_player_steaminfo_on_connect(rcon, struct_log, _, steam_id_64):
...
+    Only messaging the player should be cancelled if they disconnect early
+    Kicking players is non functional (RCON bug) so they're temp banned/pardoned
+    Temporary banning players doesn't create a timer
+    """
+    pts: list[tuple[RconInvalidNameActionType, Timer]] = pendingTimers.pop(
+        steam_id_64, []
+    )
+    for action, pt in pts:
+        # TODO: Can't kick pineapple names, don't cancel the thread if the ban removal is pending
+        if action != RconInvalidNameActionType.kick and pt.is_alive():
```
FlorianSW, line 548: this would result in timers kept, when a player disconnects on it's own with a pineapple name, right? 
However, I think this is not a big issue, the command will simply fail then which is fine.

FlorianSW (approved):

PR #378, opened 2023-11-24
Update autoMods.js

adevnylo:
Fixed the documentation for NoLeaderAutoMod, SeedingAutoMod, and NoSoloTankAutoMod. For the NoSoloTankAutoMod, I've referred to both #309 (for the comments) and #319 (for the current value names).

### DIFF
--- a/rcongui/src/components/UserSettings/autoMods.js
+++ b/rcongui/src/components/UserSettings/autoMods.js
@@ -185,6 +185,7 @@ export const NoLeaderAutoMod = ({
         
         /*
             Step number 1
+            
             If a squad has no officer all members are noted internally but no action is taken
             Since we sometimes receive wrong data from the game server we note those players
             and see if these squads still don't have an officer the next time we check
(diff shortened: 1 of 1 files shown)


rcongui/src/components/UserSettings/autoMods.js
```diff
@@ -362,7 +366,7 @@ export const SeedingAutoMod = ({
         {player_name}, {received_warnings}, {max_warnings}, {next_check_seconds} and {violation}
         are the allowed message parameters
     */
-    "warning_message": "Warning, {player_name}! You violate seeding rules on this server: {violation}\nYou will be punished after {max_warnings} warnings (you already received {received_warnings}), then kicked.\nNext check will happen automatically in {next_check_seconds}s.",
+    "warning_message": "Warning, {player_name}! You violate seeding rules on this server: {violation}\\nYou will be punished after {max_warnings} warnings (you already received {received_warnings}), then kicked.\\nNext check will happen automatically in {next_check_seconds}s.",
```
ElGuillermo, line 369: Not sure about adding extra "\" (?)

adevnylo: > Not sure about adding these extra backslashes (?)

If you go to `https://localhost:9010/#/settings/automod-seeding` and scroll to see the Documentation, you'll see that part of the messages will go on a new line because "`\n`" gets interpreted as a new line, not as part of the string.
In all the other instances, "`\n`" is escaped as well with "`\\n`".

cemathey (approved):
Thanks for the fix!

PR #379, opened 2023-11-25
Update seeding_rules.py

ElGuillermo:
Some `server_player_count` conditions weren't tested or wrongly set.

disallowed-roles :
- was enabled from 0 to `max_players` included
- `min_players` wasn't take in consideration :/

disallowed_weapons :
- was enabled if `server_player_count` is equal or greater than `min_players` (**good**)
- was disabled if greater or equal to `max_players` (**bad** : should be enabled if equal)

enforce-cap_fight:
- was enabled if `server_player_count` is equal or greater than `min_players` (**good**)
- was disabled if greater or equal to `max_players` (**bad** : should be enabled if equal)

### DIFF
--- a/rcon/automods/seeding_rules.py
+++ b/rcon/automods/seeding_rules.py
@@ -264,20 +264,43 @@ class SeedingRulesAutomod:
                 )
 
                 drc = self.config.disallowed_roles
-                if server_player_count >= drc.max_players:
+                if (
+                    server_player_count < drc.min_players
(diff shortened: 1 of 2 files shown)


cemathey:
I added some tests to confirm that we don't enforce below minimum player counts.

cemathey (approved):
looks good to me, all the tests pass and I manually tested both the cap fight/disallowed roles on our event server.

PR #386, opened 2023-11-29
Update miscellaneous.js

adevnylo:
Added (an initial) documentation for RconConnectionSettings. Need to look more into "max_open" and "max_idle".
With this, we have now reached parity with the old default_config.yml file.


cemathey (approved):
Thanks!

PR #392, opened 2023-12-04
Feature---send a message on connect (remind rules, announce event, give Discord url, etc)

ElGuillermo:
Any player that enters the server will receive a message.
Could be useful to display server rules, team Discord url, events reminders, etc.

Message can be different according to seed time or non-seed time (different rules apply).

We can imagine it may be extended in a near future to display personal stats, VIP expiration date, etc.


cemathey:
Haven't forgotten this, I'll help update it once we complete/merge the trigger words branch because it'll be able to use a lot of the same stuff.

ElGuillermo:
> Haven't forgotten this, I'll help update it once we complete/merge the trigger words branch because it'll be able to use a lot of the same stuff.

It may be time to close this... I think the idea is cool, but we should try to find some way to unit the different messages that pop on connection (info message on connect + seed alert + level alert)...

Players tend to close these too frequent messages without even reading them.

PR #394, opened 2023-12-06
Delete rcon/vips.py

ElGuillermo:
This file and the function defined in it aren't used anywhere (?)


FlorianSW:
was probably something that was used from the command line directly. However, I don't think that this is really needed anyway. If we would like to have such a feature (list of vips that were not seen for x days) then we should add that in a way that is usable for the majority of users.

PR #395, opened 2023-12-06
Update automod.py

ElGuillermo:
There is no need to throw an exception everytime aplayer can't be warned/punished/kicked. Punish fails are expected when the player isn't on map (already dead, in the lobby, etc)

### DIFF
--- a/rcon/automods/automod.py
+++ b/rcon/automods/automod.py
@@ -107,7 +107,7 @@ def _do_punitions(
                     aplayer.details.author,
                 )
         except (CommandFailedError, HLLServerError):
-            logger.exception("Failed to %s %s", repr(method), repr(aplayer))
+            # logger.exception("Failed to %s %s", repr(method), repr(aplayer))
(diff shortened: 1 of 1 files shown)


caranci:
The change will only log when `ActionMethod.PUNISH` fails but not `ActionMethod.KICK` fails.

Maybe being pedantic, but isn't `logger.exception()` just `logger.debug()` with exception information logged as well? I didn't think it throws exceptions.

ElGuillermo:
We may add a log output when any method fails, for sure :)
The exception doesn't halt the script, but it overflows the logs with not-so-useful data, as we already *know* why punishes fail a lot.

caranci:
> The exception doesn't halt the script, but it overflows the logs with not-so-useful data, as we already _know_ why punishes fail a lot.

I totally agree!

cemathey (approved):

FlorianSW (approved):

PR #396, opened 2023-12-08
Fix VIP uploads

cemathey:
* Correctly handle requests that contain form data but not a JSON body

I tested this, VIP upload/download works properly again and it doesn't affect other API requests.

### DIFF
--- a/rconweb/api/utils.py
+++ b/rconweb/api/utils.py
@@ -12,11 +12,15 @@ logger = getLogger(__name__)
 def _get_data(request: HttpRequest) -> QueryDict | dict[str, Any]:
     if request.method == "GET":
         return request.GET
-
-    # Don't silently swallow JSON parsing errors
-    # login_required decorator will return a reasonable API response on failure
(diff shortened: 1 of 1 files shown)


ElGuillermo (approved):

rconweb/api/utils.py
```diff
@@ -12,11 +12,15 @@
...
     if request.method == "GET":
         return request.GET
-
-    # Don't silently swallow JSON parsing errors
-    # login_required decorator will return a reasonable API response on failure
-    # endpoints that don't require login should not be accepting POST requests
-    return json.loads(request.body)
+    # This is only used (currently) by the async_upload_vips endpoint which doesn't
+    # include a parseable JSON body, but form data and it pulls the data itself
+    elif request.method == "POST" and request.FILES:
```
FlorianSW, line 17: wouldn't it make sense to also check for the content type of the request? Like:
```
if request.content_type == "application/json":
  return json.loads(request.body)
else:
  return {}
```

FlorianSW (approved):
One comment, though :)

PR #401, opened 2023-12-09
Reject wrong HTTP method and log missing/invalid content type

FlorianSW:


cemathey (approved):
Looks good to me without actually building/testing it

FlorianSW:
> Looks good to me without actually building/testing it

I built it and run quite some APIs to test (not all, though).

PR #408, opened 2023-12-13
Convert get_audit_logs endpoint to GET not POST

cemathey:
* Convert the `get_audit_logs` endpoint to require `GET` not `POST`
* Update the GUI to call it correctly

Built it locally and did some quick minimal testing and it filters correctly.

### DIFF
--- a/rconweb/api/audit_log.py
+++ b/rconweb/api/audit_log.py
@@ -91,7 +91,7 @@ def get_audit_logs_autocomplete(request):
 @csrf_exempt
 @login_required()
 @permission_required("api.can_view_audit_logs", raise_exception=True)
-@require_http_methods(['POST'])
+@require_http_methods(['GET'])
 @require_content_type()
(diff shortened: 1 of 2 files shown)


ElGuillermo (approved):

rconweb/api/audit_log.py
```diff
@@ -91,7 +91,7 @@ def get_audit_logs_autocomplete(request):
 @csrf_exempt
 @login_required()
 @permission_required("api.can_view_audit_logs", raise_exception=True)
-@require_http_methods(['POST'])
+@require_http_methods(['GET'])
 @require_content_type()
```
FlorianSW, line 95: when GET you should also remove the require_content_type decorator. GET requests should not have or rely on the body.

cemathey: Whoops, thanks dude. Fixed in #411

PR #416, opened 2023-12-22
Update backend to support windows store player IDs

cemathey:
* Update the kill/team kill log parsing regex to support windows store IDs
* Refactor the player info command result parsing for easier testing
* Update various steam utilities to only call the steam API if the player ID is an actual steam ID
* Add various tests

I did build this, run the tests (some of the tests fail but it's unrelated to this and should be failing on master as well), assign a steam API key and connect it to a server with actual players on it and let it run a bit and poked around, I believe this works correctly.

I am curious/suspect there might be a better way to do the filtering decorators and I am all ears.

I do have to fix some cache handling errors from #410 but you'll only see that issue if redis is unavailable, and docker desktop (where I almost exclusively do this stuff) likes to have random DNS lookup issues which is how I noticed.

### DIFF
--- a/rcon/rcon.py
+++ b/rcon/rcon.py
@@ -44,7 +44,7 @@ from rcon.utils import (
     ALL_ROLES,
     ALL_ROLES_KEY_INDEX_MAP,
     INDEFINITE_VIP_DATE,
-    _get_default_info_dict,
+    default_player_info_dict,
     get_server_number,
     parse_raw_player_info,
 )
@@ -204,7 +204,7 @@ class Rcon(ServerCtl):
             except Exception:
                 logger.error("Failed to get info for %s", futures[future])
                 fail_count += 1
-                player_data = _get_default_info_dict(futures[future][NAME])
+                player_data = default_player_info_dict(futures[future][NAME])
             player = futures[future]
             player.update(player_data)
             players_by_id[player[STEAMID]] = player
--- a/rcon/rcon.py
+++ b/rcon/rcon.py
@@ -44,7 +44,9 @@ from rcon.utils import (
     ALL_ROLES,
     ALL_ROLES_KEY_INDEX_MAP,
     INDEFINITE_VIP_DATE,
+    _get_default_info_dict,
     get_server_number,
+    parse_raw_player_info,
 )
 
 STEAMID = "steam_id_64"
@@ -105,9 +107,8 @@ class Rcon(ServerCtl):
     player_info_regexp = re.compile(r"(.*)\(((Allies)|(Axis))/(\d+)\)")
     log_time_regexp = re.compile(r".*\((\d+)\).*")
     connect_disconnect_pattern = re.compile(r"(.+) \((\d+)\)")
-    # Checking for steam ID length so people can't exploit it with a name like: short(Axis/123) ->
     kill_teamkill_pattern = re.compile(
-        r"(.*)\((?:Allies|Axis)\/(\d{17})\) -> (.*)\((?:Allies|Axis)\/(\d{17})\) with (.*)"
+        r"(.*)\((?:Allies|Axis)\/(.*)\) -> (.*)\((?:Allies|Axis)\/(.*)\) with (.*)"
     )
     camera_pattern = re.compile(r"\[(.*)\s{1}\((\d+)\)\] (.*)")
     teamswitch_pattern = re.compile(r"TEAMSWITCH\s(.*)\s\((.*\s>\s.*)\)")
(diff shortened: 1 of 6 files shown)


rcon/rcon.py
```diff
@@ -44,7 +44,9 @@
     ALL_ROLES,
     ALL_ROLES_KEY_INDEX_MAP,
     INDEFINITE_VIP_DATE,
+    _get_default_info_dict,
```
FlorianSW, line 47: when we're on that already, some suggestions:
- make this method non-private (remove the underscore)
- rename the method to something more meaningful, e.g.: `new_from_player_name`

what do you think?

rcon/rcon.py
```diff
@@ -44,7 +44,9 @@
...
     get_server_number,
+    parse_raw_player_info,
```
FlorianSW, line 49: to make it consistent with my suggestion from above, what about renaming this one to:
`new_from_player_info`? :)

rcon/rcon.py
```diff
@@ -105,9 +107,8 @@ class Rcon(ServerCtl):
     player_info_regexp = re.compile(r"(.*)\(((Allies)|(Axis))/(\d+)\)")
     log_time_regexp = re.compile(r".*\((\d+)\).*")
     connect_disconnect_pattern = re.compile(r"(.+) \((\d+)\)")
-    # Checking for steam ID length so people can't exploit it with a name like: short(Axis/123) ->
     kill_teamkill_pattern = re.compile(
-        r"(.*)\((?:Allies|Axis)\/(\d{17})\) -> (.*)\((?:Allies|Axis)\/(\d{17})\) with (.*)"
+        r"(.*)\((?:Allies|Axis)\/(.*)\) -> (.*)\((?:Allies|Axis)\/(.*)\) with (.*)"
```
FlorianSW, line 111: a bit greedy, we would probably be able to have discrete parts for either a numeric steam ID or a UUID (which seems to be the Windows Store ID). However, as we have no _guarantee_ that this is actually always a UUID, and given that the other possibilities of the whole line are expressed already, being greedy here is most likely no issue.

Scratch all what I wrote, I just was thinking out loud 😅

rcon/steam_utils.py
```diff
@@ -15,6 +17,39 @@
 logger = logging.getLogger(__name__)
 
 last_steam_api_key_warning = datetime.datetime.now()
+steam_id_64_pattern = re.compile(r"\d{17}")
+
+
+def is_steam_id_64(steam_id_64: str) -> bool:
```
FlorianSW, line 23: Would probably be more performant if we do not use regular expressions for that. A Steam ID right now is always only digits and always 17 characters long. Something like this should be alright:

```
if steam_id_64.isdigit() and len(steam_id_64) == 17:
    return True
return False
```

rcon/steam_utils.py
```diff
@@ -15,6 +17,39 @@
...
+    if re.match(steam_id_64_pattern, steam_id_64):
+        return True
+    return False
+
+
+def filter_steam_ids():
```
FlorianSW, line 29: I think for the time being, this is good enough for now. It abstracts away the explicit check if a player ID is a steam ID, however, for now that's fine.

I would wish that there will be _any_ way to get similar information about a player profile for a Windows Store version of HLL, but I doubt that we will get this in the near future or at all.

FlorianSW (changes requested):
apart from my comments, this looks good :)

I didn't look for any place where we explicitly expect a Steam ID, however, I _think_ that this looks good so far :)

rcon/steam_utils.py
cemathey: I've inquired but I doubt that we're ever going to have an API we can inquire against. It's not clear to me, but I believe these IDs are being generated by Team17 from some sort of windows store identifier :<

rcon/steam_utils.py
cemathey: Much smarter

rcon/rcon.py
cemathey: I'm partial to this name, it says exactly what it does.

I did update the docstring though which I forgot to do.

rcon/rcon.py
cemathey: Updated.

ElGuillermo (approved):

rcon/steam_utils.py
FlorianSW: > I believe these IDs are being generated by Team17 from some sort of windows store identifier

That's exactly what I got from the document shared by T17 as well 🙈

FlorianSW (approved):

PR #425, opened 2023-12-26
Feature automod no leader seed limit activation

ElGuillermo:
Add an option to disable the automod when the number of players is below a configurable number.


ElGuillermo:
Oups... always forget the tests :/ My bad : I've tested it on a live server with no apparent prob.

PR #431, opened 2023-12-27
Maint/docker health checks

cemathey:
* Add a Docker health check to the backend container (presence of the `pid` file gunicorn makes on startup)
* Require a healthy backend container for the frontend/supervisor containers to start
* Explicitly check for `HLL_DB_PASSWORD` to be set on startup and shut down with an error message if not (if the database hasn't been initialized, missing the password will cause the `postgres` container to fail the health check, but it will pass on subsequent startups if the database was ever seeded successfully)

When I originally added the health checks for redis/postgres I stupidly only set the supervisor container to check for it, but really we should check in the backend container and make everything else depend on a successful backend startup I think.

One downside is it is going to give scary looking error messages to people who don't have multiple servers configured even though there isn't any sort of actual problem.

Example where postgres won't start (no password set on initial setup):
image

Example with a healthy backend 1 but unhealthy (unconfigured backend 2 and 3):
image

Example with an unhealthy backend 1 and unconfigured backend 2 and 3:
image


ElGuillermo:
Sorry, I do not dare to approve this, as I'm pretty lost with anything docker related :/

FlorianSW:
> One downside is it is going to give scary looking error messages to people who don't have multiple servers configured even though there isn't any sort of actual problem.

What error message are you referring to? 🤔

ElGuillermo:
When you recreate/restart containers (?)
Sounds OK, though... We may add some explanation in the README to warn this is normal as long as there is only one game server configured.

cemathey:
> > One downside is it is going to give scary looking error messages to people who don't have multiple servers configured even though there isn't any sort of actual problem.
> 
> What error message are you referring to? 🤔

Image 2 in the PR shows what it looks like, right now on `master` and all the releases if the server isn't configured it'll just shut down, on this branch it will actively show you an error which I think might confuse some users, but it will also make it more apparent if they have configuration problems and are trying to use those containers.

Now that I think of it, I'd like to see how this output change would interact with ansible as I use that do updates.

caranci:
> > > One downside is it is going to give scary looking error messages to people who don't have multiple servers configured even though there isn't any sort of actual problem.
> > 
> > 
> > What error message are you referring to? 🤔
> 
> Image 2 in the PR shows what it looks like, right now on `master` and all the releases if the server isn't configured it'll just shut down, on this branch it will actively show you an error which I think might confuse some users, but it will also make it more apparent if they have configuration problems and are trying to use those containers.
> 
> Now that I think of it, I'd like to see how this output change would interact with ansible as I use that do updates.

I find the extraneous containers annoying so I edit the `docker-compose.yml` to only spin-up the containers that are needed for the hll servers that we have.

Would it be less scary for people to choose between pre-canned versions of the `docker-compose.yml` that match their situation?

ElGuillermo:
> I find the extraneous containers annoying so I edit the `docker-compose.yml` to only spin-up the containers that are needed for the hll servers that we have.
> 
> Would it be less scary for people to choose between pre-canned versions of the `docker-compose.yml` that match their situation?

That would add files and a configuration step, though. Now it's working OK without any need to change anything.

Users that can rent a VPS and pull a github project on it shouldn't be scared of some warning messages, as the'yre supposed to have read the documentation, where it's been said "these messages are OK if you're not running more than one server".

... I would leave it as it is...

cemathey:
> > > > One downside is it is going to give scary looking error messages to people who don't have multiple servers configured even though there isn't any sort of actual problem.
> > > 
> > > 
> > > What error message are you referring to? 🤔
> > 
> > 
> > Image 2 in the PR shows what it looks like, right now on `master` and all the releases if the server isn't configured it'll just shut down, on this branch it will actively show you an error which I think might confuse some users, but it will also make it more apparent if they have configuration problems and are trying to use those containers.
> > Now that I think of it, I'd like to see how this output change would interact with ansible as I use that do updates.
> 
> I find the extraneous containers annoying so I edit the `docker-compose.yml` to only spin-up the containers that are needed for the hll servers that we have.
> 
> Would it be less scary for people to choose between pre-canned versions of the `docker-compose.yml` that match their situation?

We could certainly do that yeah but then we'd have more of a maintenance burden like elguillermo noted, so I'd prefer not. It's already tough for us that to remember to update the arm versions of stuff.

I think it's okay if we have it report as an error, but I'm going to do some more homework because I use ansible to update RCONs and if it's going to cause ansible playbooks to fail, I'll find some other way to do this.

caranci:
> > > > > One downside is it is going to give scary looking error messages to people who don't have multiple servers configured even though there isn't any sort of actual problem.
> > > > 
> > > > 
> > > > What error message are you referring to? 🤔
> > > 
> > > 
> > > Image 2 in the PR shows what it looks like, right now on `master` and all the releases if the server isn't configured it'll just shut down, on this branch it will actively show you an error which I think might confuse some users, but it will also make it more apparent if they have configuration problems and are trying to use those containers.
> > > Now that I think of it, I'd like to see how this output change would interact with ansible as I use that do updates.
> > 
> > 
> > I find the extraneous containers annoying so I edit the `docker-compose.yml` to only spin-up the containers that are needed for the hll servers that we have.
> > Would it be less scary for people to choose between pre-canned versions of the `docker-compose.yml` that match their situation?
> 
> We could certainly do that yeah but then we'd have more of a maintenance burden like elguillermo noted, so I'd prefer not. It's already tough for us that to remember to update the arm versions of stuff.
> 
> I think it's okay if we have it report as an error, but I'm going to do some more homework because I use ansible to update RCONs and if it's going to cause ansible playbooks to fail, I'll find some other way to do this.

NP. Just a thought.

FlorianSW (approved):
thx for the additional screenshots explaining what error message you talked about. Tbh: I think the docker-compose.yml file in the repo should be used as a quick-start or reference file, only. If someone doesn't want to have the exact same setup as proposed in the example, they should really create their own version of the compose file. I do the same, for the very same reason of not having additional containers start up, which I do nnot have configured.

PR #432, opened 2023-12-28
Hotfix/timezones

cemathey:
Python 3.12 deprecates `datetime.datetime.utcnow` and `datetime.datetime.utcfromtimestamp` which return naive (no time zone) datetimes (https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated)

I don't think we actually *need* to do anything because Python takes forever (literally years) to actually remove anything that is deprecated but the works done anyway. From my non exhaustive experimenting I don't think an actual real problem exists with the wrong timestamps being persisted to the database even when run from a system with the timezone set to something besides UTC.

Our postgres configuration is set up with the time zone set to UTC (which is good) and all of the data types for datetimes should be `timestamp without time zone`. I dumped the schema and looked, Django has some stuff which I'm not touching, and the `audit_log` and `player_vip` tables had the time zone specified (removed).

This should mean that regardless of how we've generated a timestamp (unless we've truly screwed up somehow) means that all datetimes that are going into the database are either already UTC and being converted to UTC, or a local timezone (as configured on the system the user is hosting the RCON on) that is then being converted to UTC when it's persisted in the database.

All game log absolute timestamps are UTC as they come out of the game server logs.

Ideally every datetime we work with is UTC and anything sent into or from the API or into or from the database is UTC and then the end user (front end, API users, whatever) converts it to a localized time when they display it to the user if desired.

We have some unreleased stuff `rcon.server_stats` which I've never really looked at but would also maybe require updates if we release it.

* Converts every instance of `datetime.datetime.utcnow` and `datetime.datetime.utcfromtimestamp` to `.now()` or `fromtimestamp()` with the timezone set to UTC
* Alter `audit_log` and `player_vip` tables to remove the time zone
* Adds a `TypeDecorator` to force sqlalchemy to add the UTC time zone to every datetime retrieved from the database, and strip the timezone before persisting (UTC -> UTC so nothing changes)
* Adds a time zone to every naive datetime to allow comparisons (raises a `TypeError: cant compare offset-naive and offset-aware datetimes` if you ever try to compare them otherwise)
* Removes some deprecated vote map/map recording stuff
* The `get_map_scoreboard` (and possibly other endpoints) now appends the timezone information as part of the result
* Updates the frontend to deal with the `get_map_scoreboard` endpoint that appends the timezone information

I did build this and connect it to our game server with actual people playing on it, let it record sessions, games, etc. and grep'd the logs until I had squashed all of the issues I could find with it between naive/timezone aware datetimes.


ElGuillermo (approved):

cemathey:
I'm going to hold on merging this until we have more consensus.

cemathey:
This is out of date and I'm not sure we even want to do this.

PR #435, opened 2023-12-31
Allow numbers in admin pings

cemathey:
Allow numbers in admin ping words.


ElGuillermo:
We also may add backslahes "/" and exclamation marks "!" as allowed chars, as a lot of players use them by default when asking for admin or any other trigger command.

ElGuillermo (approved):

cemathey:
> We also may add backslahes "/" and exclamation marks "!" as allowed chars, as a lot of players use them by default when asking for admin or any other trigger command.

I forgot about this but I'll add that soon, just want to get the windows store release out.

PR #448, opened 2024-01-07
Hotfix/steam db enrichment

cemathey:
* Update `enrich_db_users` log messages to be more clear
* Exit if a steam API key is not set as it's mandatory
* Reduce chunk size (number of profiles fetched per steam API request) from `100` to `25` to try to reduce the chance of steam rate limiting
* Don't fetch windows store ID players from the database for enrichment
* Add some tests to make sure the steam API decorators are filtering windows store IDs properly

### DIFF
--- a/rcon/steam_utils.py
+++ b/rcon/steam_utils.py
@@ -1,16 +1,19 @@
 import datetime
 import logging
 import math
+import sys
 import time
+from dataclasses import dataclass
 from functools import wraps
 from typing import Any, Mapping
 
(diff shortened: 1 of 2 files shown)


ElGuillermo (approved):

ElGuillermo (approved):

cemathey:
Working on a more robust steam API change in another branch.

Plus I misunderstood the chunk size we want to pull as many steam ID results from a single API request as we can

PR #452, opened 2024-01-10
Update default.env

ElGuillermo:
Adding a comment to tell the '$' sign is forbidden in RCONWEB_API_SECRET


cemathey:
#453 allows this now, but we might want to update `.env` and the README to add some more warnings in general because `$` can cause Docker to try to apply parameter expansion to environment variables

PR #454, opened 2024-01-12
Refactor steam API usage to try to reduce rate limiting

cemathey:
It's hard to say for sure since this involves external services, but I think this should cut down or maybe even eliminate users being rate limited by the Steam API. Previously, clearing the redis cache could also cause CRCON to hammer the steam API a bit since it would call it every time `get_players` was called (very often) until the values were re-cached, this avoids that.

I did build this, let it run, did quite a bit of poking around/manual testing and I don't believe I've changed the API results or broken anything.

This should cut down how often the steam API is called, but if a lot of `on_connected` hooks are processed at once (for instance you reconnect your CRCON to a game server that has been offline and it reprocesses a lot of old log lines at once), and of course rate limits will be shared across a steam API key if it's used for more than one game server, so there is still improvements we could make.

tl;dr stop pulling info from the steam API so much, rely on our local database and update steam info when players connect (critical for VAC/game ban checks) and make `enrich_db_users` work better.

* Update the `on_connected` hook to allow you to specify its order (so we can force the steam API calls to be made when players connected, before we check for VAC bans
* Only call the steam API to fetch profiles/bans when a player connects, or when `enrich_db_users` is called (it's on a `cron` job to run daily)
  * Previously we would make API calls based on `get_players` calls which could happen frequently and result in many overlapping pulls because it was cached based on the steam ID list it was called with which would vary any time players connected/disconnected
* Add some miscellaneous typing
* Add a singleton method (like for the redis pool, sqlalchemy engine, etc.) so per interpreter we only maintain a single initialized connection to the Steam API
* Rename `rcon.steam_util` functions to make it more clear which calls will touch the steam API (`fetch_` functions) and which ones simply query our local database (`get_` functions)
* Rename `rcon.steam_util` functions to make it more clear which ones should be called with a single steam ID and which ones should be called with multiple steam IDs (I found the singular `s` names hard to distinguish at a glance
* Update `enrich_db_users` query to identify more records that are missing information and need to be fetched
* Update `enrich_db_users` logging statements to provide more/clearer information
* Sort/remove duplicate steam IDs for `filter_steam_ids` to cause more cache hits (the redis cache would consider `[1,2,3]` and `[3,2,1]` to be different cache values even though they produce identical API results for our purposes
* Reduce TTL caches to 12 hours, we don't want to make excessive calls but this means each player should only be updated once a day, without skipping multiple days if a player tends to play at the same time each day

### DIFF
--- a/rcon/automods/automod.py
+++ b/rcon/automods/automod.py
@@ -6,6 +6,7 @@ from typing import List
 from pydantic import HttpUrl
 from redis.client import Redis
 
+import rcon.game_logs
 from rcon.automods.level_thresholds import LevelThresholdsAutomod
 from rcon.automods.models import ActionMethod, PunishPlayer, PunitionsToApply
 from rcon.automods.no_leader import NoLeaderAutomod
@@ -14,7 +15,6 @@ from rcon.automods.seeding_rules import SeedingRulesAutomod
 from rcon.cache_utils import get_redis_client
 from rcon.commands import CommandFailedError, HLLServerError
 from rcon.discord import send_to_discord_audit
-from rcon.game_logs import on_connected, on_kill
 from rcon.hooks import inject_player_ids
 from rcon.rcon import Rcon, get_rcon
 from rcon.types import StructuredLogLineType
@@ -190,7 +190,7 @@ def audit(discord_webhook_url: HttpUrl | None, msg: str, author: str):
         )
 
 
-@on_kill
+@rcon.game_logs.on_kill
 def on_kill(rcon: Rcon, log: StructuredLogLineType):
     red = get_redis_client()
     if not is_first_run_done(red):
@@ -216,7 +216,7 @@ def on_kill(rcon: Rcon, log: StructuredLogLineType):
 pendingTimers = {}
 
 
-@on_connected
+@rcon.game_logs.on_connected()
 @inject_player_ids
 def on_connected(rcon: Rcon, _, name: str, steam_id_64: str):
     red = get_redis_client()
--- a/rcon/game_logs.py
+++ b/rcon/game_logs.py
@@ -120,7 +120,7 @@ def on_connected(insert_at: int | None = None):
     """Insert the given hook at `insert_at` position, or the end"""
 
     def wrapper(func):
-        if insert_at:
+        if isinstance(insert_at, int):
             HOOKS[AllLogTypes.connected.value].insert(insert_at, func)
         else:
             HOOKS[AllLogTypes.connected.value].append(func)
--- a/rcon/game_logs.py
+++ b/rcon/game_logs.py
@@ -120,7 +120,7 @@ def on_connected(insert_at: int | None = None):
     """Insert the given hook at `insert_at` position, or the end"""
 
     def wrapper(func):
-        if insert_at and isinstance(insert_at, int):
+        if insert_at:
             HOOKS[AllLogTypes.connected.value].insert(insert_at, func)
         else:
             HOOKS[AllLogTypes.connected.value].append(func)
--- a/rcon/game_logs.py
+++ b/rcon/game_logs.py
@@ -116,9 +116,18 @@ def on_chat_allies(func):
     return func
 
 
-def on_connected(func):
-    HOOKS[AllLogTypes.connected.value].append(func)
-    return func
+def on_connected(insert_at: int | None = None):
+    """Insert the given hook at `insert_at` position, or the end"""
+
+    def wrapper(func):
+        if insert_at and isinstance(insert_at, int):
+            HOOKS[AllLogTypes.connected.value].insert(insert_at, func)
+        else:
+            HOOKS[AllLogTypes.connected.value].append(func)
+
+        return func
+
+    return wrapper
 
 
 def on_disconnected(func):
--- a/rcon/steam_utils.py
+++ b/rcon/steam_utils.py
@@ -10,6 +10,7 @@ import steam.exceptions
 from sqlalchemy import or_, select
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import Session, joinedload
+from sqlalchemy.sql.expression import func
 from steam.webapi import WebAPI
 
 from rcon.cache_utils import ttl_cache
@@ -143,9 +144,7 @@ def fetch_steam_player_summary_mult_players(
     for chunk in batched(steam_id_64s, page_size):
         chunk_steam_ids = ",".join(chunk)
         try:
-            logger.info(
-                "Fetching player summaries for %s steam IDs", len(chunk_steam_ids)
-            )
+            logger.info("Fetching player summaries for %s steam IDs", len(chunk))
             raw_result = api.ISteamUser.GetPlayerSummaries(steamids=chunk_steam_ids)
             chunk_profiles: list[SteamPlayerSummaryType] = raw_result["response"][
                 "players"
@@ -186,7 +185,7 @@ def fetch_steam_bans_mult_players(
     try:
         for chunk in batched(steam_id_64s, page_size):
             chunk_steam_ids = ",".join(chunk)
-            logger.info("Fetching player bans for %s steam IDs", len(chunk_steam_ids))
+            logger.info("Fetching player bans for %s steam IDs", len(chunk))
             raw_result = api.ISteamUser.GetPlayerBans(  # type: ignore
                 steamids=chunk_steam_ids
             )
@@ -358,10 +357,14 @@ def enrich_db_users(chunk_size=100, update_from_days_old=30):
     with enter_session() as sess:
         # Missing JSONB records are surprisingly difficult to identify depending
         # on how they've been persisted but JSONB.NULL should cover the different cases
-        query = (
-            sess.query(PlayerSteamID)
+        # This query can also return many thousands of results, using stream_results and
+        # yield_per doesn't fetch all of the results at once, but pages through them
+        stmt = (
+            select(PlayerSteamID)
+            .execution_options(stream_results=True, yield_per=chunk_size)
             .outerjoin(SteamInfo)
-            .filter(
+            .where(func.length(PlayerSteamID.steam_id_64) == 17)
+            .where(
                 or_(
                     PlayerSteamID.steaminfo == None,
                     SteamInfo.updated <= max_age,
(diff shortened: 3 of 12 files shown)


ElGuillermo (approved):

rcon/steam_utils.py
```diff
@@ -2,35 +2,72 @@
...
+def get_steam_api_key() -> str | None:
+    """Return the servers steam API key if configured"""
+    global last_steam_api_key_warning
+
+    config = SteamUserConfig.load_from_db()
+    if config.api_key is None or config.api_key == "":
+        timestamp = datetime.datetime.now()
+        # Only log once every 5 minutes or it is super spammy
+        if (timestamp - last_steam_api_key_warning).total_seconds() > 300:
+            logger.warning("Steam API key is not set some features will be disabled.")
```
FlorianSW, line 46: do we need this at all? What about checking this at startup once and then never log it again? :D

rcon/steam_utils.py
```diff
@@ -67,234 +106,297 @@ def wrapper(steamd_id: str, *args, **kwargs):
     return decorator
 
 
-def get_steam_api_key() -> str | None:
-    global last_steam_api_key_warning
+@ttl_cache(60 * 60 * 12, cache_falsy=False, is_method=False)
```
FlorianSW, line 109: does this really need to be cacheable? I would say, as `fetch_steam_player_summary_mult_players` is cached already, we can safely omit this caching :)

rcon/steam_utils.py
```diff
@@ -67,234 +106,297 @@ def wrapper(steamd_id: str, *args, **kwargs):
...
-        # Only log once every 5 minutes or it is super spammy
-        if (timestamp - last_steam_api_key_warning).total_seconds() > 300:
-            logger.warning("Steam API key is not set some features will be disabled.")
-            last_steam_api_key_warning = timestamp
 
-    return config.api_key
+@ttl_cache(60 * 60 * 12, cache_falsy=False, is_method=False)
+@filter_steam_ids()
+def fetch_steam_player_summary_mult_players(
+    steam_id_64s: Iterable[str], page_size=100
```
FlorianSW, line 128: is there a need to have the page_size as a parameter? I have trouble to find use cases, where one would _force_ to have more requests than actually needed. And it more or less sounds like an implementation details to me, which callers should not have a way to manipulate.

FlorianSW (changes requested):
Some comments

rcon/steam_utils.py
caranci: The steam api key can be set (or invalidated) at any point via the config page now, not just at startup.
This is also the same check that exists in current code.

ElGuillermo:
> > * Only call the steam API to fetch profiles/bans when a player connects, or when `enrich_db_users` is called (it's on a `cron` job to run daily)
> 
> Strangely, the crontab runs `enrich_db_users` twice a day: at both `23:01` and `02:01`.

Yep. This is either dumb, either a strange trick I do no have any clue about.

rcon/automods/automod.py
```diff
@@ -190,7 +190,7 @@ def audit(discord_webhook_url: HttpUrl | None, msg: str, author: str):
         )
 
 
-@on_kill
+@rcon.game_logs.on_kill
```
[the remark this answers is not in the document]
cemathey: Just being more explicit because it shares the name with the `on_kill` function defined here, it confused me for a moment at first.

rcon/game_logs.py
```diff
@@ -116,9 +116,18 @@ def on_chat_allies(func):
...
 
 
-def on_connected(func):
-    HOOKS[AllLogTypes.connected.value].append(func)
-    return func
+def on_connected(insert_at: int | None = None):
+    """Insert the given hook at `insert_at` position, or the end"""
+
+    def wrapper(func):
+        if insert_at and isinstance(insert_at, int):
```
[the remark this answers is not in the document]
cemathey: Just being a bit paranoid in case someone ever tried to pass a non integer index, probably better to have it explicitly not work though.

rcon/steam_utils.py
cemathey: I moved it to a constant in the module, it wasn't actually passed in at all anywhere it was used, it sounds like the sort of thing that will probably never change since it's in the steam API.

rcon/steam_utils.py
cemathey: I concur

rcon/steam_utils.py
cemathey: Maybe we just bump up the time frame it logs for? I added it because it was spamming my logs while I was testing other stuff in other branches and found it quite infuriating.

I think caranci is right though since it can be set while CRCON is live, it's still useful,especially if we end up in a situation where we're accidentally resetting user config values again (although I think we fixed all of those)

cemathey:
Made those changes. I also made some additional changes to `enrich_db_users`

Surprisingly, I don't think this was ever working as intended, I think it was only ever pulling the same 100 IDs every loop, unless I'm reading this wrong because we weren't also applying an `offset` with the `limit`.

https://github.com/MarechJ/hll_rcon_tool/blob/e4d4d43e151c3f241625789b8134a984717389fe/rcon/steam_utils.py#L274C55-L274C60
https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.limit

I changed it to use sqlalchemy's partition and the execution options to pull `chunk_size` results at a time (defaulting to 100 because that's the max you can query in a single API call from steam) because some people have *huge* steam ID/steam info databases, for example with @ElGuillermo's database copy I used for testing I had something like ~18k rows to update the first time I ran it.

rcon/game_logs.py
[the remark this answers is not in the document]
FlorianSW: Yeah, I get that, what I think about is the first condition :) I would have assumed, that the isinstance check is enough.

rcon/steam_utils.py
FlorianSW: Hmm, in this case, I would either up the timer to an incredible high value (like one hour, or maybe even one day) or omit this log message at all. Configuring (or better, not doing so) a steam API key is an explicit action, I don't think why we would need that log message at all.

FlorianSW:
> > > * Only call the steam API to fetch profiles/bans when a player connects, or when `enrich_db_users` is called (it's on a `cron` job to run daily)
> > 
> > 
> > Strangely, the crontab runs `enrich_db_users` twice a day: at both `23:01` and `02:01`.
> 
> Yep. This is either dumb, either a strange trick I do no have any clue about.

It _might_ be to retry when steam API is down at this time, but even then it doesn't make that much sense. We _could_ remove the second invocation. Or, even better, invoking the script again should not do anything when there is no work to do. E.g. when all steam profiles are up-to-date.

FlorianSW:
> Surprisingly, I don't think this was ever working as intended, I think it was only ever pulling the same 100 IDs every loop, unless I'm reading this wrong because we weren't also applying an `offset` with the `limit`.
> 
> https://github.com/MarechJ/hll_rcon_tool/blob/e4d4d43e151c3f241625789b8134a984717389fe/rcon/steam_utils.py#L274C55-L274C60 https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.limit

It always fetched 100 steam profiles, which were _older_ than 30 days:
https://github.com/MarechJ/hll_rcon_tool/blob/e4d4d43e151c3f241625789b8134a984717389fe/rcon/steam_utils.py#L269

With that (as these profiles _should_ not be older than 100 days after the run), an offset if not really needed at all :)

rcon/game_logs.py
```diff
@@ -120,7 +120,7 @@ def on_connected(insert_at: int | None = None):
     """Insert the given hook at `insert_at` position, or the end"""
 
     def wrapper(func):
-        if insert_at and isinstance(insert_at, int):
+        if insert_at:
```
FlorianSW, line 123: I think `isinstance(insert_at, int)` is better, for the same reasons you pointed out. Sorry when my first comment wasn't clear enough :) I wanted to modify this statement to:

```
if isinstance(insert_at, int):
```
as this will check for None as well.

rcon/steam_utils.py
```diff
@@ -358,10 +357,14 @@ def enrich_db_users(chunk_size=100, update_from_days_old=30):
     with enter_session() as sess:
         # Missing JSONB records are surprisingly difficult to identify depending
         # on how they've been persisted but JSONB.NULL should cover the different cases
-        query = (
-            sess.query(PlayerSteamID)
+        # This query can also return many thousands of results, using stream_results and
+        # yield_per doesn't fetch all of the results at once, but pages through them
+        stmt = (
+            select(PlayerSteamID)
+            .execution_options(stream_results=True, yield_per=chunk_size)
```
FlorianSW, line 364: didn't know that that actually exists ❤️

FlorianSW (approved):
just one more thing, otherwise looks good! :)

rcon/game_logs.py
cemathey: I am bad at reading!

rcon/steam_utils.py
cemathey: I think the value comes from users who might not be aware they need to input a steam API key for them to take advantage of the feature but will spot the warning in the logs, I think an hour is fine though, it used to output constantly and I just wanted to cut down on that because it was drowning out everything else.

cemathey:
> > Surprisingly, I don't think this was ever working as intended, I think it was only ever pulling the same 100 IDs every loop, unless I'm reading this wrong because we weren't also applying an `offset` with the `limit`.
> > https://github.com/MarechJ/hll_rcon_tool/blob/e4d4d43e151c3f241625789b8134a984717389fe/rcon/steam_utils.py#L274C55-L274C60 https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.limit
> 
> It always fetched 100 steam profiles, which were _older_ than 30 days:
> 
> https://github.com/MarechJ/hll_rcon_tool/blob/e4d4d43e151c3f241625789b8134a984717389fe/rcon/steam_utils.py#L269
> 
> With that (as these profiles _should_ not be older than 100 days after the run), an offset if not really needed at all :)

I didn't realize until I was logging some debug output while working on it, but it was pulling the same 100 profiles (or whatever the first 100 records the query returned) for each page, because it was never `OFFSET`ing the limit.

cemathey:
> > > > * Only call the steam API to fetch profiles/bans when a player connects, or when `enrich_db_users` is called (it's on a `cron` job to run daily)
> > > 
> > > 
> > > Strangely, the crontab runs `enrich_db_users` twice a day: at both `23:01` and `02:01`.
> > 
> > 
> > Yep. This is either dumb, either a strange trick I do no have any clue about.
> 
> It _might_ be to retry when steam API is down at this time, but even then it doesn't make that much sense. We _could_ remove the second invocation. Or, even better, invoking the script again should not do anything when there is no work to do. E.g. when all steam profiles are up-to-date.

#456 updates it to run once a day, and with the query changes here even if we ran it more frequently it should still only query for records that need to be updating, so I don't think the frequency matters too much anymore. Hopefully by only making 2 API calls (1 for player profiles, 1 for player bans) every second we should never end up rate limited again.

cemathey:
I'm going to merge this for now because it's still an improvement over the previous system, but @caranci has some rate limiting code we can likely get in soon, and he also pointed out this runs in all the backend containers even though it only needs to run in one, so we should also fix that somehow.

caranci:
I think the crontab only has rotate logs and enrich db at the moment. Both of which should only run in one container...

PR #457, opened 2024-01-17
Feature automod level level1 bug protection enable/disable

ElGuillermo:
Adds a parameter to allow admins to enable/disable the "level 1 bug" protection.


cemathey:
I didn't actively test it, but looks good to me and the unit tests for automod all still pass. Wouldn't be a bad idea to add some tests to explicitly check this.

cemathey (approved):

ElGuillermo:
> I didn't actively test it, but looks good to me and the unit tests for automod all still pass. Wouldn't be a bad idea to add some tests to explicitly check this.

Yep. I've realised there is no test file for this automod :/

PR #460, opened 2024-01-17
Create black.yml

cemathey:
Taken without change from https://github.com/marketplace/actions/run-black-formatter

We still have some formatting changes (pre-commit not installed, different black versions maybe) leaking through sometimes. This should handle formatting automatically with a consistent version of `black` across all our pull requests.


cemathey:
Actually I think we can just close it, I think it's applied anyway.

PR #462, opened 2024-01-19
Update README.md

ElGuillermo:
strikethrough RCONWEB_SERVER_URL part, as a lot of users don't get that right

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Per server settings
     HLL_PORT
     HLL_PASSWORD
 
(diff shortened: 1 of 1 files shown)


caranci:
I'm not sure that strikethrough is what is needed here. I think most people interpret a strikethrough to mean the text has been deleted or removed and that that they should disregard the text.

ElGuillermo:
Yep. That's it : this parameter has disappeared, replaced by another one in UI config.
This doc has to be rewritten almost entirely :/

caranci:
Ack. I was thinking in terms that the server_url needs to be set... I somehow didn't clue-in to the fact it was the `.env` version of the setting.

cemathey:
I think we should just rewrite the install steps from scratch, if nobody beats me to it I'll spend some time doing it tomorrow.

PR #466, opened 2024-01-26
Add support for win-store playerid "profile" links.

caranci:
Add support for win-store playerid "profile" links.

Win-store playerids will link to an xboxgamertag.com search. This is a third-party site, but Microsoft does not have a player lookup web service that can be used. Note that gamertags are unique and cost money to change so gamertags should be more stable than steam player names.
- update `makeSteamProfileLink` to detect steam vs. win-store playerids + return appropriate url
- update live views to use `faWindows` icon for win-store playerids

*`faWindows` icon:*

Screenshot 2024-01-16 005330-copy

*Old screenshot (no win-store players the last few days):*

Screenshot 2024-01-16 000210-marked-up

### DIFF
--- a/rcongui/src/components/PlayerInfo/PlayerInfo.js
+++ b/rcongui/src/components/PlayerInfo/PlayerInfo.js
@@ -24,7 +24,7 @@ import MessageHistory from "../MessageHistory";
 import { toast } from "react-toastify";
 import { useParams } from "react-router-dom";
 import CollapseCard from "../collapseCard";
-import makeSteamProfileUrl from "../../utils/makeSteamProfileUrl";
+import makePlayerProfileUrl from "../../utils/makePlayerProfileUrl";
 
 // return a label for steam and windows ids types
 const getLinkLabel = (id) => {
@@ -317,7 +317,7 @@ const PlayerInfoFunc = ({ classes }) => {
                 </Grid>
                 <Grid item>
                   <Typography variant="h6">
-                    <Link href={makeSteamProfileUrl(steamId64, names[0]?.name)}>
+                    <Link href={makePlayerProfileUrl(steamId64, names[0]?.name)}>
                       {getLinkLabel(steamId64)} Profile
                     </Link>
                   </Typography>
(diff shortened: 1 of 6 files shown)


FlorianSW:
Wait, player tag changes cost money at xboxgamertag.com or with Microsoft? 🤔

ElGuillermo:
Microsoft...
As said here (old stuff, btw) : first change is free, then 10$/each
https://www.reddit.com/r/xbox/comments/qmtzi2/its_insane_that_you_have_to_pay_to_change_your/

caranci:
I think that xboxgamertag.com is using the Microsoft Xbox services API to give real-time information for gamertags since Microsoft doesn't have a service that does that.

My comment on the cost was more a comment on how I think that gamertags are much more stable that steam names since they actual money to change, after the one free change. Gamertags are also unique.

From the official Microsoft support faq (https://support.xbox.com/en-CA/help/account-profile/profile/gamertag-update-faq):
_Your first gamertag change is free, regardless of how many changes you made before the update in the past. Subsequent changes are the same as before ($9.99 USD)._

FlorianSW:
yeah, as they seem to be more static, searching for them actually makes much more sense (than steam player names).

rcongui/src/components/PlayerInfo/PlayerInfo.js
```diff
@@ -304,8 +317,8 @@ const PlayerInfoFunc = ({ classes }) => {
                 </Grid>
                 <Grid item>
                   <Typography variant="h6">
-                    <Link href={makeSteamProfileUrl(steamId64)}>
-                      Steam Profile
+                    <Link href={makeSteamProfileUrl(steamId64, names[0]?.name)}>
```
FlorianSW, line 320: could you rename the function to something more general (not mentionning steam in the name anymore :))

FlorianSW (changes requested):

rcongui/src/components/PlayerInfo/PlayerInfo.js
caranci: Thanks for calling out my laziness. :)

cemathey (approved):
Looks good to me but I am very javascript/react ignorant

FlorianSW (approved):

PR #470, opened 2024-02-04
Rudimentary highlighting for Logs view

caranci:
Add rudimentary highlighting for Logs view.
- added new classes for the logs lines types
- added colours for
  - Axis chat (red)
  - Allies chat (blue)
  - Team kill (magenta bold)
  - other messages are grey
- added switch to enable / disable the text hilighting.

Untitled


ElGuillermo:
Hey !

That should be "Highlight", not "Hilite" ;)

cemathey:
I'm going to defer on this one I'm very javascript ignorant, maybe @FlorianSW could take a look when he has time.

rcongui/src/components/LogsView/logs.js
```diff
@@ -21,6 +22,49 @@ import ListItemText from "@material-ui/core/ListItemText";
...
+        return classes.logsTeamKill;
+      case "kill":
+        return classes.logsKill;
+      case "teamswitch":
+        return classes.logsTeamSwitch;
+      case "disconnected":
+        return classes.logsDisconnected;
+      case "connected":
+        return classes.logsConnected;
+      default:
```
FlorianSW, line 61: default case not really needed here. However, that would only be important, if we want to change the var name of the default color (and don't want to do it at two places).

rcongui/src/components/LogsView/logs.js
```diff
@@ -132,6 +180,11 @@ class Logs extends React.Component {
     localStorage.setItem("logs_limit", limit);
   }
 
+  sethighlightLogs(highlightLogs) {
```
FlorianSW, line 183: shouldn't this be camel case: setHighlightLogs

rcongui/src/components/LogsView/logs.js
```diff
@@ -158,6 +212,17 @@ class Logs extends React.Component {
               <IconButton onClick={onFullScreen}>
                 {isFullScreen ? <FullscreenExitIcon /> : <FullscreenIcon />}
               </IconButton>
+            <FormControlLabel
```
FlorianSW, line 215: I'm not quite sure if we want to have this in the main UI tbh. It is a setting which just makes the whole UI more stuffed with things. What about moving it to the settings page (next to the theme switcher or something)?

rcongui/src/components/useStyles.js
```diff
@@ -64,10 +64,77 @@ const useStyles = makeStyles((theme) => {
...
+    // LIGHT_RED = 0xF93A2F
+    // BLUE = 0x006798
+    // LIGHT_BLUE = 0x0099E1
+    // GREEN = 0x07DA63
     logs: {
       margin: 0,
       whiteSpace: "pre-wrap",
     },
+    logsChatAxis: {
+      color: "#F93A2F",
```
FlorianSW, line 78: these colors work well with the light theme. Did you check all other themes as well? Would probably make sense to use a different color palette per theme. What do you thinnk?

rcongui/src/components/LogsView/logs.js
caranci: Yes, I missed this. The coding standard at work uses snake_case, making me somewhat allergic to camelCase.

rcongui/src/components/useStyles.js
caranci: They aren't great with dark mode, from what I recall.

I don't know how to correctly interact with the theming engine to switch between colours for dark vs. light mode.

I put the switch on the logs page to make it a bit obvious on how to turn it off if the colours weren't usable.

caranci: Actually the colors are better than I remembered in the dark theme. A slight tweak to make the blue lighter and it looks pretty good on both.
Screenshot 2024-02-14 085032
Screenshot 2024-02-14 085008

Hopefully I'll have time to merge the changes into the PR branch tonight.

rcongui/src/components/LogsView/logs.js
caranci: updated.

rcongui/src/components/LogsView/logs.js
caranci: Fixed.

rcongui/src/components/LogsView/logs.js
caranci: I don't have the patience to figure out how to move the switch to the other page. :(

I don't really like it here either, but it is only taking up existing whitespace. There really isn't any other frontend-UI setting other than the theme setting, which is global rather than for a particular widget so I don't see any existing pattern to follow.

And the switch is very discoverable when it is right on the page. :)

image

caranci:
@FlorianSW I'm not sure how to request a follow-up review.

rcongui/src/components/LogsView/logs.js
FlorianSW: Ok, then let's leave it there for now. There should be a better, more thought-through UI experience at some point, anyway :D

rcongui/src/components/useStyles.js
FlorianSW: ok, that looks better than I expected as well 👍

FlorianSW (approved):

PR #471, opened 2024-02-04
fix vip upload only accepts steam ids, not windows store ids

caranci:
Add a check so that VIP upload will accept both steam ids and windows store ids. Manually tested.

### DIFF
--- a/rcon/win_store_utils.py
+++ b/rcon/win_store_utils.py
@@ -1,0 +1,13 @@
+from typing import Final
+import re
+
+WIN_STORE_PLAYER_ID_REGEX: Final = (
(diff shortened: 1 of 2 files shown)


cemathey (approved):
Looks good to me, this is a really inelegant upload file format I think in the future we should transition to also supporting a JSON endpoint that won't rely on goofy parsing.

cemathey:
Ah I'm sorry, I meant as an option. We could auto detect on file upload, add a toggle or something.

It would be nice if there was an endpoint people could use that would accept the output of `api/get_vip_ids` 

I don't want to take this file away, I find it handy.

PR #492, opened 2024-02-19
add support for async discord webhooks

caranci:
Add support for async discord webhooks, allowing handling of discord rate limiting.

Support is via a singleton class that creates an asyncio event loop in a new thread.

Use `DiscordAsyncio().send_webhook(hook)` to send an existing `DiscordWebhook` asynchronously. The fields in `hook` will be copied to a new `AsyncDiscordWebhook` before being sent.

Currently only used in `send_to_discord_audit()`, mostly because it rate-limits hard when using https://github.com/cemathey/hll_seed_vip.

### DIFF
--- a/rcon/discord.py
+++ b/rcon/discord.py
@@ -112,8 +112,6 @@ def send_to_discord_audit(
         # we get a future, not a response, but i don't see code using the responses
         for hook in dh_webhooks:
             DiscordAsyncio().send_webhook(hook)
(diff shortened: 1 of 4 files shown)


cemathey (approved):
I'm a bit ignorant when it comes to this async stuff, but sounds great so we don't silently swallow rate limited stuff, even better that it retries!

Holding the merge for now in case anyone else has any other input on it.

PR #510, opened 2024-03-04
Use abu style maps and update votemap for skirmish

cemathey:
This is kind of a big one in terms of lines added/removed, apologies.

@timraay was kind enough to share some code from one of his private tools so now we have fully featured objects to represent maps and game modes instead of plain strings that come back from the game server.

There are now `Map`s which represent a base map (`carentan`, `driel`, etc.) and `Layer`s which represents the environment (day, night, etc.) and the game mode (and which team is attacking if it is offensive).

They're hashable, convert to strings/pretty names nicely, all that great stuff and assuming that T17 is consistent with their map naming going forward (lol) they should just work out of the box for future maps, the historical maps we have right now are already established.

I didn't update every place they're used internally and I tried to preserve API output for this hotfix for votemap stuff, but I think ultimately our goal should be to start returning these `Layer` objects as API output instead of map names so tool consumers don't have to maintain their own look up tables/parsing. I don't want to do that until after we've released this hotfix and the auto settings branch we have pending so we can do it on a major release without stopping people from getting bug fixes before they have time to update their tools.

I moved most of the stuff into its own `maps.py` file from `utils.py` and removed all the legacy map lookup stuff.

There is some miscellaneous typing that is fixed or added throughout the PR and some `black` formatting that aren't actual changes.

The UI (I did a shitty job but it works) is updated to reflect the vote map changes.

Vote map changes:
* Removed the offensive ratio option
* Removed the total number of selections options (it's calculated in the UI)
* Added options to specify the specific number of warfare, offensive and control skirmish maps to suggest
* Added options to prevent consecutive skirmish maps and to consider skirmish maps the same as warfare/offensive
* Vote map selections are now sorted to force consistent order across the border (when persisting in redis, messaging the user, etc.)

I did build this and made sure all of the vote map UI buttons work and correspond to the correct fields and I tested the vote map through the client to make sure I didn't break anything.

I have this running to collect some real games (for the scoreboard) to make sure I didn't accidentally change API output there so I'm going to hold on merging this until tomorrow if it gets approved early.


rcon/broadcast.py
```diff
@@ -36,22 +34,28 @@ def __str__(self):
...
     wins = winning_maps[:display_count]
     if display_count == 0:
         wins = winning_maps
-    return ", ".join(
-        f"{LONG_HUMAN_MAP_NAMES[m]} ({count} vote(s))" for m, count in wins
-    )
+
+    # Example warfare map: Carentan Warfare (2 vote(s))
+    # Example offensive map: Driel Off. AXIS (2 vote(s))
+    return ", ".join(f"{map_} ({num_votes} vote(s))" for map_, num_votes in wins)
```
timraay, line 58: `Layer.__str__` returns the layer's ID, not a prettified version. Use `map_.pretty()` instead.

rcon/broadcast.py
```diff
@@ -106,23 +110,24 @@ def format_by_line_length(possible_votes, max_length=50):
...
-    human_map_mod = (
-        NO_MOD_SHORT_HUMAN_MAP_NAMES if short_names else NO_MOD_LONG_HUMAN_MAP_NAMES
-    )
+
+    # 0: map 1, 1: map 2, etc.
     vote_dict = numbered_maps(selection)
+    # map 1: 0, map 2: 1, etc.
     maps_to_numbers = dict(zip(vote_dict.values(), vote_dict.keys()))
-    items = [f"[{k}] {human_map.get(v, v)}" for k, v in vote_dict.items()]
+    items = [f"[{k}] {v}" for k, v in vote_dict.items()]
```
timraay, line 130: `v.pretty()`

rcon/broadcast.py
```diff
@@ -176,8 +202,8 @@ def _get_vars(ctl):
     ]
 
     def get_next_map():
-        m = ctl.get_next_map()
-        return LONG_HUMAN_MAP_NAMES.get(m, m)
+        map_name: str = ctl.get_next_map()
+        return map_name
```
timraay, line 206: `map_name.pretty()`

rcon/broadcast.py
timraay: Also no idea how this doesn't fail the test you wrote for this? Perhaps I'm wrong somehow?

rcon/maps.py
```diff
@@ -0,0 +1,849 @@
...
+        else:
+            out += f" {self.gamemode.value.capitalize()}"
+        if self.environment != Environment.DAY:
+            out += f" ({self.environment.value})"
+        return out
+
+    @property
+    def opposite_side(self) -> Literal[Team.AXIS, Team.ALLIES] | None:
+        if self.gamemode == Gamemode.OFFENSIVE:
+            return Team.AXIS if self.attackers == Faction.GER else Team.ALLIES
```
timraay, line 188: I honestly don't remember if I changed `attackers` from `Faction` to `Team` or whether you did, but either way, now that it is `Team`, this will always return `Team.ALLIES`

timraay (changes requested):

rcon/broadcast.py
cemathey: No it fails, I think I wrote that part before I made some other changes? I don't remember but I definitely did not re-run all the tests before opening a PR (whoops lol).

Fixed.

rcon/broadcast.py
cemathey: Fixed

rcon/broadcast.py
cemathey: Fixed

rcon/maps.py
cemathey: I don't think it does actually, works as expected (but I do prefer your updated `attacking_faction` property so I'm going to swap it all over to that.

```python
@pytest.mark.parametrize(
    "team, expected",
    [
        (Team.ALLIES, Team.AXIS),
        (Team.AXIS, Team.ALLIES),
    ],
)
def test_get_opposite_side(team, expected):
    assert get_opposite_side(team) == expected
```

cemathey: Oh lol I am dumb and bad at reading, I was testing the utility function for it I made and not the property. I'm going to just update the property to use the properly working function which we use in the votemap settings anyway.

cemathey:
Okay I think I got everything updated, fixed all the tests and merged all our upstream changes (had to fix a merge conflict).

timraay (approved):

PR #511, opened 2024-03-04
Fix map rotation UI for skirmish maps

cemathey:
This is super quick and dirty and gross (but it does work) and we should come back to it after the API returns more useful map information (thanks Abu)

image

### DIFF
--- a/rcongui/src/components/MapManager/DraggableListItem.js
+++ b/rcongui/src/components/MapManager/DraggableListItem.js
@@ -29,9 +29,9 @@ const DraggableListItem = ({ item, index, onRemove }) => {
   const getLabels = (fullName) => {
     const labels = [];
 
-    if (fullName.search("offensive") !== -1 || fullName.search("off") !== -1) {
+    if (fullName.toLowerCase().includes("offensive") || fullName.toLowerCase().includes("off")) {
       labels.push("offensive");
-    } else if (fullName.toLowerCase().search('skirmish') !== -1) {
+    } else if (fullName.toLowerCase().includes('skirmish')) {
       labels.push('skirmish')
     }
     else {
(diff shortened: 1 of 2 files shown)


ElGuillermo (approved):

rcongui/src/components/MapManager/DraggableListItem.js
```diff
@@ -31,7 +31,10 @@ const DraggableListItem = ({ item, index, onRemove }) => {
 
     if (fullName.search("offensive") !== -1 || fullName.search("off") !== -1) {
       labels.push("offensive");
-    } else {
+    } else if (fullName.toLowerCase().search('skirmish') !== -1) {
```
FlorianSW, line 34: there is `String.prototype.includes` for that :D

rcongui/src/components/MapManager/DraggableListItem.js
```diff
@@ -43,7 +46,7 @@ const DraggableListItem = ({ item, index, onRemove }) => {
     if (fullName.toLowerCase().endsWith("ger")) {
       labels.push("axis");
     }
-    if (fullName.search("night") !== -1) {
+    if (fullName.toLowerCase().search("night") !== -1) {
```
FlorianSW, line 49: ah, copy&paste :D We should probably use includes here as well.

rcongui/src/components/Scoreboard/utils.js
```diff
@@ -1,12 +1,25 @@
 export const UNKNOWN_MAP_NAME = 'unknown';
 
+export const SKIRMISH_MAPS = [
```
FlorianSW, line 3: do we need this array? 🤔

cemathey: No, I think I added that when I was originally messing around with this and forgot all about it lol.

rcongui/src/components/MapManager/DraggableListItem.js
cemathey: Much better, thanks! I am so JS ignorant.

cemathey:
Updated built and tested, still works fine.

Thanks! Sorry I neglected to read the github emails and didn't realize there were comments.

PR #517, opened 2024-03-11
Update/simplify compose and docker files

cemathey:
This also includes the changes in PR #512 (I didn't discuss those changes here).

This takes advantage of the fact that all of our upstream Docker dependencies (`redis` and `postgres` and `node`) all provide multiple architecture images.

* Remove the architecture specific Dockerfiles (the standard one will pull appropriate upstream images)
* Remove the architecture specific Compose files
* Update `default.env` to no longer hard code `redis` and `postgres` image names in the compose files so people can manage their deployments easier
* Update `default.env` to include the server specific variables for up to 10 servers
* Add a `docker-build.yaml` file for users who want to locally build their images (I know you can do that with some of the other compose files, but this has no extraneous compose settings in it and is explicitly named so it should be more clear I think)
* Updates `docker-compose-common-components` to be architecture agnostic and to hold the container definitions for service wide containers (`redis`, `postgres` and `maintenance`)
* Removes `docker-compose.yml` from source control so that users can modify their deployments without `git` complaining about conflicting files (important for hands off updates using tools like `ansible`)
* Creates templates for a single game server (the most common scenario for starting users) and one with 10 game servers
  * I kept the server number comments in the single game server template because it makes it much easier to manage when using any sort of diff/select/compare tool (for instance visual studio code)
  * Users who want to add more servers just need to copy the relevant chunk out of the ten server template
* Reorganizes the docker compose template to make server 1 optional (large parts of the environment variables were defined for server 1) and to make the server specific portions as small/simple as possible
* Removes the env var check from `rcongui` `entrypoint.sh` because we now only create that container with the health checks if the backend is running so it's redundant.

I have a Raspberry Pi 5 (arm64) set up at home so I was able to test some of this stuff on that environment and amd64. It should just work on any common architecture (I explicitly checked that `redis`, `postgres` `node` and `python` images that we use all support `amd64`, `arm32` and `arm64` so that covers probably 95%+ of the environments people are trying to run this under.

This isn't ready to be released yet, I didn't want to waste my time updating the readme in case we don't want to merge this.

We also need to change how we're building images for docker hub, to do the multiple architecture build (I tested this on my fork of the repo and with my docker hub account) we would need to add a github action (which does require us to add docker credentials to the repo which only maresh has access to).

I did reach out to Maresh so hopefully if we're all happy with these changes we can get that worked out, I think we probably want to hold this until after we do a release with U14.8 bugs.

Also to be clear this is going to require users to make manual changes on the update (they will have to add the redis/postgres env vars from `default.env` as well as make a new compose file.

I think the minor upgrade steps are worth it to be able to provide native arm images on the repo, simplify building/deployment and provide users more options without having to manually edit their compose files and deal with `git` complaining about differences.


cemathey:
I updated this to also move the server number and redis database number that were hard coded in the compose files into env variables.

That removes all hard coded values out of the compose files I think.

The only complication is people will want to clear their cache as part of the upgrade process (I matched the DB number to the server number where as before the DB was zero indexed) so they aren't having old cache values.

That is as simple as:

```
docker compose down # Shut everything down so the cache isn't being updated
docker compose up -d redis  # Start redis
docker compose exec redis redis-cli flushall # Use the CLI to clear everything
```

ElGuillermo (approved):

cemathey:
Let's be sure to hold merging this until after we've done our U14.8 bug fix release, because this one will require users to take some manual steps.

The U14.8 bug fix release will be a standard pull image/checkout the release in git upgrade.

cemathey:
Oh and so I don't forget, this would also require README and Wiki updates, I didn't want to do that until we're happy with this though.

ElGuillermo (approved):

entrypoint.sh
```diff
@@ -15,19 +30,18 @@ then
...
-  ./manage.py makemigrations --no-input
-  ./manage.py migrate --noinput
   ./manage.py collectstatic --noinput
   # If DONT_SEED_ADMIN_USER is not set to any value
   if [[ -z "$DONT_SEED_ADMIN_USER" ]]
   then
     echo "from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'admin@example.com', 'admin') if not User.objects.filter(username='admin').first() else None" | python manage.py shell
   fi
   export LOGGING_FILENAME=api_$SERVER_NUMBER.log
+  # Successfully running gunicorn wlil create the pid file which is how Docker determines the container is healthy
```
caranci, line 44: `wlil` --> `will`

cemathey:
Okay I believe this is ready for prime time, pending Maresh having time to put my Docker Hub credentials into the repository secrets so that we can use the github actions.

ElGuillermo (approved):

PR #525, opened 2024-03-23
add option to only announce level automod to players impacted by curr…

caranci:
add option to only announce level automod to players impacted by current settings
- add only_announce_impacted_players option, defaults to False.
- automod.on_connected() gets detailed player info and passes to modules


rcon/automods/automod.py
```diff
@@ -231,11 +231,16 @@ def on_connected(rcon: Rcon, _, name: str, steam_id_64: str):
         logger.debug("No automod is enabled")
         return
 
+    # get detailed player info for use by on_connected_hook
+    detailed_player_info: GetDetailedPlayer = rcon.get_detailed_player_info(name)
```
cemathey, line 235: This will fail on anyone with a whitespace or pineapple name and not all mods require it (like weapon enforcement in seeding).

I think we want to try/catch this so people with bugged names don't cause the rest to fail?

rcon/automods/level_thresholds.py
```diff
@@ -69,7 +69,9 @@ def watch_state(self, team: str, squad_name: str):
                 pickle.dumps(watch_status)
             )
 
-    def on_connected(self, name: str, steam_id_64: str) -> PunitionsToApply:
+    def on_connected(
+        self, name: str, steam_id_64: str, detailed_player_info: GetDetailedPlayer
```
cemathey, line 73: Because detailed player info can fail, not sure if we want to support passing in a `None` `detailed_player_info` or an empty dict or something else.

rcon/automods/automod.py
caranci: Good, um, catch. I will take a look at it maybe returning None if it fails.

ElGuillermo (approved):

PR #565, opened 2024-04-29
Fix error in log stream when server is idle

ExcuseMi:
Fixes TypeError: cannot unpack non-iterable NoneType object when server has be idle in the log stream. 
The scenario occurs when last_seen is None and the tail returns None

image

### DIFF
--- a/rcon/game_logs.py
+++ b/rcon/game_logs.py
@@ -336,9 +336,10 @@ class LogStream:
         """Return a list of logs more recent than the last_seen ID"""
         try:
             if last_seen is None:
-                logs: list[tuple[StreamID, StructuredLogLineWithMetaData]] = [
-                    self.log_stream.tail()
-                ]
+                logs: list[tuple[StreamID, StructuredLogLineWithMetaData]] = []
+                tail_log: tuple[StreamID, StructuredLogLineWithMetaData] = self.log_stream.tail()
+                if tail_log:
+                    logs.append(tail_log)
             else:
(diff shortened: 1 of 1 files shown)


cemathey (approved):

rcon/game_logs.py
```diff
@@ -336,9 +336,10 @@ def logs_since(
...
-                logs: list[tuple[StreamID, StructuredLogLineWithMetaData]] = [
-                    self.log_stream.tail()
-                ]
+                logs: list[tuple[StreamID, StructuredLogLineWithMetaData]] = []
+                tail_log: tuple[StreamID, StructuredLogLineWithMetaData] = self.log_stream.tail()
+                if tail_log:
+                    logs.append(tail_log)
             else:
                 logs: list[tuple[StreamID, StructuredLogLineWithMetaData]] = (
                     self.log_stream.read(last_id=last_seen, block_ms=block_ms)
```
caranci, line 345: can log_stream.read also return None?

ExcuseMi: https://github.com/MarechJ/hll_rcon_tool/blob/1b5651b4fc810243a191d6f08e2211f947fcdf9e/rcon/utils.py#L163

It returns a list or a StreamNoElements exception (which is handled). Doesn't have the same problem as tail and head as these just want to return a tuple instead of list.

caranci: Cool. I didn't dig any deeper... just wondered. My C++ sensibilities are still bothered by the two `logs` variable "declarations" at a scope deeper than the usage. Oh python...

PR #566, opened 2024-04-29
Update README.md

ElGuillermo:
Rewrote mostly the "install" part to be more noob-friendly :)

Cleaned out the markdown errors.

+ little modifs : Docker 'compose' plugin is now mandatory as 'docker-compose' doesn't support 'include' instructions contained in compose files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -786,7 +786,7 @@ docker compose up -d --remove-orphans
 
 There's a public endpoint available to anybody without password on <http://yourVPSIP:7010/api/scoreboard>
 
-See [User Guide](USERGUIDE.md) for more information on how to use certain features of the app.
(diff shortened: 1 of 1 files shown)


cemathey (approved):
I skimmed it and it all looks good to me.

My only note would be in the Docker/Docker compose sections at the top, I would link to Dockers install instructions instead of telling people to `apt install` things because not everyone is necessarily on a Debian like OS and you have to take preliminary steps like adding Dockers repo to the apt repo.

ElGuillermo:
Or, as debian is often used, some warning like "these commands are meant to be used in a debian-based distro" ?

ElGuillermo:
Edited. Only "apt" left is for nano, but I wrote a note to say it's for debian-based distros.

PR #568, opened 2024-05-05
Salvage stats when player reconnects

FlorianSW:
Previously, when a player disconnects and then reconnects to the server the stats gained in the previous session where lost. The CRCON would always only save the 'higher' stats value to compensate this a bit, which was far from perfect.
With this commit, the CRCON will try to salvage the previously gained stats and combine them after the round ended.

### DIFF
--- a/rcon/game_logs.py
+++ b/rcon/game_logs.py
@@ -414,14 +414,20 @@ class LogLoop:
                 steam_id,
                 PlayerStat(
                     combat=player["combat"],
+                    p_combat=0,
                     offense=player["offense"],
+                    p_offense=0,
                     defense=player["defense"],
+                    p_defense=0,
                     support=player["support"],
+                    p_support=0,
                 ),
             )
             for stat in ["combat", "offense", "defense", "support"]:
-                if player[stat] > p[stat]:
-                    p[stat] = player[stat]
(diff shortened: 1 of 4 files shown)


FlorianSW:
> I'm fairly unfamiliar with the stats portion of CRCON, this won't double count if people disconnect/reconnect multiple times will it?

No, it should not. The stats are only "recovered", if the stats reported by the server are _less than_ the stats we have in our records. This can only happen once, when the user disconnects and reconnects. And it will always start from zero. What will happen is, if the player disconnects again, that the points they gained _during_ the time of the reconnect and second disconnect, are added up in the prevvious value.

E.g.:
- join server - 0 CE (previous CE: 0, displayed in Stats page: 0 CE)
- some kills etc - 50 CE (previous CE: 0, displayed in Stats page: 50 CE)
- disconnect
- reconnect - 0 CE (previous CE: 50, displayed in Stats page: 50 CE)
- some kills etc - 20 CE (previous CE: 50, displayed in Stats page: 70 CE)
- disconnect
- reconect - 0 CE (previous CE: 70, displayed in Stats page: 70 CE)
- some kills - 40 CE (previous CE: 70, displayed in Stats page: 110 CE)
- match end
What the stats page will show at the end are 110 CE for that player.

PR #588, opened 2024-06-20
Properly compress frontend assets

FlorianSW:
Before this commit, the "biggest" asset of the frontend, the javascript, was not compressed, even though compression was enabled. Enabling compression for the actually used javascript mime type variant reduces the initial payload that nneeds to transferred from 3.7 MB to "just" 1 MB.

Note: Sorry for the indentation changes in the diff :(


cemathey (approved):
I know nothing about nginx configs, but less bandwidth sounds great!

PR #589, opened 2024-06-20
Remove pandas to make build working again

FlorianSW:
Upgrading to the latest version of pandas would've worked, too. However, the server stats API is not finnisshed, not run on regular users workloads (except they intentionally enabled it) and was most likely not finished at all. No need in keeping it for much longer.


rcon/models.py
```diff
@@ -612,77 +612,6 @@ def to_dict(self) -> PlayerCommentType:
         }
 
 
-class ServerCount(Base):
```
FlorianSW, line 615: I wasn't sure if the corresponding database tables should be removed as well. However, keeping in mind that the feature itself might be interesting, I didn't want to add pointless migrations (just to add them in a later update). The tables should be empty anyway.

rconweb/api/models.py
```diff
@@ -225,7 +225,6 @@ class Meta:
                 "Can view the amount of time left in the round",
             ),
             ("can_view_server_name", "Can view the server name"),
-            ("can_view_server_stats", "Can view the get_server_stats endpoint"),
```
FlorianSW, line 228: I don't know enough about the permissions in Django tbh :( Do I need to do anything else here to make this change flawless? Like, will the permission automatically be removed from the groups and users that own them right now?

rcon/models.py
cemathey: I agree we should hang onto them in case we use them later. Some (very old) installs will have some computed data in there anyway that could be used in the future.

rconweb/api/models.py
cemathey: I'd probably just leave the permissions in case we want to use them later, they don't really hurt anything.

But this `models.py` file needs to be updated and then we'd need to modify migration 2 or 3 (I forget) which creates the default groups

FlorianSW: > I'd probably just leave the permissions in case we want to use them later, they don't really hurt anything.

I don't think that this would be a very good thing. People might misinterpret the permission, especially given when it does not do anything :/

> But this models.py file needs to be updated and then we'd need to modify migration 2 or 3 (I forget) which creates the default groups

👍

cemathey:
This got resolved on `master` with a change to `requirements.txt` to pin `numpy`; do we want to keep this or close it?

FlorianSW:
I mean, pandas is a big dependency, not sure how useful it is, if we don't actually use it 😅

FlorianSW:
I'll look into them after my event I play today 👍

FlorianSW:
@cemathey rebased the PR, could you take another look? ❤️

cemathey (approved):
lgtm and built and launched without issue

PR #590, opened 2024-06-22
Persist sscore/result of round for Scoreboard

FlorianSW:
In addition to the player stats, the map and the game duration, the result at the end is the most interesting information of a specific map. This commit adds persisting this information for future maps as well as showing it on the scoreboard.

I didn't take too much effort for a design of that, as I'm not sure how this will all unfold with the frontend rework :)
image


cemathey:
Sorry about that, been out of town. That info would be fantastic.

Would you mind if we hold on this change until we get the banlist release out? That one has database migrations as well and that one will be harder to fix merge conflicts on.

FlorianSW:
> Would you mind if we hold on this change until we get the banlist release out? That one has database migrations as well and that one will be harder to fix merge conflicts on.

Sure :)

cemathey (approved):

PR #609, opened 2024-07-26
Bugfix: Check if API Key is used and valid

LordofAgents:
Hello,
currently the CRCON does not check when using the Stats API if an API key is used and the API is locked for the public in the CRCON settings.
This leads to "You must be logged in to use this" errors.

This PR introduces that the API key is also checked. I have used the source code of "login_required" for this. 
Is it perhaps smarter to outsource the code to an extra function? But because of two occurrences a separate function seemed a bit stupid to me.

I am open to feedback!

Thank you for this amazing software!

### DIFF
--- a/rconweb/api/auth.py
+++ b/rconweb/api/auth.py
@@ -248,7 +248,7 @@ def stats_login_required(func):
 
     @wraps(func)
     def wrapper(request, *args, **kwargs):
-        request = check_api_key(request)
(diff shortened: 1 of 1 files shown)


LordofAgents:
I have placed the code in question in an extra function. Now it should be fine and hopefully good to read.

cemathey (approved):

PR #621, opened 2024-08-05
Fix typo: Change "commmand" to "command

LordofAgents:
Hey, 
I found a typo in the API :(
The role "commmand" should be "command".
If you/we already do breaking changes, we should fix that too.

Thank you for looking into it!


LordofAgents:
Please ignore the last two commits, wrong branch xD

cemathey (approved):

PR #623, opened 2024-08-06
UI: Blacklist dialog for GameView and PlayerView

Dorfieeee:
Implemented dialog windows for adding players to Blacklists directly from `GameView` and `LiveView`

### GameView
- A user is able to blacklist all selected players

### LiveView
- A user is able to blacklist a single player

### BlacklistDialog
_This is a subject to change as I did not fully understand the initial idea behind the helper text content. I hope I got it right._
- Updated "blacklist form input helper text" by following conditions
  - `blacklist.servers === []` -> _"This blacklist does not affect any servers!"_
  -  `blacklist.servers === [s1, s2, ...] & current_server NOT IN blacklist.servers` -> _"This blacklist DOES NOT affect THIS server! Affected servers: [s1, s2, ...]"_

Example: Multiple players selected
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Compose files are templated now
 compose.yaml
-
+frontend/
 db_data/
 .env
 broadcasts.txt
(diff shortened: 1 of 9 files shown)


timraay:
> This is a subject to change as I did not fully understand the initial idea behind the helper text content. I hope I got it right.
The idea of showing that the blacklist "does not affect any servers" was to warn the user that blacklisting this player would have no effect. Which, honestly, isn't super helpful as you will rarely have any blacklists that do not target any servers.

What I imagine would be most useful is if it warned you when the blacklist does not target the current server. Say, there's a player on server 1 that I want to blacklist, but the blacklist I've selected only bans players from server 2. I'm not entirely sure why I did not implement that originally. Probably forgot or overlooked, but there may also be a more technical reason for it?

Dorfieeee:
Code updated & issue comment edited.
The dialog itself now fetches data about the current server from `api/get_status`.
If the fetch fails, the user gets a warning about the failure and that the selected blacklist may not affect the current server followed by list of affected servers.

PR #624, opened 2024-08-06
UI: VIP Dialog for the LiveView

Dorfieeee:
!!! This commit is dependant on the previous Blacklist Dialog commit !!!

### LiveView
- A user should be able to add VIP to a single player


Dorfieeee:
I guess branching from a not yet merged branch creates a mess...

cemathey:
Closing this only because we don't want to merge it into master yet, I don't think I have access to do so because it's from your fork, but could you re-open it with it merging into feat/banlists ?

PR #626, opened 2024-08-07
UI: VIP Dialog for the LiveView

Dorfieeee:
### LiveView
- A user should be able to add/remove VIP to/from a single player


Dorfieeee:
When I created this PR it showed me merge to the banlists branch but it's trying to merge to master again. How's that?

PR #627, opened 2024-08-09
UI: VIP Dialog design & better management

Dorfieeee:
### Overall
- More compact and cleaner design.
- A user cannot set `expiry_date < now`.
- "Help to join!" button to set vip for 15 minutes.
image

### Player History View
- [old] - When a player has vip, the star button only allows a user to remove it.
- [new] - When a player has vip, the star button now opens the Vip dialog.
  - A user may want to only modify the length of the VIP so he does not need to remove it first and add it again with a new date.
image
image

### Blacklist Dialog
- A user can now select a reason from shared/local messages.
image

### Blacklist View
- FIX: Crashing app due to reading undefined/nullish values
- Added Skeleton UI to signal loading state
image


cemathey:
Thanks so much for the work on the UI! I'm truly helpless at that stuff.

Please see the Discord channel for a few comments on this.

PR #634, opened 2024-08-21
HLL U15.2 support - Game layout change & new maps

timraay:


rcon/commands.py
```diff
@@ -719,6 +719,18 @@ def get_gamestate(self) -> list[str]:
...
 
+    def get_objective_row(self, row: int):
+        if not (0 <= row <= 4):
+            raise ValueError('Row must be between 0 and 4')
+        return self._get_list(f'get objectiverow_{row}')
+
+    def set_game_layout(self, objectives: Sequence[str]):
+        if len(objectives) != 5:
+            raise ValueError("5 objectives must be provided")
+        return (
```
FlorianSW, line 730: it might be a lot of commands here, but it might be worth doing a get_objective_row call for each row and check if the provided capture points _actually_ exists for the current map? 🤔

Especially as we cache the get objectiverow call for up to 10 seconds, which might then even be a new map already.

=> Nevermind, it's in rcon.py already 👍

rcon/rcon.py
```diff
@@ -1312,6 +1314,99 @@ def set_maprotation(self, map_names: list[str]) -> list[Layer]:
...
+    def get_objective_row(self, row: int):
+        return super().get_objective_row(row)
+    
+    def get_objective_rows(self) -> List[List[str]]:
+        try:
+            return [
+                self.get_objective_row(row)
+                for row in range(5)
+            ]
+        except CommandFailedError:
```
FlorianSW, line 1328: I'm not quite sure if we should default to an empty list here. A proper error message would probably make more sense. Especially as there might be errors, which are not related to the map not having objective rows (like Skirmish). What do you think?

rcon/commands.py
```diff
@@ -366,12 +366,30 @@ def _read_list(self, raw: bytes, conn: HLLConnection) -> list[str]:
 
     @_auto_retry
     def _get_list(
-        self, item: str, can_fail=True, conn: HLLConnection | None = None
+        self,
+        item: str,
+        can_fail=True,
+        conn: HLLConnection | None = None,
+        fail_msgs: Sequence[str] = [],
```
FlorianSW, line 373: Is there any other known list command that returns other things than the actual list or FAIL? If not, what about having a more generic handling: when not fail and not a list, then it's a command failed error?

timraay: There's also the possibility that there's still junk data on the buffer. While I doubt it'd have any noticeable impact, it shouldn't be handled the same way as a game server `FAIL`.

cemathey (approved):
This worked for me when I tested it earlier.

Holding the merge until you guys indicate you're good with it.

PR #635, opened 2024-08-21
Set blacklists to an empty list on initialization

cemathey:
* Use an empty array for the default value instead of `undefined` to allow the player history page to load correctly


LordofAgents (approved):
Looking good

ElGuillermo (approved):

PR #639, opened 2024-08-22
Fix for being unable to select individual blacklist servers & blacklist record search parameters not working

timraay:


cemathey (approved):

rcongui/src/components/Blacklist/BlacklistRecords.js
```diff
@@ -15,10 +15,10 @@ import { BlacklistRecordCreateButton } from "./BlacklistRecordCreateDialog";
...
 
 async function getBlacklistRecords(searchParams) {
-  let path = "get_blacklist_records?"
-  // remove all params = 0 or being falsy
-  let searchParamsString = Object.entries(searchParams).filter(param => param && param !== 0)
-  path += searchParamsString
+  let path = "get_blacklist_records?" + new URLSearchParams(
+    Object.entries(searchParams)
+      .filter(([_, v]) => v && v !== 0)
+  );
```
Dorfieeee, line 21: Since the `blacklist_id` can be `0` and the user should be able to filter by the `default` blacklist the filter function needs to filter out all empty values only. However, it seems the server excludes expired blacklist records in either case when `exclude_expired = true OR exclude_expired = false` are provided. Therefore the filter needs to filter out false values as well.

```
  path += new URLSearchParams(
    Object.entries(searchParams).filter(
      ([_, value]) => value !== "" && value !== null && value !== undefined && value !== false
    )
  );
```

rcongui/src/components/Blacklist/BlacklistListCreateDialog.js
```diff
@@ -49,7 +49,7 @@ export default function BlacklistListCreateDialog({
   }
 
   function toggleAllServers(enabled) {
-    setServerNumbers(enabled ? null : Object.keys(servers).map(parseInt));
+    setServerNumbers(enabled ? null : Object.keys(servers).map(n => parseInt(n)));
```
Dorfieeee, line 52: As far as I know, this makes no difference.

Dorfieeee (changes requested):

rcongui/src/components/Blacklist/BlacklistRecords.js
Dorfieeee: image

rcongui/src/components/Blacklist/BlacklistListCreateDialog.js
timraay: It definitely did. The problem is that the function passed to `map` takes up to three arguments, and will toss as many as it can fit inside of `parseInt`, which we do not want and was causing `['1', '2', '3']` to be mapped to `[1, null, null]`.

Dorfieeee: Oh. Good point.

PR #643, opened 2024-08-23
Hotfix/blacklist penalty count

cemathey:
This will add a new player action (temp or perma) when adding a new blacklist record and will record a new action when editing an existing ban (temp -> perma or perma -> temp).

This means the penalty counts that are used in various places (live view, player profile page, maybe others?) will be recorded which mimics the old style behavior when doing a temp/perma native ban or when using the old blacklist.

I made a small refactor to remove some hard coded values (added an enum).

I haven't finished testing however it does appear to work properly when adding/editing a blacklist record; I need to make sure I didn't break anything for recording other actions like punishes/messages/etc.

### DIFF
--- a/rcon/blacklist.py
+++ b/rcon/blacklist.py
@@ -51,13 +51,13 @@ SERVER_NUMBER = int(get_server_number())
 
 def update_penalty_count(
     player_id: str,
-    player: PlayerID,
+    player_names: list[str],
     action: PlayerActionState,
     admin_name: str = "",
 ):
     """Save the action when blacklist records are added/edited for penalty count calculation"""
     try:
-        player_name = player.names[0].name
+        player_name = player_names[0]
     except IndexError:
         player_name = ""
 
@@ -508,6 +508,7 @@ def add_record_to_blacklist(
 ):
     with enter_session() as sess:
         player = _get_set_player(sess, player_id)
+        player_names = [n.name for n in player.names]
         if not player:
             raise RuntimeError(
                 "Unable to create player Steam ID, check the DB connection"
@@ -544,7 +545,10 @@ def add_record_to_blacklist(
         action = PlayerActionState.PERMABAN
 
     update_penalty_count(
-        player_id=player_id, player=player, action=action, admin_name=admin_name
+        player_id=player_id,
+        player_names=player_names,
+        action=action,
+        admin_name=admin_name,
     )
 
     return res
@@ -558,6 +562,8 @@ def edit_record_from_blacklist(
 ):
     with enter_session() as sess:
         record = get_record(sess, record_id, True)
+        player_names = [n.name for n in record.player.names]
+
         # Save old state
         old_record = record.to_dict()
         old_server_mask = record.blacklist.servers
@@ -604,7 +610,7 @@ def edit_record_from_blacklist(
         if old_record["expires_at"] and not new_record["expires_at"]:
             update_penalty_count(
                 player_id=record.player.player_id,
-                player=record.player,
+                player_names=player_names,
                 action=PlayerActionState.TEMPBAN,
                 admin_name=new_record["admin_name"],
             )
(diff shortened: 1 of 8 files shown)


rcon/blacklist.py
```diff
@@ -558,6 +600,23 @@ def edit_record_from_blacklist(
                 )
             )
 
+        # temp ban -> perma ban
+        if old_record["expires_at"] and not new_record["expires_at"]:
+            update_penalty_count(
+                player_id=record.player.player_id,
+                player=record.player,
+                action=PlayerActionState.TEMPBAN,
```
timraay, line 608: Shouldn't this be `PERMABAN`? And the instance further below `TEMPBAN`?

rcon/blacklist.py
```diff
@@ -558,6 +600,23 @@ def edit_record_from_blacklist(
...
+            )
+        # perma ban -> temp ban
+        elif not old_record["expires_at"] and new_record["expires_at"]:
+            update_penalty_count(
+                player_id=record.player.player_id,
+                player=record.player,
+                action=PlayerActionState.PERMABAN,
+                admin_name=new_record["admin_name"],
+            )
+
```
timraay, line 619: The ideal way to "unban" someone is by expiring the record. Would it be worth checking here if it is past the new expiration date?

If so, there's a few other cases to handle here as well, such as going from expired to a temporary ban.

rcon/models.py
```diff
@@ -139,13 +141,17 @@ def vip(self) -> Optional["PlayerVIP"]:
         )
 
     def get_penalty_count(self) -> PenaltyCountType:
-        penalities_type = {"KICK", "PUNISH", "TEMPBAN", "PERMABAN"}
-        counts: PenaltyCountType = {"KICK": 0, "PUNISH": 0, "TEMPBAN": 0, "PERMABAN": 0}
+        counts = defaultdict(int)
         for action in self.received_actions:
-            if action.action_type in penalities_type:
-                counts[action.action_type] += 1
+            if PlayerActionState[action.action_type] in PlayerActionState:
```
timraay, line 146: This is a weird condition. Previously it asserted that `action.action_type` was a legitimate action type. Here it will throw a `ValueError`(?).

This should be something like `if action_state := PlayerActionState._member_map_.get(action.action_type):` instead.

rcon/rcon.py
```diff
@@ -1136,7 +1133,7 @@ def punish(self, player_name: str, reason: str, by: str) -> bool:
         safe_save_player_action(
             rcon=self,
             player_name=player_name,
-            action_type="PUNISH",
+            action_type=PlayerActionState.MESSAGE,
```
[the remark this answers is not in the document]
cemathey: Yes, yes it should

rcon/player_history.py
```diff
@@ -497,15 +505,15 @@ def remove_flag(
...
     with enter_session() as sess:
         player = sess.query(PlayerID).filter_by(player_id=player_id).one_or_none()
-        actions: list[PlayerActionType] = []
+        actions: list[PlayerActionState] = []
         if player:
             actions = [
                 action.to_dict()
                 for action in player.received_actions
-                if action.action_type == "MESSAGE"
+                if action.action_type == PlayerActionState.MESSAGE.value
```
timraay, line 516: Unless I'm going crazy, `action` is an instance of `rcon.models.PlayersAction`, making `action.action_type` a string. And we're comparing it to a string here.

timraay (changes requested):

rcon/blacklist.py
cemathey: Whoops! Fixed

rcon/models.py
cemathey: I agree, I was focused on the previous test; I changed it to a more conventional test

rcon/player_history.py
```diff
@@ -497,15 +505,15 @@ def remove_flag(
     return player, old_flag
 
 
-def get_player_messages(player_id: str) -> list[PlayerActionType]:
+def get_player_messages(player_id: str) -> list[PlayerActionState]:
     with enter_session() as sess:
         player = sess.query(PlayerID).filter_by(player_id=player_id).one_or_none()
-        actions: list[PlayerActionType] = []
+        actions: list[PlayerActionState] = []
```
timraay, line 511: Minor thing but this should be `PlayerActionType`, same for the function's return type

timraay (changes requested):

timraay (approved):

PR #648, opened 2024-08-28
Merge foreign key IDs for duplicated player IDs

cemathey:
Courtesy of SourceCode in the Discord.

I just translated this to use the SQLAlchemy session for running the textual SQL instead of the raw connection.

Ran it against a copy of our database that has duplicate records, looks like it works fine and it allows the player IDs to convert after the merging command runs.

Took about 8 seconds to run on our VPS.

### DIFF
--- a/rcon/cli.py
+++ b/rcon/cli.py
@@ -35,7 +35,7 @@ from rcon.user_config.webhooks import (
     BaseWebhookUserConfig,
 )
 from rcon.models import PlayerID, enter_session
-from sqlalchemy import update
+from sqlalchemy import update, select, text
 from sqlalchemy import func as pg_func
 from rcon.utils import ApiKey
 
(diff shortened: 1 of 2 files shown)


rcon/cli.py
```diff
@@ -423,6 +423,46 @@ def reset_user_settings(server: int):
...
+def merge_duplicate_player_ids():
+    logger.info(f"Merging duplicate player ID records")
+    players = {}
+
+    with enter_session() as session:
+        stmt = select(PlayerID)
+        rows = session.execute(stmt).scalars()
+
+        for player in rows:
+            id_, steamid, date = player.id, player.player_id, player.created
```
timraay, line 436: `date` isn't being used? Not an issue on itself but I thought the idea was to preserve oldest?

On that topic, wouldn't it be better to preserve newest, since recent data will be more relevant than old data?

timraay (approved):

PR #653, opened 2024-08-30
Game Layout selector fixes + more

timraay:
- Fix Omaha and Utah having the wrong orientation
- Fix Hurtgen tacmap not loading correctly
- Fix log history view "From" filter not working
  - Made the `get_historical_logs` endpoint also accept `from` again as a parameter, which was previously renamed to `from_`.
- Fix some night maps showing up as `unknown` in the map history


Dorfieeee (approved):
LGTM

PR #662, opened 2024-09-01
Typo   missing 's' in permission strings

ElGuillermo:
As found by "Sniper" : https://discord.com/channels/685692524442026020/706075642885832764/1279884170369175682


cemathey (changes requested):
This one we have to do a little bit different.

The change to `models.py` is correct, but we don't need to alter all those other migrations.

We do need to generate a new migration though (or the typo will persist in everyones current installs); the easiest way I find to do that is to just make the change to `models.py`; build/launch the service and copy the auto generated migration of out of the container.

I can walk you through that or do it for you if I'm not explaining it clearly.

ElGuillermo:
Oups. Didn't think that could generate new migrations... We also can ignore that all, log junkies aren't too many :)

cemathey (approved):

cemathey:
Okay beats me, I can't get it to generate the new migration (must not notice changes in only the description) or something, we'll worry about the migration later because I don't feel like doing it by hand.

Issue #652, opened 2024-08-29
`get_public_info` assumes PUBLIC_STATS_PORT is an int


timraay:
The `default.env` file says that you can prefix `PUBLIC_STATS_PORT` with `127.0.01:` to disable access. While the stats can also be disabled through Settings > CRCON Settings > `lock_stats_api`, this is still useful if you want to disable the HTTP port while leaving the HTTPS port accessible.

The implementation of `get_public_info` does not account for the case where the variable is prefixed and will error when trying to convert it to an int. If the value cannot be converted to an int it should instead return `None`.

There might be other instances where this same issue can occur.

PR #675, opened 2024-09-04
Update default.env

ElGuillermo:
Typos and outdated infos removed

### DIFF
--- a/default.env
+++ b/default.env
@@ -13,7 +13,7 @@
 # Redis/Postgres Repository & Version
 # These are the specific redis and postgres images CRCON will pull and use, they support
 # all the common platforms by default (amd64, arm64, arm32, etc) but you can change which
-# specific image is being pulled here if you know what you are doine.
+# specific image is being pulled here if you know what you are doing.
 # Be careful with upgrading postgres major versions (going from 12 to 13, etc) because the
 # database versions aren't necessarily binary compatible with each other and may require you
 # to take manual steps to upgrade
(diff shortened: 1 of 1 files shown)


default.env
```diff
@@ -122,24 +122,21 @@ HLL_HOST=
...
 HLL_PASSWORD=
 
 # -----------------------------
 # RCON Web Setttings
 # -----------------------------
 
-# This is the port where the Rcon Website will be accessible. Feel free to change. 
-# If you're running multiple instances you have to change it per instance
-# If you want to put a reverse proxy on your rcon (e.g. for HTTPS) and block the direct 
-# access then use: 127.0.0.1:8010
```
timraay, line 135: This is still possible, no?

default.env
```diff
@@ -122,24 +122,21 @@ HLL_HOST=
...
+# This is the port where the CRCON website will be accessible. Feel free to change. 
+# If you're running multiple instances, you have to change it per instance
 RCONWEB_PORT=8010
 
 # Same as above but for https
 RCONWEB_PORT_HTTPS=9010
 
 # Ports for the public version (statistics)
-# If you don't want the public site write add "127.0.0.1:" in front of the port number, example: 127.0.0.1:7010
```
timraay, line 142: Even  though it no longer works, it still has a use case, hence I wouldn't consider this outdated. See #652

timraay (changes requested):

default.env
ElGuillermo: So we should find another way to present it, as we got two or three bug reports related to this the last days (?)

default.env
ElGuillermo: Maybe. Not sure, as I do not use reversing. + See the other note below.

cemathey:
This is something I'm very unfamiliar with this, I know some people use some of this stuff but when I set up URLs I do it differently.

default.env
FlorianSW: This is absolutely possible and we should keep that comment :)

ElGuillermo:
Let's stall this one...

PR #676, opened 2024-09-04
Blacklist delete confirmation dialog

### LINKED ISSUE #668: Frontend: Add confirmation before blacklist deletion
You can delete an entire blacklist including all of its records with a single click. Pretty dangerous.

Dorfieeee:
Resolves #668 

> Exhibit
image


FlorianSW (approved):
didn't test it locally, but code looks good :)

PR #677, opened 2024-09-04
Fix votekick threshold due to API change

Dorfieeee:
The frontend failed to handle the `POST /api/set_votekick_thresholds` request because the endpoint now expects an `[int, int][]` array in version 10, whereas it accepted a `string` in version 9.


cemathey (approved):
I can't believe I didn't notice this when I was looking at it, I only looked on the backend

PR #678, opened 2024-09-04
[FEATURE] Allow arbitrary RCon commands in chat commands

FlorianSW:
There are communities that would like to use chat commands with a more powerful set of actions that can be triggered when a player uses a chat command. Two real-worl example from the "event server for prepping" use case:

### Switching player side
Changing a side without needing to redeploy for themself, a command "!switch" is useful. Issuing this command will forward a switch_player_now RCon command with the player_name of the issuing player.

### Redeploying with 10 instead of 20 seconds
Redeploying with the `punish` command might be useful when needing to redeploy for multiple times. A command "!redeploy", which issues a `punish` command for the issuing player, might be useful here.

This feature is implemented with this PR. A chat command can either have a `message` key or a `command` key (message takes precendence). The `command` config is an object that has the same signature and functinality as auto-settings commands.
This allows for a powerful set of tools, while it also has some risks, as _everything_ can now be executed with a player issuing a specific chat command.

The parameters of the command will replace message parameters same as the message itself, however, only context keys are populated. No other variables are available in commands params.

### DIFF
--- a/rcon/api_commands.py
+++ b/rcon/api_commands.py
@@ -1769,3 +1769,11 @@ class RconAPI(Rcon):
 
     def get_objective_row(self, row: int):
         return super().get_objective_row(int(row))
+
+    def set_chat_command_enabled(self, word: str, enabled: bool):
+        c = ChatCommandsUserConfig.load_from_db()
+        for command in c.command_words:
+            if word not in command.words:
+                continue
+            command.enabled = enabled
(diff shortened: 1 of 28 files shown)


cemathey (approved):
Couple of things in here I think we should probably tweak but it builds/worked for me and I don't want to be too picky.

PR #714, opened 2024-09-30
Add new standardized message template feature

cemathey:
* Adds the endpoints listed at the bottom to the API 
* Adds a new database table for storing message templates
* Adds the new endpoint permissions (owner/admin can do everything, moderator can view)

I didn't extensively test this but did poke around a bit in Postman; everything but add/delete (`POST`) are `GET` methods

```python
def get_message_templates(
       category: MessageTemplateCategory
) -> list[MessageTemplateType]:
    """Get all possible message type categories"""
    ...

def get_message_template_categories(self) -> list[MessageTemplateCategory]:
    return get_message_template_categories()

def get_message_template(id: int) -> MessageTemplateType | None:
    """Return the message template for the specified record if it exists"""
    ...

def get_all_message_templates(self) -> AllMessageTemplateTypes:
    """Get all message templates by category"""
    ...

def add_message_template(
     title: str, content: str, category: str | MessageTemplateCategory, by: str
) -> int:
    """Add a new message template and return the ID of the new record"""
    ...

def delete_message_template(id: int) -> bool:
    """Delete a specific message template"""
    ...
```


Dorfieeee:
I've built the UI using this API and have NOT found any major issues.
However, there is an issue that should not be forgotten
- When a parameter is missing it does not return JSON response ( => uncaught error)

Future usage: This platform could be used for saving and editing presets like `auto-settings`

Dorfieeee (approved):

PR #715, opened 2024-10-03
Bugfix: include is_vip and profile in get_detailed_player_info

LordofAgents:
Hello,
`get_detailed_player_info` did not override the default `false` for `is_vip` and `null` for `profile` with the correct values. 
This PR fixes that and removes the workaround in `get_team_view` to not fetch the profiles twice. This also fixes the wrong value `is_vip` in `get_detailed_players`.

Thank you!

### DIFF
--- a/rcon/rcon.py
+++ b/rcon/rcon.py
@@ -254,12 +254,6 @@ class Rcon(ServerCtl):
         players_by_id = detailed_players["players"]
         fail_count = detailed_players["fail_count"]
 
-        logger.debug("Getting DB profiles")
-        steam_profiles = {
-            profile[PLAYER_ID]: profile
-            for profile in get_profiles(list(players_by_id.keys()))
-        }
-
         logger.debug("Getting VIP list")
         try:
             vips = set(v[PLAYER_ID] for v in super().get_vip_ids())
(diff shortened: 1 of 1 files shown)


LordofAgents:
The last commit fixes the not showing time in the Team View Frontend Site.

rcon/rcon.py
```diff
@@ -467,8 +451,15 @@ def get_detailed_player_info(self, player_name: str) -> GetDetailedPlayer:
...
 
         """
+        # Add VIP Status
+        player_data = parse_raw_player_info(raw, player_name)
+        vip_player_ids = set(v[PLAYER_ID] for v in super().get_vip_ids())
+        player_data["is_vip"] = player_data["player_id"] in vip_player_ids
 
-        return parse_raw_player_info(raw, player_name)
+        # Add Profile
+        profile = get_profiles([player_data["player_id"]])
```
cemathey, line 460: I think you should use `get_player_profile` which you can call by player ID to avoid the (probably slim) chance of index errors if we get back an empty list from `get_profiles`

I don't have access to your repo so I can't change it 😄

cemathey (changes requested):

cemathey (approved):

PR #720, opened 2024-10-08
Include required permissions by endpoint in `get_api_documentation`

cemathey:
* Creates our own `permission_required` decorator which just passes through to Django's method so that we can track which permissions are required by endpoint
* Include the permission(s) required in `get_api_documentation` to make it easier to determine which access a user requires to consume the API (whether through a tool or the UI)

Output looks like:

```json
 {
      "endpoint": "add_admin",
      "arguments": {
        "player_id": {
          "default": null,
          "annotation": null
        },
        "role": {
          "default": null,
          "annotation": null
        },
        "description": {
          "default": null,
          "annotation": null
        }
      },
      "return_type": "<class 'bool'>",
      "doc_string": null,
      "auto_settings_capable": true,
      "permissions_required": [
        "api.can_add_admin_roles"
      ],
      "allowed_http_methods": [
        "POST"
      ]
    },
```

### DIFF
--- a/rconweb/api/audit_log.py
+++ b/rconweb/api/audit_log.py
@@ -2,14 +2,13 @@ import json
 import logging
 from functools import wraps
 
-from django.contrib.auth.decorators import permission_required
 from django.views.decorators.csrf import csrf_exempt
 from sqlalchemy import and_, or_
 
 from rcon.models import AuditLog, enter_session
 
 from .auth import api_response, login_required
-from .decorators import require_http_methods
+from .decorators import require_http_methods, permission_required
 from .utils import _get_data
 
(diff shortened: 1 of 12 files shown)


LordofAgents:
@cemathey Is this still relevant? Awesome work, but maybe we should merge it in master now and resolve the merge conflicts

cemathey:
> @cemathey Is this still relevant? Awesome work, but maybe we should merge it in master now and resolve the merge conflicts

Yeah still relevant, I'll fix the merge conflicts and catch up with master and make sure nothing else needs to get updated at the same time.

cemathey:
Okay updated, tested it and it works still, should be mergeable.

PR #734, opened 2024-10-24
Req/directory

FlorianSW:
Builds and runs locally without any obvious problems :)


cemathey:
Just leaving a comment here so we don't forget; this does not work with Postgres 12 so we can't merge unless we also handle updating that.

PR #741, opened 2024-10-28
fix: vip_name in reward_players typo

BauerPh:
Log:
```
[2024-10-28 20:27:19,871][INFO][teamphoenix1PublicEvent][v10.5.1] rcon.seed_vip.utils utils.py:reward_players:269 | config.dry_run=False adding VIP to player_id='XXXXX' player=None vip_name='Thank you for helping us seed.\n\nThe server is now live!' expiration_date=datetime.datetime(2024, 11, 2, 20, 27, 19, 662685, tzinfo=datetime.timezone.utc)
```
VIP List:
image

PR should fix this?!

### DIFF
--- a/rcon/seed_vip/utils.py
+++ b/rcon/seed_vip/utils.py
@@ -257,8 +257,8 @@ def reward_players(
             continue
 
         vip_name = (
-            player.player.name
             if player
+                player.player.name            
             else format_vip_reward_name(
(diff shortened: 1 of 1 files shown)


BauerPh:
nevermin, I don't work with python normally :/

ElGuillermo:
Thanks for trying ! Already fixed in https://github.com/MarechJ/hll_rcon_tool/pull/732

PR #757, opened 2024-11-22
Add U16 layers, update layer details and images, update layer parser

timraay:
- Adds layers and images for Elsenborn Ridge
- Added missing Hill 400 Dawn and PHL Rain images
- Updated Omaha Dusk image which was outdated
- Removed vignette from map images that had them
- Added speculative layer names and images for rumoured upcoming layers
- Changed PHL Warfare environment from `Day` to `Rain`
- Changed Omaha Night environment from `Night` to `Dusk`
- Fixed Mortain shortname being `MOR` rather than `Mortain`
- Updated the layer name parser to support the `{TAG}_L_{YEAR}_Warfare` format
- Fixed an issue that could cause the layer name parser to fail due to a conditionally unbound variable


Dorfieeee:
Test my commit please as I have no access to the current build at the moment.
- [] - Votemap Whitelist reset button
- [] - Votemap page does not crash due to unknown map id
- [] - Can turn on/off map rotation shuffle

ElGuillermo (approved):

PR #762, opened 2024-11-28
Fix parameter handling in RCon Chat Commands

FlorianSW:
These commands can be passed through to any type of rcon api, which might not just take strings as parameters´, but lists, dicts and so on as well.

### DIFF
--- a/rcon/hooks.py
+++ b/rcon/hooks.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 from datetime import datetime, timedelta, timezone
 from functools import wraps
 from threading import Timer
-from typing import Final
+from typing import Any, Final
 
 from discord_webhook import DiscordEmbed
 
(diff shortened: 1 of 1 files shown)


rcon/hooks.py
```diff
@@ -238,17 +238,71 @@ def chat_rcon_command(
...
             for k, v in params.items():
-                for i, a in enumerate(args):
-                    v = v.replace(f"${i + 1}", a)
-                commands[name][k] = format_message_string(v, context=ctx)
+                commands[name][k] = replace_params(ctx, args, v)
 
         do_run_commands(rcon, commands)
 
 
+def replace_params(ctx: dict[str, str], args: list[str], v: Any) -> Any:
```
timraay, line 251: Figured I'd quickly comment on why I've changed this from `str or list or dict` to `Any`. First off, using `or` in annotations isn't officially supported syntax (and I doubt many type checkers support it either). Either way, this function accepts more than just those three types, for instance floats and ints, hence `Any`.

PR #767, opened 2024-12-04
Feature/player autocomplete

marcelhillesheim:
Fetching history players to autofill name and steam id when adding vips or console admin.

Also set default role to spectator for adding console admin as the other roles are usually never used.

### DIFF
--- a/rcongui/src/components/SettingsView/admins.js
+++ b/rcongui/src/components/SettingsView/admins.js
@@ -14,6 +14,7 @@ import {
 } from "@material-ui/core";
 import DeleteIcon from "@material-ui/icons/Delete";
 import AddIcon from "@material-ui/icons/Add";
+import PlayerAutocomplete from "./playerAutocomplete";
 
 const AdminRole = ({ classes, role, setRole, roles }) => (
   <FormControl className={classes.formControl}>
@@ -40,14 +41,7 @@ const AddAdminItem = ({
   <ListItem>
     <Grid container>
       <Grid item xs={4} className={classes.paddingRight}>
-        <TextField
-          InputLabelProps={{
-            shrink: true,
-          }}
-          label="Name"
-          value={name}
-          onChange={(e) => setName(e.target.value)}
-        />
+        <PlayerAutocomplete classes={classes} name={name} setName={setName} playerId={playerId} setPlayerId={setPlayerId}/>
       </Grid>
       <Grid item xs={4} className={classes.paddingLeft}>
         <TextField
--- a/rcongui/src/components/SettingsView/admins.js
+++ b/rcongui/src/components/SettingsView/admins.js
@@ -70,7 +70,7 @@ const AddAdminItem = ({
           onAdd(name, playerId, role).then(() => {
             setName("");
             setPlayerId("");
-            setRole("");
+            setRole("spectator");
           })
         }
       >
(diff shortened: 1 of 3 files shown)


rcongui/src/components/SettingsView/admins.js
```diff
@@ -76,7 +70,7 @@ const AddAdminItem = ({
           onAdd(name, playerId, role).then(() => {
             setName("");
             setPlayerId("");
-            setRole("");
+            setRole("spectator");
```
FlorianSW, line 73: While I like this addition, there is one little caveat to this: The available roles are defined by the server. And while most servers will sureley never ever touch this config at all, yet alone delete the spectator role, it _might_ still happen. Hence, we should definitely check that the spectator role actually exists, and if not, select any of the other available roles as default, as a fallback :)

rcongui/src/components/SettingsView/playerAutocomplete.js
```diff
@@ -0,0 +1,98 @@
...
+        if (data.failed) {
+          return;
+        }
+        setSuggestions(data.result.players);
+      })
+      .catch(handle_http_errors);
+  };
+
+  const debouncedFetchSuggestions = useMemo(
+    () => debounce(fetchSuggestions, 300),
```
FlorianSW, line 46: I like that you thought about debouncing here, big +1 :)

FlorianSW (changes requested):
@marcelhillesheim big thanks for this PR. It overall looks good, I just have one little comment :)

Also, as I wasn't able yet to build the frontend locally, would you be able to post a screenshot of how this looks in the add console admin dialog? If it is not too big of a burden, otherwise I'll just build it later locally :)

rcongui/src/components/SettingsView/admins.js
marcelhillesheim: Didn't know that the roles are defined in a config. Should be fixed now. 

player_suggestion

Dorfieeee:
I don't think this is necessary if we're going to roll out the updated UI soon

marcelhillesheim:
> I don't think this is necessary if we're going to roll out the updated UI soon

I did checkout the new ui and you still have to enter the steamid manually or is there a planned feature which hasnt been pushed?

FlorianSW:
> I don't think this is necessary if we're going to roll out the updated UI soon

Oh, this is actually a good point 😅 



> > I don't think this is necessary if we're going to roll out the updated UI soon
> 
> I did checkout the new ui and you still have to enter the steamid manually or is there a planned feature which hasnt been pushed?

This feature might not exist there, either. Would you be willing to port your PR to the frontend/v2-beta brancht? ❤️

Dorfieeee:
I may add another action "ADD/REMOVE ADMIN" to the list. That should be effectively the same.

marcelhillesheim:
> I may add another action "ADD/REMOVE ADMIN" to the list. That should be effectively the same.

Ah okay that makes sense. But there is still the console admin page so i would still want to add this feature just as a comfort option. So that both saved players and players, who didn't join the server yet, can be added in one place.

> > I don't think this is necessary if we're going to roll out the updated UI soon
> 
> This feature might not exist there, either. Would you be willing to port your PR to the frontend/v2-beta brancht? ❤️

Yeah i took a look at the v2 branch, there are a lot of changes. I will adjust my branch later this week and will create a new PR

PR #769, opened 2024-12-06
fix replacing message context for rcon chat commands

FlorianSW:


cemathey (approved):
gave it a quick skim, looks good to me

PR #771, opened 2024-12-06
Hotfix/automod flags

cemathey:
Prevent crashing when a player has a `None` profile (should investigate why this is possible later) but this will allow the automods to work again.

### DIFF
--- a/rcon/automods/models.py
+++ b/rcon/automods/models.py
@@ -125,7 +125,7 @@ class PunitionsToApply:
                             role=p.get("role"),
                             lvl=p.get("level"),
                         )
-                        for p in squad.get("players", [])
+                        for p in squad.get("players", []) if p
                     ],
                 )
             )
--- a/rcon/automods/models.py
+++ b/rcon/automods/models.py
@@ -111,23 +111,23 @@ class PunitionsToApply:
         try:
             if any(s.team == team and s.name == squad_name for s in self.squads_state):
                 return
-            self.squads_state.append(
-                ASquad(
-                    team=team,
-                    name=squad_name,
-                    players=[
-                        PunishPlayer(
-                            player_id=p.get("player_id"),
-                            name=p.get("name"),
-                            squad=p.get("unit_name"),
-                            team=p.get("team"),
-                            flags=p.get("profile", {}).get("flags", []),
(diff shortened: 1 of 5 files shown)


rcon/automods/models.py
```diff
@@ -111,23 +111,23 @@ def add_squad_state(self, team: str, squad_name: str, squad: dict):
...
+
+            players = []
+            for player in squad.get("players", []):
+                profile = player.get("profile", {})
+                punish_player = PunishPlayer(
+                    player_id=player.get("player_id"),
+                    name=player.get("name"),
+                    squad=player.get("unit_name"),
+                    team=player.get("team"),
+                    flags=profile.get("flags", []) if profile else [],
```
FlorianSW, line 123: the if case shouldn't be needed with the default value in `get()`, isn't it? :)

rcon/automods/models.py
```diff
@@ -111,23 +111,23 @@ def add_squad_state(self, team: str, squad_name: str, squad: dict):
...
+
+            players = []
+            for player in squad.get("players", []):
+                profile = player.get("profile", {})
+                punish_player = PunishPlayer(
+                    player_id=player.get("player_id"),
+                    name=player.get("name"),
+                    squad=player.get("unit_name"),
+                    team=player.get("team"),
+                    flags=profile.get("flags") if profile else [],
```
FlorianSW, line 123: profile can never be None. flags on the other hand can now. So I would suggest this line:
```
flags=profile.get("flags", []),
```

FlorianSW (changes requested):

rcon/automods/models.py
ElGuillermo: Are you sure about the fact "profile can never be None" ?

Because that's what I understand whe I see
```
flags=p.get("profile", {}).get("flags", []),
          ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```
?

FlorianSW: Ah, yes, you're right. I misinterpreted the default in get where the key exists in the dict, but the corresponding value is None.

FlorianSW (approved):

PR #779, opened 2024-12-15
Feature/player autocomplete v2

marcelhillesheim:
Added player autocompletion for console admin and vip pages. 

Set default role to spectator for adding console admin.

### DIFF
--- a/rcongui/src/components/form/custom/PlayerAutocompletion.jsx
+++ b/rcongui/src/components/form/custom/PlayerAutocompletion.jsx
@@ -1,0 +1,87 @@
+import React, {useEffect, useMemo, useRef} from "react";
+import {debounce} from "lodash";
+import {handle_http_errors, postData, showResponse} from "@/utils/fetchUtils";
+import Autocomplete from "@mui/material/Autocomplete";
+import Typography from "@mui/material/Typography";
+import {Avatar, Box} from "@mui/material";
+import TextField from "@mui/material/TextField";
+
+export default function PlayerAutocompletion({state, setState}) {
+
+  const [suggestions, setSuggestions] = React.useState([]);
+
+  const nameRef = useRef(state.name);
+
+  useEffect(() => {
+    debouncedFetchSuggestions();
+    nameRef.current = state.name;
+  }, [state.name]);
+
+  const fetchSuggestions = () => {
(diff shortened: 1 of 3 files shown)


rcongui/src/components/form/custom/PlayerAutocompletion.jsx
```diff
@@ -0,0 +1,87 @@
...
+import {debounce} from "lodash";
+import {handle_http_errors, postData, showResponse} from "@/utils/fetchUtils";
+import Autocomplete from "@mui/material/Autocomplete";
+import Typography from "@mui/material/Typography";
+import {Avatar, Box} from "@mui/material";
+import TextField from "@mui/material/TextField";
+
+export default function PlayerAutocompletion({state, setState}) {
+
+  const [suggestions, setSuggestions] = React.useState([]);
```
FlorianSW, line 11: I would prefer an import of `useState` instead of using `React.useState`, as it is consistent with the project and with other usages, like `useRef`, etc.

rcongui/src/components/form/custom/PlayerAutocompletion.jsx
```diff
@@ -0,0 +1,87 @@
+import React, {useEffect, useMemo, useRef} from "react";
+import {debounce} from "lodash";
+import {handle_http_errors, postData, showResponse} from "@/utils/fetchUtils";
+import Autocomplete from "@mui/material/Autocomplete";
+import Typography from "@mui/material/Typography";
+import {Avatar, Box} from "@mui/material";
+import TextField from "@mui/material/TextField";
+
+export default function PlayerAutocompletion({state, setState}) {
```
FlorianSW, line 9: just nit-picking: I would use `player` and `setPlayer` instead, as it is more meaningful than `state` in this context :)

FlorianSW (changes requested):
two little things, let me know what you think :)

rcongui/src/components/form/custom/PlayerAutocompletion.jsx
marcelhillesheim: fixed it 🙂

FlorianSW (approved):
Awesome :) Looks good, thanks for your contribution!

PR #791, opened 2024-12-17
make toggle button for log type filter on/off consistent in their beh…

FlorianSW:
…aviour

Before this commit the button group would behave differently on what happens when it is on or off and what happens when on or off is clicked again respectively.

With this commit, the button, which is clicked is always the active one, even though it was active before.

Ref:
https://discord.com/channels/685692524442026020/706075642885832764/1318593170354868285

### DIFF
--- a/rcongui/src/components/table/selection/LogActionHighlightMenu.jsx
+++ b/rcongui/src/components/table/selection/LogActionHighlightMenu.jsx
@@ -1,18 +1,23 @@
 import {
+  Box,
+  Button,
+  Checkbox,
   List,
   ListItem,
   ListItemButton,
-  Box,
   ListItemIcon,
-  Checkbox,
   ListItemText,
-  ToggleButtonGroup,
   ToggleButton,
+  ToggleButtonGroup,
+  Tooltip,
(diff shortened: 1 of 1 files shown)


Dorfieeee:
What is the cause by the way?

FlorianSW:
> What is the cause by the way?

A toggle group can have the "no button is enabled" state. In exclusive mode (where otherwise only one item can be enabled), this means a value in the `onChange` handler of `null`. `null` in our code meant `off` (we check if `value` is `on`). I think it is a different idea of `ToggleButtonGroup`. I would've expected to "toggle between buttons", however, it seems you can still enable and disable different toggle buttons 🙈 

However, the doc indicates it pretty clearly, once I read it is was obvious 😅 
https://mui.com/material-ui/api/toggle-button-group/
> value of the selected buttons. When exclusive is true this is a single value; when false an array of selected values. If no value is selected and exclusive is true the value is null; when false an empty array.

Dorfieeee (commented):
LGTM

Dorfieeee (approved):

PR #793, opened 2024-12-17
remove code handling whitespace/pineapple name bug

FlorianSW:


cemathey (approved):
Looks good to me, I built on master and then swapped to this one to make sure upgrading wouldn't cause any issues.

All I had to do was add a missing import.

FlorianSW:
Thanks for the review :)

> All I had to do was add a missing import.

Weird, I can't remember removing this imports intentionally 🙈

PR #794, opened 2024-12-17
Feat/watch killrate

cemathey:
This is a mostly port of ElGuillermo's kill rate plug in but I changed a few things and added some different options.

This uses the stats calculated by CRCON instead of calculating the KPM again (with different math) so they should always be consistent.

I integrated it into the services stuff, it's off by default.

The KPM is adjustable for different weapon categories (everything, armor, artillery and machine guns) so people can adjust as they see fit.

By default it will only report a player once a match to cut down on KPM spam because presumably they only need to be investigated once.

I did not touch the UI so right now it's only configurable through the API.

Otherwise it is pretty straight forward, it pulls stats, checks for people that exceed the KPM threshold who aren't whitelisted in some way and then sleeps for the poll period.


rcon/watch_killrate.py
```diff
@@ -0,0 +1,389 @@
...
+from time import sleep
+import discord
+from rcon.user_config.rcon_server_settings import RconServerSettingsUserConfig
+from rcon.scoreboard import current_game_stats
+
+logger = getLogger(__name__)
+
+# TODO: move these to a common location and get translations from T17
+# sourced from https://gist.github.com/timraay/5634d85eab552b5dfafb9fd61273dc52#playerinfo-player_name
+ARTILLERY: dict[str, bool] = {
```
FlorianSW, line 24: With the weapons.py file from https://github.com/MarechJ/hll_rcon_tool/pull/797 this could be removed. We should, no matter which one we keep, not have to maintain two different sets of Weapon lists in CRCon 😅

cemathey: Yeah I very much agree; do you want to approve for now and then I will update once that gets merged?

I have asked T17 (and just bumped the request) to get translations for map names/weapon names for all the versions they've localized for, and once we have those I'd like to add some options to localize that in places where it gets displayed to the user (like chat commands and so on)

FlorianSW (approved):
approved, considering the before-submitted comment is worked on :)

cemathey:
I forgot to put the reported cache in redis, so the hook to clear it on match start currently does nothing.

FlorianSW:
@cemathey this one is finished right? As in: This can be released with the next release?

cemathey:
> @cemathey this one is finished right? As in: This can be released with the next release?

AFAIK nobody has updated the frontend to include it; I realized the other day we're using JSON forms to generate settings pages, so I think it's possible it's simple enough for me to add when I have some time this week if nobody beats me to it

cemathey:
Updated this to work with the webhook service and verified it works correctly.

Also fixed a few miscellaneous issues so the webhook service will start properly even when its env vars aren't set.

PR #797, opened 2024-12-19
Feature: Public frontend team stats

marcelhillesheim:
Changes regarding team stats in public frontend.

1. Calculating team based on weapon usage and enriching players with team prop.

2. Displaying team in game table (only for games in history not live). Added Team filter to table.

3. Added Router for game stats and added button group to switch between the different routes.

4. Added first chart (scatter chart).

crcon_table_team
crcon_team_stats

### DIFF
--- a/rcon/hooks.py
+++ b/rcon/hooks.py
@@ -251,61 +251,6 @@ def chat_rcon_command(
         do_run_commands(rcon, commands)
 
 
-def replace_params(ctx: dict[str, str], args: list[str], v: Any) -> Any:
-    """
-    Replaces arguments (from args) and message parameters (from ctx) into the provided parameter value (v).
-    The parameter value can be a str, list or dict, the arguments and message parameters are going to be replaced at
-    any appropriate value:
-    * in lists: for each element of the list; if a list item is a list or dict, each item is processed independently to
-      ensure nested values are covered as well
-    * in dicts: for each value, but not for keys. If a value is a list or dict, it is going to be processed independently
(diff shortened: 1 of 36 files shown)


FlorianSW:
I feel like the team indicator should be in the backend and exposed through the API. For reference, HeLO does a team determination based on the weapons already as well. If CRCon exposes this in the API, it is mich easier for other tools (and the frontend) to act on this.

If you like, I can do that in the backend :)

marcelhillesheim:
> I feel like the team indicator should be in the backend and exposed through the API. For reference, HeLO does a team determination based on the weapons already as well. If CRCon exposes this in the API, it is mich easier for other tools (and the frontend) to act on this.
> 
> If you like, I can do that in the backend :)

Would be nice. I dont have a backend dev env setup yet ❤️

marcelhillesheim:
> I may need to do some refactoring regarding the routing. I'll see into it. I don't like that there is routing defined within a components since everything else is defined inside the router file.

I can also refactor it, if you want.

Dorfieeee:
The main change is in moving the routing logic into router.tsx and that there is now game-stats-container which is re-used for both live and completed games. There may be tabs such as `table` and `charts` that don't need the player details on the side that's why it is not included in the game layout component directly.

FlorianSW (approved):

Dorfieeee:
Poor guy working hard on that team decisioning logic there and you scrape it like a dog 💩 😆

PR #806, opened 2024-12-20
Add missing `api` prefix for permission name spacing

cemathey:
Add the missing `api` prefix for the message template endpoints which is required for Django to look up the permission correctly.


cemathey:
Duplicated of https://github.com/MarechJ/hll_rcon_tool/pull/805/files

PR #807, opened 2024-12-20
Add pagination to audit log query

cemathey:
Adds pagination to the audit log query which was necessary because large tables would cause timeouts.

New parameters `page` and `page_size` which default to `1` and `10` respectively so it won't blow up if people don't pass them.

The SQL offset is `0` indexed, we account for that so people use `1` indexed page numbers, `1`, `2`, etc.

If you ask for a page # beyond the number of pages you'll receive an empty result `[]` instead of an error.

I have not touched the UI but I imagine it is pretty easy to integrate these changes.

### DIFF
--- a/rcongui/src/pages/records/audit-logs/index.jsx
+++ b/rcongui/src/pages/records/audit-logs/index.jsx
@@ -38,17 +38,39 @@ const columns = [
   { field: "command_result", headerName: "Result", flex: 1, sortable: false },
 ];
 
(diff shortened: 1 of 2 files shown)


LordofAgents (approved):

FlorianSW (changes requested):
Should not be merged until the UI is done (or at least proven it can work with the pagination out-of-the box)

PR #812, opened 2024-12-21
Run basic checks on pull requests

FlorianSW:


tests/test_ban_tk_connect.py
```diff
@@ -29,13 +31,13 @@ def test_ban_excluded_weapon(*args):
...
         "timestamp_ms": 1612695641000,
         "action": "TEAM KILL",
-        "player_name_1": "[ARC] DYDSO ★ツ",
+        "player_name_1": "[ARC] DYDSO",
         "player_id_1": 76561198091327692,
         "player_name_2": "Francky Mc Fly",
-        "player_id_1": 76561198091327692,
+        "player_id_2": 76561198091327692,
         "weapon": "None",
-        "raw": "[646 ms (1612695641)] TEAM KILL: [ARC] DYDSO ★ツ(Axis/76561198091327692) -> Francky Mc Fly(Axis/76561198133214514) with None",
```
LordofAgents, line 37: Any reasons that the special characters were removed? Wasn't it to test the handling of special characters?

tests/test_message_variables.py
```diff
@@ -138,8 +138,8 @@ def test_populate_next_map(monkeypatch, map_name, expected):
     logger.error(f"{parse_layer(map_name)=}")
     monkeypatch.setattr(
         rcon.message_variables.Rcon,
-        "next_map",
-        parse_layer(map_name),
+        "get_next_map",
```
LordofAgents, line 141: I do not understand that change. If I understand correctly, shouldn't test that the variables in a message? The variable for the next map is `next_map`, to my knowledge.

tests/test_no_leader_watch.py
```diff
@@ -826,22 +831,22 @@ def test_punish_small_squad(team_view):
...
         min_server_players_for_punish=0,
         immune_player_level=10,
         immune_roles=[],
     )
-    mod = NoLeaderAutomod(config, None)
+    mod = mod_with_config(config)
 
     watch_status = WatchStatus()
-    player = team_view["allies"]["squads"]["able"]["players"][0]
+    player = team_view["axis"]["squads"]["baker"]["players"][0]
```
LordofAgents, line 842: Any reason that this was changed to baker?

tests/test_ban_tk_connect.py
FlorianSW: Hehe, good catch, dammit, wasn't meant to be committed. I'm running on Windows and for whatever reason python uses some windows charmap for strings, and this one does not support these unicode characters. As I wanted to have the tests run, I removed them, as they are not needed for the test itself 👍

tests/test_message_variables.py
FlorianSW: yes, it is, however, this is a mocking for the underlying function that is used to _fill_ the `next_map` variable. And this changed at some time from `next_map` to `get_next_map` in the `Rcon` class without being changed here :)

tests/test_no_leader_watch.py
FlorianSW: The test is supposed to test against a "small" squad (less than 6 players). Able in Allies however is a full squad in our test data (see somewhere at the top). Hence the test failed. Baker in Axis is a squad with 2 players only, which fits perfectly for the intention of the test :)

tests/test_ban_tk_connect.py
LordofAgents: Do you want to revert that commit to make sure that characters causes no issues?

LordofAgents: If you can't, I can check that.

FlorianSW: done :)

cemathey (approved):
Does this let us run a postgres instance in the github action? That's pretty awesome.

FlorianSW:
> Does this let us run a postgres instance in the github action? That's pretty awesome.

Basically yes. We can run anything in a docker container 😅

PR #813, opened 2024-12-21
do not use mixed as team side for player with marginal kills/deaths w…

FlorianSW:
…ith the "other" side

### DIFF
--- a/tests/test_player_stats.py
+++ b/tests/test_player_stats.py
@@ -111,7 +111,8 @@ def test_detect_ignores_some_other_side():
         'MG42': 1,
         'SATCHEL': 1,
         'BOMBING RUN': 1,
-        # this is a teamkill or a wrongly attributed kill and should not result in a MIXED team confidence
+        # this is a teamkill or a wrongly attributed kill or a kill by an explosion of a vehicle or something
+        # it should not result in a MIXED team confidence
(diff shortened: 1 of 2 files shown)


tests/test_player_stats.py
```diff
@@ -91,3 +91,28 @@ def test_detect_complex_inf():
...
+        'KARABINER 98K': 7,
+        'FG42 x4': 6,
+        '150MM HOWITZER [sFH 18]': 5,
+        'STG44': 3,
+        'COAXIAL MG34 [Sd.Kfz.181 Tiger 1]': 2,
+        '88 KWK 36 L/56 [Sd.Kfz.181 Tiger 1]': 2,
+        'MG42': 1,
+        'SATCHEL': 1,
+        'BOMBING RUN': 1,
+        # this is a teamkill or a wrongly attributed kill and should not result in a MIXED team confidence
```
Dorfieeee, line 114: it's not a teamkill nor wrongly attributed in my opinion. the player was probably killed by an explosion of the Sherman which must have been caused by any kind of explosive used by the opposite team eg. artillery, tank, satchel, at gun, ...

FlorianSW: let me add that as additional info to the comment 👍

LordofAgents (approved):

PR #815, opened 2024-12-21
Add optional columns to public stats page

FlorianSW:
image

### DIFF
--- a/rcongui_public/src/components/game/statistics/game-columns.tsx
+++ b/rcongui_public/src/components/game/statistics/game-columns.tsx
@@ -1,7 +1,7 @@
 'use client'
 
-import {ColumnDef} from '@tanstack/react-table'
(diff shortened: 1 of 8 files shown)


Dorfieeee:
I haven't checked yet but we also need to make sure it's mobile friendly

FlorianSW:
Looks as okay as before I would say. Without any optional columns (I already scrolled a bit to the right on both screenshots):
image
After: 
image

Dorfieeee:
Until you start searching for a player

FlorianSW:
> Until you start searching for a player

Good point. What about that:
image
And when searching:
image

It's basically just flex-wrap set to wrap 🤷

Dorfieeee (approved):
Good for now. Let's use localeStorage next time so the table can persist user's preferences.

PR #816, opened 2024-12-21
remove weapons data from frontend

FlorianSW:
It is not used anywhere, yet.
And if the frontend needs a list of all the weapons, it should request that list from the backend in the context of where the data is needed (e.g. for stats of a match, etc). There is no need to ship a lot of data in the bundle just because the data might be worthful in the future.

Secondly, the weapon data is already present in the backend, with the information that was previously covered by the data in the frontend. Maintaining both lists is an overhead which we can reduce.


cemathey:
I did request translations from T17; haven't heard back but they are on their holiday break still.

FlorianSW:
we could add translations to the backend or frontend, I do not care, however, for now there are none, so this data should be removed so we do not need to maintain it :D

cemathey (approved):

PR #830, opened 2024-12-29
Fix: Show only player that are currently blacklisted, fixes #811

### LINKED ISSUE #811: API: `get_players_history` blacklisted filter
Expected behaviour: When a player is not actively blacklisted, it should not be returned by the query.
Actual: It is returned as long as it has some blacklist record.
image

LordofAgents:
Hello,
this fixes #811 . The current implementation includes player where the blacklist record is already expired.


FlorianSW (approved):

FlorianSW:
@LordofAgents any idea of how this could be tested in an integration test? 🤔

LordofAgents:
Going to check that on Thursday

PR #834, opened 2024-12-30
Feature/stats player search

marcelhillesheim:
1. Removed unused utils (weapon related stuff got moved to backend)
2. Added missing hover effect for normal select (used for team select)
3. Display player count for team filter
4. Added new component select box and replaced existing player filter with it

team_filter_count
player_select_empty
player_select_overflow


rcongui_public/src/components/game/statistics/game-table.tsx
```diff
@@ -88,41 +104,30 @@ export function DataTable<TData extends Player, TValue>({columns, data, tableId}
...
-              <TagList
-                value={((table.getColumn('player')?.getFilterValue() as []) ?? [])}
-                onChange={(v) => table.getColumn('player')?.setFilterValue(v)}/>
-            </Fragment>
+            <SelectBox
+              options={playerFilterOptions}
+              multiple
+              value={playerFilter}
+              onChange={(values) => Array.isArray(values) && setPlayerFilter(values)}
+              placeholder={`${t('searchPlayer')}...`}
```
Dorfieeee, line 129: Missing `placeholderInput`

rcongui_public/src/components/ui/select-box.tsx
Dorfieeee, line 1: It would be cool if you could set `limitTags={2}` to display only 2 names and `+x`. Not necessary.

rcongui_public/src/components/game/statistics/game-table.tsx
Dorfieeee: It's breaking out of the layout on smaller screens.
I know this may not be an app heavily used on mobile phones but it would be nice to handle this somehow.
Perhaps use `useMediaQuery` and if small screen, limit visible tags to 1 & set overlay hidden.

image

Dorfieeee (commented):
That's exactly what I wished for. Great job.
If we could tweak the little details I'll approve.

rcongui_public/src/components/ui/select-box.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  className="absolute inset-y-0 right-0 flex cursor-pointer items-center pr-3 text-muted-foreground hover:text-foreground"
+                  onClick={() => setSearchTerm("")}
+                >
+                  <XIcon className="size-4" />
+                </div>
+              )}
+            </div>
+            <CommandList>
+              <CommandEmpty>
+                {emptyPlaceholder ?? "No results found."}
```
Dorfieeee, line 156: Also not translated.

Dorfieeee (commented):
Missing translation

rcongui_public/src/components/game/statistics/utils.ts
FlorianSW, line 1: This is untelated to this PR and bloats it up a bit. Once we get this one here reviewed, we can get rid of this change here in this PR:
https://github.com/MarechJ/hll_rcon_tool/pull/816

rcongui_public/src/components/game/statistics/game-table.tsx
marcelhillesheim: inputPlaceholder default value is "Search...". I changed it before to the same value as for the placeholder, but this doesnt look good/right.

Regarding the breaking layout: I tested the layout before i created the MR and i cant recreate the behaviour from the screenshot. I used chrome dev tools to test mobile resolutions. Its also weird that you still have the background image visible as this is the first element, which isn't visible anymore when you reduce the resolution. Do you have a local change?

I still added your suggestion with the limited tags for mobile as the bigger tag field lead to the popupover to be outside the screen.

rcongui_public/src/components/ui/select-box.tsx
marcelhillesheim: done

rcongui_public/src/components/game/statistics/utils.ts
marcelhillesheim: done

Dorfieeee (approved):
LGTM 👍🏻

PR #837, opened 2024-12-31
UI: VIP Forms

Dorfieeee:
Issue 1: Users cannot add additional text for the VIP description
Issue 2: Users cannot add "Never expires" VIP in the VIP settings view

Solution 1: Added `prefix` & `suffix` fields to the form with disabled example preview field
image

Solution 2: Added "Never expires" button & allowed very high date for the datetime field
image


Dorfieeee:
I want to reword this view at some point. This is just a temporary fix.

cemathey (approved):

PR #852, opened 2025-01-05
properly handle forward values (can be string or boolean, even though…

FlorianSW:
… frontend always uses strings :shrug:)


Sbosvk:
Just a thought, perhaps also covering strings of "1" and "0" as well?

Dorfieeee:
Checkbox uses string values for true and false

cemathey (approved):

PR #857, opened 2025-01-06
Feature/stats game table

marcelhillesheim:
fixes:
1. Columns would be sorted in asc order when clicking on the header 
2. removed unused (duplicated) component
3. fix broken ui above table

feature:
1. replaced collapsible in player stats with accordion
2. player names in encounters are now clickable
4. changed column alignments (right side for numbers)
5. refactored column hiding (use tanstack buildin hiding feature)
6. hiding ingame points by default
7. added kd column
8. added counts to category tooltip


Dorfieeee:
Looking great.

I like you can click on the player name now but I would personally prefer if the click selected the player in the table instead.
That is just an idea. If you don't share that with me that's fine.

I would like if a user could preserve his table's settings preference using the locale storage.

You can use https://github.com/MarechJ/hll_rcon_tool/blob/master/rcongui/src/hooks/useStorageState.js
or some better version of the hook.

Dorfieeee:
I am also thinking to change the default width of the main content since we're adding more and more content to it.
What do you think?

marcelhillesheim:
> Looking great.
> 
> I like you can click on the player name now but I would personally prefer if the click selected the player in the table instead. That is just an idea. If you don't share that with me that's fine.
> 
> I would like if a user could preserve his table's settings preference using the locale storage.
> 
> You can use https://github.com/MarechJ/hll_rcon_tool/blob/master/rcongui/src/hooks/useStorageState.js or some better version of the hook.

I wrote you on discord regarding some issues. 

I like your playerclick approach more, so i reverted the player click. I will do a new PR for the updated playerclick.

Same with local storage would do that in a new PR.

FlorianSW (approved):

PR #860, opened 2025-01-08
Feature/horizontal game list performance

marcelhillesheim:
1. performance improvements (down from avg. 70ms to avg. 10ms on local test): 
- memo gamecard
- memo expensive calc 
- extracted ref from gamecard as it would force rerender 

2. refactoring: 
- inverting hiddenGames (gamesVisibility)
- creating cards array instead of conditional render of datecard in mapping function
- using only one ref for gamecard
3. fixed: 
- due to z-index sticky sticky date would cover navigation bar when scrolling down


rcongui_public/src/pages/games/horizontal-games-list.tsx
```diff
@@ -1,41 +1,56 @@
...
 import {ScrollArea, ScrollBar} from "@/components/ui/scroll-area";
 import {validGame} from "@/pages/games/list";
 import MapFigure from "@/components/game/map-figure";
-import dayjs from "dayjs";
+import dayjs, {Dayjs} from "dayjs";
 import {cn, dayjsLocal} from "@/lib/utils";
 import {useTranslation} from "react-i18next";
 import {useLocale} from "@/i18n/locale-provider";
 
+const pathRegex = (gameId: number) => new RegExp(`^/games/${gameId}(?:/|$)`);
```
FlorianSW, line 12: Why a regex, when this is can be a simple startsWith? 🤔 Sounds like an overkill, and the browser needs to compile the regex for each game again, which sounds a waste of resources :(

rcongui_public/src/pages/games/horizontal-games-list.tsx
```diff
@@ -1,41 +1,56 @@
...
+
 export const HorizontalGamesList = ({ games }: { games: ScoreboardMaps }) => {
   const { pathname } = useLocation()
-  const [hiddenDates, setHiddenDates] = useState<Record<string, boolean>>({});
+  const [gamesVisibility, setGamesVisibility] = useState<Record<number, boolean>>({});
+  const gamesVisibilityRef = useRef(gamesVisibility);
```
FlorianSW, line 17: for what is this actually needed? Wouldn't `gamesVisibility` be enough in all cases where the ref is currently used?

FlorianSW (commented):
some questions, shouldn't be a blocker for a merge, but I would be interested to know your opinion on them before approving :)

rcongui_public/src/pages/games/horizontal-games-list.tsx
marcelhillesheim: The games are memorized. The issue with startswith is:
path is '/game/23/stats'

path starts with '/game/23' AND '/game/2'

FlorianSW: Then let's split by / and compare the ID if it is the same. A regex is really an overkill for this use case I assume ❤️

Dorfieeee (approved):
LGTM 👍🏻

FlorianSW (changes requested):
see regex comment

rcongui_public/src/pages/games/horizontal-games-list.tsx
marcelhillesheim: done

FlorianSW (approved):

PR #862, opened 2025-01-08
FE: stats: focus player in table on name click

marcelhillesheim:
focus row in table, when clicking on names playerdetails (encounters) and also the name in the headline (currently selected player)


FlorianSW (changes requested):
When you click on the name, the table cell gets a border. This results in the whole table to get smaller in width and leads to a visual refresh of a big portion of the page. This looks bad 😅 This is also the case, when the border/highlight is removed again.
Can you make box-sizing include the border to that the table does not "move" when the name is clicked?

marcelhillesheim:
> When you click on the name, the table cell gets a border. This results in the whole table to get smaller in width and leads to a visual refresh of a big portion of the page

Didn't notice this. Will take a look. Thx

marcelhillesheim:
> Can you make box-sizing include the border to that the table does not "move" when the name is clicked?

I did try box sizing, but the issue is that the row already has a border. Ended up with outline as the easiest way to fix it.

FlorianSW (approved):
Awesome, thanks for the fix! :)

PR #864, opened 2025-01-09
Feature: custom language en24h

marcelhillesheim:
Many people prefer english ingame, but are used to 24h DD MM YYYY format. 

Adds custom language 'en24h'

in i18n it acts exactly like 'en'

in dayjs it combines the 'en' translations for weekdays, months etc with the 'it' date format.


Dorfieeee (approved):
I hope we now don't need two separate eng translations. Otherwise it looks good.

marcelhillesheim:
> I hope we now don't need two separate eng translations. Otherwise it looks good.

Nope it just uses the existing 'en' translations. Its just a marker to use the custom locale format in dayjs.

PR #868, opened 2025-01-11
Feature/weapon category improvements

marcelhillesheim:
1. added a gap between category bars to make it easier to read 
2. tooltip can be activated on mobile via long press
3. added sniper as category

### DIFF
--- a/rcongui_public/src/components/game/statistics/weapon-type-bar.tsx
+++ b/rcongui_public/src/components/game/statistics/weapon-type-bar.tsx
@@ -59,7 +59,7 @@ export const WeaponTypeBar = memo(({ totalKills, killsByType }: WeaponTypeBarPro
                     background: simpleWeaponTypeMap[type as SimpleWeaponType].color
                   }}
                 />
-                {(totalCategorizedKills < totalKills || index + 1 < displayedBars.length) && <div style={{width: GAP_SIZE + 'px'}}/>}
+                {index < gapAmount && <div style={{width: GAP_SIZE + 'px'}}/>}
               </React.Fragment>
               }
             )}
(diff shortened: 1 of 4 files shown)


rcongui_public/src/components/game/statistics/weapon-type-bar.tsx
```diff
@@ -11,19 +11,27 @@ type WeaponTypeBarProps = {
...
   if (!totalKills || !killsByType) {
     return null;
   }
 
+  /**
+   *  Sorted by a specific order. Please consider this before sorting it by value:
+   *  It is sorted from direct to indirect weapon categories
+   *  Sniper is super accurate, infantry less accurate, mg is more spray and then explosions sorted by splash size
+   */
   const simpleWeaponTypeMap: Record<SimpleWeaponType, {color: string, order: number, t: string}> = {
```
FlorianSW, line 27: if it should have a specific order, why not using an array btw? I know this is not something you did in this PR. However, I still want to bring it up.

Maintaining an order property is pretty fragile if done manually. An array looks like a perfectly fine structure here as well, especially as we use the access by weapon type only at one place, and the data size is relatively small. Hence a `find()` on an array sounds like a viable solution as well. While dropping the need for someone to read this comment to know that this should be ordered :D

FlorianSW (changes requested):
Let me know what you think :)

rcongui_public/src/components/game/statistics/weapon-type-bar.tsx
marcelhillesheim: The issue i find with getting the order from the array order is the following: Someone can add an element without intentionally setting the order. They might just add the element to the end even if the element would fit more in another position. 

Because the current order might seem random, i want to make sure that future devs are knowlingly setting an order.

FlorianSW (approved):
ok, then let's keep it like that :)

PR #869, opened 2025-01-11
FE: Stats: fixed: table rerender

marcelhillesheim:
Public info is refetched every 15 seconds. 

This would rerender the whole table -> moved download hook in its own component

### DIFF
--- a/rcongui_public/src/components/game/statistics/download-button.tsx
+++ b/rcongui_public/src/components/game/statistics/download-button.tsx
@@ -1,0 +1,37 @@
+import React from 'react'
+
+import { Player } from '@/types/player'
+import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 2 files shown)


FlorianSW:
Stupid question from my side maybe: Why would the download hook re-render the whole page again? 🤔

Dorfieeee:
@FlorianSW the hook is dependant on publicInfo query so whenever that query updates, the hook updates and so does the table.

Dorfieeee (approved):
Good catch!

FlorianSW:
> @FlorianSW the hook is dependant on publicInfo query so whenever that query updates, the hook updates and so does the table.

Ah, didn't see the use of the publicInfo hook. Thanks :)

PR #870, opened 2025-01-11
Feature/public stats awards

marcelhillesheim:
Added awards to public stats for completed games


Dorfieeee (approved):
The code runs, pages don't break.. if there are any inconsistencies I believe some people will eventually speak out

PR #872, opened 2025-01-11
Reduce time until stats are available after game start

FlorianSW:
Because of the architecture of the Community RCon tool, we had a fixed wait time of 6 minutes before stats of a game were recorded and are available in the stats pages (admin and public ones).
The main reason is, that the log recorder ran every 5 minutes to take up logs procesed by the Log loop and dumped them into the database. The record stats worker used the logs in the database to calculate the stats of players. To not loose any log events from the end of the game, the recorder waited 6 minutes. This should've ensured that all logs from the game were taken up by the log recorder and dumped to the database.

This was kind of a flawed mechanism already, as the log recorders frequency, when it picked up logs from the loop, was configurable. If someone changed the config in the supervisor to a longer time, some logs still might have been missed from a recent game when it got recorded.

This Pull request aims to address this in multiple ways:
- Log Recorder now runs every 10 seconds by default, as opposed to the 5 minutes before. This increased the load on the database in that way that it is run more often. However, the amount of data that is written is much less compared to 5 minutes of log lines. This should therefore "just" spread out the load over the timeframe.
- The log recorder was also rewritten to optimise how the player profile reference of a log line is loaded. Instead of loading it again for each log line one by one, the profile IDs are fetched for the whole batch of log lines in a single query. This should drastically reduce the load on the database (reducing the operation from O(n) to O(1) every 10 seconds)
- the record stats worker now tries to evaluate if the log events in the database are complete. This is indicated by either a MATCH ENDED event recorded, or at least two MATCH START events. The first should happen when a match ends normally, the second when the map is changed and possibly when the objectives are changed (-> this part is something I didn't test, yet, but needs heavy testing in a production-like environment)

It also refactored the log components in the rcon package to make it more readable and easier to understand. Instead of one big huge file 😅


cemathey (approved):
I gave it a skim; didn't see anything obvious. Great idea to look for consecutive match starts. Did you by chance build/test this on a live version and let it record some real games to make sure it works?

FlorianSW:
> Did you by chance build/test this on a live version and let it record some real games to make sure it works?

It's running since yesterday on a public live server, which is seeded for quite a lot of rounds now. No issues so far. And each match was coming in within seconds after it ended and the stats (at least the ones I could check ofc) were all corret 👍

PR #876, opened 2025-01-13
Feature/streamer view

marcelhillesheim:
Added new route to live view (same structure as with stats route for completed games)

Live view -> twitch button

The new route is for streamers and contains an infinite scrolling top player banner.

image

### DIFF
--- a/rcongui_public/src/i18n/locales/en/translation.json
+++ b/rcongui_public/src/i18n/locales/en/translation.json
@@ -35,6 +35,7 @@
     "pauseWidth": "Pause width",
     "text": "text",
     "showAvatars": "Show avatars",
+    "showWeapons": "Show weapons",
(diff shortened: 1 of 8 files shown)


Dorfieeee:
How is this suppose to be used?

marcelhillesheim:
> How is this suppose to be used?

Streamers can add sources to their streams with software like OBS. So they can just screen capture the banner and key out the white to make it transparent. 

I know one streamer, who is already hyped about it :D

FlorianSW:
@marcelhillesheim I just merged this, however, there is one request I would like you to take :) Can you add some documentation maybe? Like a step-by-step guide on how this can be used in OBS? And then make a PR to link this docs in the public stats page for people to find? ❤️

PR #877, opened 2025-01-13
Trick JSON forms into rendering the admin ping settings correctly

cemathey:
The UI is using JSON forms to generate elements based off the Pydantic schema the user config models generate.

It was causing issues with the admin ping settings because it did not think that `user_mentions` and `role_mentions` were required items because we default those to empty lists on the backend.

This removes the defaults, but still allows the model to be instantiated without passing them (important for resetting or generating new blank configs) by creating an empty list if they aren't provided.

The UI would also pass empty `""` strings for user/role mentions if not provided, which would fail validation so we simply filter those out.

### DIFF
--- a/rcon/user_config/log_line_webhooks.py
+++ b/rcon/user_config/log_line_webhooks.py
@@ -46,8 +46,8 @@ class LogLineWebhookUserConfig(BaseUserConfig):
 
             hook = DiscordMentionWebhook(
                 url=raw_webhook.get("url"),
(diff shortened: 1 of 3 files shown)


FlorianSW:
@cemathey can you please take a look at the failing tests? ❤️

cemathey:
> @cemathey can you please take a look at the failing tests? ❤️

fixed, thanks for adding those automated checks that is great.

FlorianSW (approved):

PR #878, opened 2025-01-14
UI: Theming & Layout change

Dorfieeee:
- A user can now select from a different color scheme
  - Set up the code so more color schemes can be added 
- Removed app header

This PR also fixes:
- team-view icons path issue
- records > players scroll issue

image
image
image


FlorianSW (changes requested):
When on the login page the first time with this PR applied, the login page is in light mode, while my system default should be dark. It burned my eyes a bit :P
Once logged in the theme switched to dark.

This should really be the case on the login page as well :)

Dorfieeee:
The AppTheme component now wraps the whole App.
That resolves the issue having loading, app error and login pages without theming.

FlorianSW (approved):

PR #879, opened 2025-01-14
Allow customizable CSRF token and session cookie domains

robbmanes:
Pass through the CSRF_COOKIE_DOMAIN and SESSION_COOKIE_DOMAIN options as an environment variable.  This is passed to the backend and read as other options are.

This is particularly useful for RCON's hosted on multiple subdomains sharing login credentials since the CSRF token or sessionid from one RCON instance can be used on another.

A potential pitfall of this option is if other applications are using CSRF or sessionids with wildcard domains then they could potentially conflict/cause false errors for logins that are not the RCON itself. To solve that, we'd have to make our csrftoken and sessionids unique to the RCON which seems to be a rare and unusual case for most users of the RCON.

Frankly, I'll probably follow this up with setting `CSRF_COOKIE_NAME` and `SESSION_COOKIE_NAME` in the django instance to something unique to cRCON so we won't ever hit the above pitfall.

Tested working on our environment, feedback appreciated.


default.env
```diff
@@ -87,6 +87,22 @@ NB_API_WORKERS=1
 # The default is 8 and should be sufficient for a large team of moderators
 NB_API_THREADS=8
 
+# Use this *only* if hosting your RCON behind a reverse proxy or on subdomains. Otherwise
+# leave it empty.
+#
+# If hosting on multiple subdomains (e.g. rcon1.example.com, rcon2.example.com, etc),
+# cookies and therefore login tokens are not shared amongst the different subdomains. This
+# value tells the RCON server to append a domain to these cookies so they can be shared among
```
FlorianSW, line 95: > to append a domain to these cookies
That doesn't make much sense, does it? The Cookies always have a domain set, by default it's the very same domain where the website is hosted.

These settings are _overwriting_ this domain to the value you specify. This should be reflected in the comment. Please also add the documentation link from Django :)

https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-SESSION_COOKIE_DOMAIN
https://docs.djangoproject.com/en/5.1/ref/settings/#csrf-cookie-domain

FlorianSW (changes requested):

LordofAgents (changes requested):
This options can be a security risk if users set it to a domain that they don't control fully. Also, if one subdomain get's under control of someone evil, the session token could also be stolen. 

I would like to add the following warning (after someone read over it):
> WARNING: DO NOT set this to a domain that you do not fully control! Also keep in mind that if a subdomain is taken over by someone, they are able to steal the session token once the user browses the malicious subdomain.

robbmanes:
I agree with said warning and also am testing this on a *new* setup, please hold merge,  I will squash the PR and resbumit with a warning.

default.env
robbmanes: Agreed, and resolved for re-review.  Sorry for the phrasing, I assume I was insane at the time.

robbmanes:
Squashed and repushed, ready for review.  I did as much e2e testing manually as I could think of but couldn't think of a good way to automate any use cases, open to suggestions.

default.env
robbmanes: > The Cookies always have a domain set, by default it's the very same domain where the website is hosted.

This actually doesn't seem to be true in my testing.  [W3 claims the domain specifier is optional](https://www.w3.org/Protocols/rfc2109/rfc2109) and Django, if no `SESSION_COOKIE_DOMAIN` is set, leaves the `domain` field empty; it only appears in the browser because the browser infers it should be set (for obvious security), so effectively you're still right and I'm being pedantic.  Just documenting for my own knowledge/pitfalls, the cookie will be assigned by the *client* in scenarios when it's empty on-wire, and both perform evaluations/sets.  I'm not a webdev at all so this is news to me and it's *very* likely I misunderstand something.

FlorianSW (approved):
okay for me.

LordofAgents (approved):

PR #883, opened 2025-01-16
FE: games table improvements (added columns, small fixes)

marcelhillesheim:
1. added gamemode column and added attacker side for offensive gamemode.
2. added weekday column
3. refactored game duration util and removed leading zero from game time 01:30:00 to 1:30:00
4. added weather icon column


Dorfieeee (approved):
Oh yeah this is looking great :]

Only one suggestion: They may add another weather, mode, etc. so I would personally have some fallback options. I don't know if cloning undefined and trying to attach an attribute to it would break the app tbh. Just something to bear in mind as the FE does not always need to be in sync with BE.

FlorianSW (approved):

PR #885, opened 2025-01-16
Add webhook service container to handle rate limiting/etc.

cemathey:
* Adds a new container `webhook_service`
* Updates python from `3.11` to `3.12` (`pydantic` requires a different `TypedDict` than the one in `typing` before `3.12`)
* Adds new env variables (defaults are set; if people don't add them they will only get the Docker warnings, but it will still work) so people can tune their service without building their own images
  * HLL_WH_SERVICE_RL_RESET_SECS
  * HLL_WH_SERVICE_RL_REQUESTS_PER
  * HLL_WH_SERVICE_RL_TIME_WINDOW
  * HLL_WH_MAX_QUEUE_LENGTH
* Adds some extra checks so that the service container can run without all of the normal backend variables being set (allows us to do imports)
* Updates `cache_utils` to allow using the global redis database (database number 0) more easily; this is the shared one that is used by all CRCON instances to manage forwarding requests
* Updates everywhere (except scorebot because this doesn't support editing messages yet) to use the new `enqueue_message` function to add webhooks to the service queues
* Removes `rcon.discord_asyncio` because it is no longer used anywhere
* Adds two new permissions `can_view_webhook_queues` and `can_change_webhook_queues`
* Adds the following endpoints:
  * `get_webhook_queue_overview`
  * `get_all_webhook_queues`
  * `get_webhook_service_summary`
  * `reset_webhook_queues`
  * `reset_all_webhook_queues_for_server_number`
  * `reset_webhook_queue`
  * `reset_webhook_queue_type`
  * `reset_webhook_message_type`

Okay; so the tl;dr of this is that we add a new webhook service which uses `redis` to let other parts of CRCON enqueue a message when they want to use a webhook and then move on with their life.

The queues are structured so their redis keys contain the server number, the webhook type (discord, etc.) and the message type so they can be selectively targeted.

The service container will run on startup and polls the queues, sends them and monitors the response headers; re-enqueuing rate limited responses.

If a queue exceeds the max (default 150 messages) it will drop older messages; this is unlikely outside of the kill feed and maybe the API itself if it gets hit with a burst of requests.

I tested this locally and it seems to work well, for the UI the `get_webhook_service_summary` returns a response similar to which is nested objects by server number, then webhook type (currently only `discord` is supported) and the underlying webhook message type, the `redis_key` is the queue ID that is used to target individual queues (for clearing them):

```json
{
    "result": {
        "is_globally_rate_limited": false,
        "global_rate_limit_reset_after": null,
        "queues": {
            "1": {
                "discord": {
                    "log_line_chat": [
                        {
                            "server_number": 1,
                            "id": "1156786342429204510",
                            "webhook_type": "discord",
                            "message_type": "log_line_chat",
                            "length": 1,
                            "rate_limit_bucket": null,
                            "rate_limited": false,
                            "redis_key": "whs:1:discord:log_line_chat:1156786342429204510"
                        }
                    ],
                    "log_line_kill": [
                        {
                            "server_number": 1,
                            "id": "1157061127470063727",
                            "webhook_type": "discord",
                            "message_type": "log_line_kill",
                            "length": 9,
                            "rate_limit_bucket": null,
                            "rate_limited": false,
                            "redis_key": "whs:1:discord:log_line_kill:1157061127470063727"
                        }
                    ]
                }
            }
        }
    },
    "command": "get_webhook_service_summary",
    "arguments": {},
    "failed": false,
    "error": null,
    "forward_results": null,
    "version": "v11.1.0-113-ga36037e3"
}
```
In the future I'd like to explore supporting other types of webhooks (`guilded` webhooks are basically identical to `discord` ones I think and people have expressed interest in that in the past), being able to bundle up similar types of messages so we can send more data without rate limiting (like bundling up to 10 embed kills in 1 message) and editing existing messages, because we can't use this for scorebot at the moment.


cemathey:
While thinking about it tonight; I think this would be pretty easy to update to support editing existing messages (for instance for scorebot) by requiring the place that wants to edit a message to make a traditional sync (or async if it's some utility running outside of our traditional sync based Django CRCON setup) request to make a message in Discord; then take the message ID and use that for subsequent requests with the webhook service, and add a parameter to indicate that it's editing versus creating a brand new message.

requirements.txt
```diff
@@ -10,7 +10,8 @@ simplejson==3.19.3
 psycopg2-binary
 sentry-sdk[django]==2.19.2
 python-dateutil==2.9.0.post0
-discord-webhook[async]
+# discord-webhook[async]
+git+https://github.com/cemathey/python-discord-webhook@132fc7b7a58ba25e7fffd71cb363e4748c82941c
```
FlorianSW, line 14: Can you make a PR for the necessary changes upstream, so we can eventually go back to the official one? :)

FlorianSW (approved):
looks good, however I was just looking from a high level of details. It all looks to make sense .

PR #886, opened 2025-01-17
Feature/additional columns

marcelhillesheim:
1. added additional columns
2. created new column selection menu
3. disabled some columns by default


Dorfieeee (dismissed):
On smaller screens render column categories in a column?
On larger screens render column selection popover aligned in the middle since it's quite wide? [API Docs](https://www.radix-ui.com/primitives/docs/components/popover#api-reference)

marcelhillesheim:
> On smaller screens render column categories in a column? 

The issue could be that on small screens the user rotates screen for table. So it should work for both orientations. I would just let flex-wrap deal with it. Also when its not one single column for the categories, you can still see the table while adjusting the columns.

> On larger screens render column selection popover aligned in the middle since it's quite wide? [API Docs](https://www.radix-ui.com/primitives/docs/components/popover#api-reference)

Mmmh in the middle it would cover up the table, which the user likely wants to see while toggling column visibility.

I made one change: 
On desktop the popover is anchored to the right of the button so the popover doesnt cover the table.
On mobile it is anchored to bottom as there is no sidebar.

FlorianSW:
Any update here, @Dorfieeee are the changes @marcelhillesheim did ok for you? :)

@marcelhillesheim could you please resolve the merge conflicts? ❤️

FlorianSW (approved):

PR #891, opened 2025-01-18
Update player_history.py

cemathey:
Made a typo a few months ago; apparently this function is currently unused.


ElGuillermo (approved):

FlorianSW (changes requested):
If it is unused (and my IDE confirms that it is not used anywhere), then we should remove, not fix it :D

ElGuillermo (approved):

FlorianSW (approved):

PR #892, opened 2025-01-18
UI: Live view QOL changes

Dorfieeee:
A lot of changes in this PR

- Created fake API for the live view using miragejs - it mocks the team_view and recent_logs endpoints at the moment
- Locale storage changes for the live view - this may be later migrated to use zustand that supports migrations
  - Going back to previous version **WILL BREAK** the live view and other views including the same tables. That can be solved by manually clearing the localStorage.
- NEW: Info panel mentioning missing commanders, SLs, vip and watched counts, next map, player lobby counts
- Player sidebar now includes info about watchlist and VIP info from all servers
- Fixes several bugs mainly about logs table configuration and filtering
- NEW: Log and Player columns
- NEW: Hide/Show columns
- NEW: Filter logs by team


LordofAgents:
First thank you for your work as always.
> Created fake API for the live view using miragejs - it mocks the team_view and recent_logs endpoints at the moment
Are you sure that this is a path we should go? Shouldn't we change or add an api method that supplies the information's you need?

Dorfieeee:
@LordofAgents this way I can 
- try my own use cases - various names, ids, ...
- not needing to run crcon instance to get data into the app
- not needing server with players

Some of the bugs there were for the reason of not having any activity on the server. For example no new logs coming - did not notice the Tanstack Table is set to reset to page 1 on data update.

rcongui/package.json
```diff
@@ -22,6 +23,7 @@
     "@dnd-kit/utilities": "^3.2.2",
     "@emoji-mart/data": "^1.2.1",
     "@emoji-mart/react": "^1.1.1",
+    "@faker-js/faker": "^9.3.0",
```
FlorianSW, line 26: why not dev dependency? 🤔

rcongui/package.json
```diff
@@ -50,6 +52,7 @@
     "immutable": "^5.0.3",
     "localforage": "^1.10.0",
     "lodash": "^4.17.21",
+    "miragejs": "^0.1.48",
```
FlorianSW, line 55: shouldn't this be a dev dependency as well?

rcongui/package.json
Dorfieeee: Most definitely

rcongui/package.json
Dorfieeee: Move to dev dependencies

rcongui/package.json
Dorfieeee: Moved to dev dependencies

Dorfieeee:
I've rebased these changes on top of 11.3.0 and
- Moved locale storage configs to be managed by `zustand` for future updates and better migrations(just like db)
- Moved miragejs & faker to dev dependencies
- Added toggle for log query filters & search input

FlorianSW (approved):
> Info panel mentioning missing commanders, SLs, vip and watched counts, next map, player lobby counts

Can't seem to find this feature :/ Apart from that, when running it, looks okay. Didn't look into all changes :(

PR #909, opened 2025-01-31
Update models.py

ElGuillermo:
Just added the `%Y` to the strings (and inverted month and day)

### DIFF
--- a/rcon/models.py
+++ b/rcon/models.py
@@ -911,14 +911,14 @@ class BlacklistRecord(Base):
         variables = {
             "player_id": self.player.player_id,
             "player_name": self.player.names[0].name if self.player.names else "",
-            "banned_at": self.created_at.strftime("%d %b %H:%M"),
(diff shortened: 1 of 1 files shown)


FlorianSW (approved):
Makes an assumption that the month should be "before" the date, which is not true in most os the world iirc, but I'm ok with that :)

ElGuillermo:
Well... I'm french, so the "normal" date is hh:mm dd/mm/yyyy... but I assume we should stick to the "most used" one :/

FlorianSW:
> Well... I'm french, so the "normal" date is hh:mm dd/mm/yyyy... but I assume we should stick to the "most used" one :/

Hard to pick the correct default here, I would be fine with either way. So, if you want you can also switch from M/D/Y to D/M/Y :)

FlorianSW:
If you want to change the format, feel free to open a new PR :)

ElGuillermo:
No, but this has to be reworked anyway, as "humanized" durations don't display right :/

PR #914, opened 2025-02-03
UI: VIP View version 1

Dorfieeee:
Very needed update for the VIP view.

As mentioned, this is the version 1 as some other features may be added in the future with updated backend and making forms the same across the app, eg. add, edit, search players.

- [ ] - Remove old file

image


FlorianSW (approved):
Some comments for revision 2 maybe :D

- clicking the edit button adds the current VIP entry to the "Add VIP" form fields. Does this really "edit" the VIP entry? And even if, it took me some seconds to realise this, probably not the best UX so far? 🤔 
- When scrolling with a big list of VIPs, would it be possible to have the header fixed?

FlorianSW:
@Dorfieeee any comment to my above points? :)

PR #916, opened 2025-02-07
Update __init__.py

cemathey:
Import the TK auto ban function so that it will actually run; if we don't import it is never added to the hooks.


ElGuillermo:
?! It has always ran fine on mine ?!
image

cemathey:
> ?! It has always run fine on mine ?! image image

That must be something to do with your local changes.

If we add logging to the hook:

```python
def on_tk(func):
    logger.info(f"Adding {func} to on_tk")
    HOOKS[AllLogTypes.team_kill.value].append(func)
    return func
```

Then we can see when it's actually run, when Python imports a file it executes all the code in it which is why all the hooks need to be imported in some way in `rcon.__init__.py` so that they're added to the list of hooks to run for a log type.

Without changing `__init__.py` and running CRCON nothing gets logged, indicating it isn't imported.

With the import:

```
[2025-02-15 03:35:41,163][INFO][MyServer1][v11.3.0-22-gb84b6e62] rcon.logs.loop loop.py:on_tk:72 | Adding <function update_tks at 0x7e920c7e1940> to on_tk
[2025-02-15 03:35:41,418][INFO][MyServer1][v11.3.0-22-gb84b6e62] rcon.logs.loop loop.py:on_tk:72 | Adding <function handle_on_tk at 0x7e920c02d6c0> to on_tk
[2025-02-15 03:35:41,419][INFO][MyServer1][v11.3.0-22-gb84b6e62] rcon.logs.loop loop.py:on_tk:72 | Adding <function auto_ban_if_tks_right_after_connection at 0x7e920c02d8a0> to on_tk
```

ElGuillermo:
I'm pretty sure I don't have any code change related to that (?) But I don't see any place it's called either. Some heavy voodoo magic maybe ?

ElGuillermo (approved):

PR #921, opened 2025-02-09
UI: Update rcongui project configuration and README

Dorfieeee:
- Remove .env file from version control
- Add example.env with environment configuration
- Update README with new project setup instructions
- Modify Vite configuration to use environment variables

### DIFF
--- a/rcongui/README.md
+++ b/rcongui/README.md
@@ -1,68 +1,82 @@
-This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+# CRCON GUI
 
-## Available Scripts
+A React-based web interface for managing Hell Let Loose game servers.
 
-In the project directory, you can run:
+## Tech Stack
 
-### `yarn start`
+- React 18
+- Vite
+- React Router 6
+- Material UI 6
+- Tanstack Query 5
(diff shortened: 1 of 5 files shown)


rcongui/README.md
```diff
@@ -1,68 +1,82 @@
...
+```
 
-Builds the app for production to the `build` folder.<br />
-It correctly bundles React in production mode and optimizes the build for the best performance.
+3. Install dependencies:
 
-The build is minified and the filenames include the hashes.<br />
-Your app is ready to be deployed!
+```bash
+npm install
```
FlorianSW, line 33: shouldn't this better be `npm ci` as long as one don't want to add/change dependencies? 🤔

FlorianSW (commented):
just one comment :)

rcongui/README.md
FlorianSW: Uhm, yeah, and? 😅 Sorry, I really don't get it :D The `package-lock.json` still includes all the development packages. When running `npm ci` without the `--omit=dev` flag, then all packages will be installed :)

Just want to make sure that people do not get changes in their working tree unintentional.

PR #929, opened 2025-02-11
UI: Updated views for Game & Audit logs

Dorfieeee:
Complete rewrite for both Game & Audit logs views.

## Audit logs view
- Detailed audit log view card

## Game logs view
- Utilizes the same table as in Live view
- Log action autocomplete
- Server filter select

### Both pages use URL search params
- you can share your search options using the same url
- when you navigate back/forward it can remember the previous search inputs

> Game Logs View
image

> Audit Logs View
image


rcongui/src/pages/dashboard.jsx
```diff
@@ -197,6 +198,299 @@ const GamesCard = ({ games }) => {
   );
 };
 
+const MapStatsCard = ({ games }) => {
```
FlorianSW, line 201: looks like this should be part of another PR (the dashboard refactor one?)

FlorianSW (approved):
just a question, didn't look at all the changes, though.

PR #936, opened 2025-02-13
UI: Add global navigation progress indicator

Dorfieeee:
UX: Added global loading navigator when routing & submitting router based forms eg., records->players.

image


FlorianSW (approved):
Finally (I had this in mind multiple times, but always forgot about it 😬)! ❤️

PR #939, opened 2025-02-14
UI: Player Profile Working Draft

Dorfieeee:


Dorfieeee:
It's not perfect but still better than the current version of the profile.
There can be many more things to be done later.

cemathey (approved):

PR #941, opened 2025-02-16
UI: Add column with colorfull name and IDs

LordofAgents:
Hello,

this is the second attempt to successfully create a PR:
This PR addeds another column on the Live View based on the already created column `message_colored`, which includes the IDs on Kills.
This was requested by admins, who needs to copy IDs fast. 

Thank you!


FlorianSW (changes requested):
Puh, idk if that is the right approach tbh 🙈 Now we have three distinct columns, message, message colored and message colored with IDs :/ Wouldn't it be possible ti just merge the latter ones?

LordofAgents:
Did merged it, please review again. 
Thx

PR #946, opened 2025-02-17
Overhaul scorebot; support editing messages in the webhook service; m…

cemathey:
* Overhauls scorebot
* Updates the webhook service to support editing messages
* Updates the compose templates to use a docker volume for the new scorebot
  * This is backwards compatible; I tried it both ways but for new installs/anyone who updates it will cut down on their docker filesystems growing because of the layered filesystem docker uses for changes
* Updates `entrypoint.sh` to port legacy scorebot public stat URL/webhook URLs, we can remove this in a few releases
* `get_scorebot_config` is still available as an endpoint so people can easily get their old settings
* Renames `rcon.scoreboard.py` to `rcon.player_stats.py`  to be more clear (and I wanted to use the `scoreboard` name
* The new scorebot system is `scoreboard` to differentiate/make troubleshooting easier when supporting users
* There are some miscellaneous code formatting changes and I also improved some miscellaneous typing here and there

So legacy scorebot used a single message and didn't contain a map rotation.

This splits the message into three sections, the header/game state, map rotation and player stats to help get around inherent Discord API limitations on message length/etc. This allows people to show off more player stats if they prefer.

It's very customizable; people can turn off individual sections/translate text for the different sections.

It also now works internally with an RCON instance instead of making requests through the docker network; the only mandatory settings are a public stats port, and at least one webhook URL which should be easier for users to configure.

I tested this locally and it seems to work just fine.

image
image
image


config/supervisord.conf
```diff
@@ -76,10 +76,10 @@ startretries=100
 startsecs=1
 autostart=true
 
-[program:scorebot]
-command=python -m rcon.scorebot
-environment=LOGGING_FILENAME=scorebot_%(ENV_SERVER_NUMBER)s.log
-startretries=24
+[program:scoreboard]
```
FlorianSW, line 79: When doing a release, this should be explicitly mentioned. Some people (including me) have their "own" supervisord.conf, where they need to make these changes well

docker-templates/one-server.yaml
```diff
@@ -63,6 +66,7 @@ x-supervisor: &supervisor
     - ./logs:/logs/
     - ./supervisor_data:/data
     - ./config:/config/
+    - scoreboard_db:/scoreboard_db
```
FlorianSW, line 69: another thing that _needs_ to be documented, people are supposed to copy-paste this file iirc.

rcon/maps.py
```diff
@@ -122,6 +122,27 @@ class Faction(pydantic.BaseModel):
     name: str
     team: Team
 
+    if TYPE_CHECKING:
```
FlorianSW, line 125: wow, I didn't know about that :)

FlorianSW (approved):
Looks good. Comments are just reminders to add some notes to the release notes of the next release to ensure people take over necessary changes to their files they might have copied.

PR #950, opened 2025-02-19
UI: Game Logs timezone, 24-hour format date pickers

Dorfieeee:
- Add timezone plugin to dayjs in App.js and game-logs page
- Modify date-time pickers to use 24-hour format (ampm=false)
- Update game logs column to convert event time to local timezone
- Remove explicit UTC timezone setting in players records page


Dorfieeee:
Please, try it out first if you can before approving.

cemathey (approved):
Worked for me when I built/launched it locally and poked around the logs page/player records page

PR #951, opened 2025-02-19
Hotfix/webhook service

cemathey:
* Set the webhook service to always restart
* Correct our `SCAN` method to use `scan_iter`; previously we were not updating the cursor and always scanning from the same starting location which would fail when not all of the keys could be returned in one pass
* Handle HTTP 404 errors for non-existent messages

### DIFF
--- a/rcon/webhook_service.py
+++ b/rcon/webhook_service.py
@@ -106,6 +106,7 @@ class DiscordErrorResponse(TypedDict):
 
     http_401: bool
     http_403: bool
+    http_404: bool
 
 
 class WebhookMessageType(StrEnum):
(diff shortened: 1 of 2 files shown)


rcon/webhook_service.py
```diff
@@ -468,9 +491,6 @@ async def _dequeue_message(
             datetime.now().timestamp() + bucket_data.reset_after_secs
         )
 
-        bucket_id = bucket_data.id[-4:] if bucket_data and bucket_data.id else "N/A"
```
FlorianSW, line 471: what was this code for tbh? 🤔 And why remove it?

PR #957, opened 2025-02-20
Fix Live Stat Calculation: Handle changing names in round

LordofAgents:
Hello,
someone found a bug, which occurs, if a player leaves a round and rejoins with another name in the same round.
This led to multiple player with the same id, which led to a bug in the frontend framework.

This PR does fix this in the following way:
- Check if there already exists stats for that ID
- If yes, add the stats to the former profile

I changed the key of the stats dictonary to the player ID, which should be no problem as it will be redacted later anyway.

There is another bug which occurs in the final stat calculation related to this, but still needs to find it.

### DIFF
--- a/rcon/player_stats.py
+++ b/rcon/player_stats.py
@@ -150,6 +150,20 @@ class BaseStats:
     def _get_player_first_appearance(self, player):
         raise NotImplementedError("_get_player_first_appearance")
 
+    def _add_stats(self, l, stats, p, streaks):
+        actions_processors = {
+            "KILL": self._add_kill,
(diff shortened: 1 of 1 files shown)


LordofAgents:
The checks are currently failing due to payment issues...?
Netherless this PR was only tested locally without docker! I do not have the chance to test it to a real deployment atm.

LordofAgents:
Doesn't work like intented atm. I will convert this to a valid PR when I found a fix.
If you can fix it, you can commit stuff if you guys want to.

Dorfieeee:
Resolved in #1318

PR #958, opened 2025-02-20
Test for presence of scoreboard volume on startup

cemathey:
* Checks for the presence of the scoreboard volume on startup and log/exit if it isn't present. 
  * This makes defining the volume in the `compose.yaml` file mandatory now.
* Handles HTTP 404 errors (message edits) more gracefully and will purge the message queue of remaining messages on failure
* Properly accept steam quick connect URLs for the scoreboard
* Include descriptions in the scoreboard fields
* Remove `show_map_rotation` field which was unused (superseded by a different field anyway)

I tested locally and made sure that the volume is accessible on the host file system and that data is still persisted even when the container is torn down/created again.

### DIFF
--- a/rcon/scoreboard.py
+++ b/rcon/scoreboard.py
@@ -31,6 +31,7 @@ from rcon.webhook_service import (
     WebhookMessageType,
     WebhookType,
     enqueue_message,
+    get_message_edit_404_failure,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


ElGuillermo:
As I understand it, the "scoreboard.db" is created in CRCON's root folder.

Wouldn't it be better to put it in "supervisor_data/" folder (as Scorebot did), so the root stays as clean as possible ?
I think the less there is in root, the easier it is to manage files when doing backups, manually updating, etc.

(?)

cemathey:
> As I understand it, the "scoreboard.db" is created in CRCON's root folder.
> 
> Wouldn't it be better to put it in "supervisor_data/" folder (as Scorebot did), so the root stays as clean as possible ? I think the less there is in root, the easier it is to manage files when doing backups, manually updating, etc.
> 
> (?)

This is using a Docker volume instead of a bind mount, so it isn't actually in the CRCON directory at all but managed separately by Docker.

ElGuillermo (approved):

PR #959, opened 2025-02-21
Update scoreboard.py

ElGuillermo:
- Time comparison between refreshes was the same for all webhooks urls, so only the first one was refreshed, the other being considered as already sent

- the sleep time between refreshes was applied between different urls

### DIFF
--- a/rcon/scoreboard.py
+++ b/rcon/scoreboard.py
@@ -442,7 +442,7 @@ def send_message(
             value=message_id,
         )
     else:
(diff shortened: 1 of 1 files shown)


cemathey (approved):
I made some small tweaks to this so that it wouldn't reset the dict each loop; and used defaultdicts so that the type system is happier and we don't have to `.get()` keys; if it isn't in the dict it just returns `None` indicating we don't know the last time it was updated.

cemathey:
I also built/tested this locally and configured 2 webhooks and verified that they both update.

PR #966, opened 2025-02-23
Fix admin player ID registration bugs

cemathey:
This fixes two separate issues with admin accounts that are associated with player ID records.

Django accounts that are associated with a player ID are used for the in game/online (logged into CRCON) features which can be used in various places such as automatically enabling/disabling player vote kicks in game.

CRCON users can mark Django accounts as either `active` or not; however previously we would register *all* Django accounts with a player ID; regardless of their active status.

This means that people with disabled but not deleted accounts would still be counted in game.

* Only register mods where the account `is_active`

The second issue is that the call to get registered mods was cached for `1 hour` but not invalidated anywhere. The update calls were used anytime a `User` or `SteamPlayer` model instance was modified, but it would take up to `1 hour` for this effect to be noticed.

It's also possible (but unlikely) that people are modifying accounts outside of the Django admin site; such as through the Django Python shell and when that happens the cache isn't invalidated either; so the cache is invalidated on startup.

* Invalidate `get_moderators_accounts` on startup and whenever a record is modified

### DIFF
--- a/rconweb/api/apps.py
+++ b/rconweb/api/apps.py
@@ -1,6 +1,18 @@
 from django.apps import AppConfig
+from rcon.cache_utils import invalidates
 
 
-# TODO: this isn't used anywhere
 class ApiConfig(AppConfig):
     name = "api"
+
+    def ready(self):
+        from rcon.audit import set_registered_mods
+
+        # Can't import from rconweb.api until Django is ready
+        from .auth import get_moderators_accounts
+
+        # Invalidate the cache on start up because you can modify Django
+        # records while CRCON is offline (through the CLI, etc.)
+        with invalidates(get_moderators_accounts):
+            # Register active admin accounts on startup for the ingame/online mods feature
+            set_registered_mods(get_moderators_accounts())
--- a/rconweb/api/auth.py
+++ b/rconweb/api/auth.py
@@ -19,11 +19,12 @@ from django.db.models.signals import post_delete, post_save
 from django.http import HttpResponse, QueryDict, parse_cookie
 from django.views.decorators.csrf import csrf_exempt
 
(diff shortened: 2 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -58,12 +58,17 @@ jobs:
...
         run: docker compose -f docker-compose.dev.yml up -d postgres redis
-      - name: run tests
+      - name: perform database migrations
         run: |
           export HLL_DB_URL=postgresql://${HLL_DB_USER}:${HLL_DB_PASSWORD}@${HLL_DB_HOST}:${HLL_DB_HOST_PORT}/${HLL_DB_NAME}
           export HLL_REDIS_URL=redis://${HLL_REDIS_HOST}:${HLL_REDIS_PORT}/${HLL_REDIS_DB}
           
           PYTHONPATH=. alembic upgrade head
           PYTHONPATH=. python ./manage.py init_db
+          PYTHONPATH=. python ./rconweb/manage.py makemigrations --no-input
```
FlorianSW, line 68: is this actually a good idea? 🤔 I just see we do that in the entrypoint as well, however, I would say that the pipeline and image should only work with the persisted migrations and do not automatically add any new migrations isn't it?
Maybe I misunderstand `makemigrations`, though 🙈

FlorianSW (approved):
Just one side quest comment :)

.github/workflows/pr.yml
cemathey: Yeah actually; you may be correct since the migrations *should* be (and they are) in source control; but it will also auto detect any changes to the user models we use in Django and run those.

I added that while trying to troubleshoot an issue that was only appearing in tests that I could not duplicate in a real build; I'm going to take it out because in the event we haven't committed a change and it matters I'd like to find out while the tests are running.

PR #973, opened 2025-02-25
Hotfix/webhook service loop

cemathey:
tl;dr Treating the scoreboard messages as queue-able messages was a mistake; there is 0 value in older messages unlike things like audit or team kill ones. This converts those style messages to key/value pairs instead of a queue so we only store the most recent one.

Thanks to the fine people over at Independent Contractors (IC) I was able to test this fix on a CRCON install with 6 servers and it seems to be working much better.

* Change the webhook service restart policy from `always` to `unless-stopped` to allow people to turn it off if they desire
* Change the scoreboard messages from being a queue to being simple key/value pairs in Redis so that we only store the most recently queued update
* Change the sleep times to `0`; specifying a sleep time can prevent larger numbers of queues from being able to empty faster than they can be filled
* Limit the webhook service to use `1` CPU core; it is inherently limited by (relatively) slow network processes; this helps mitigate the CPU usage from looping so frequently from the sleep change
* Updates `get_webhook_service_summary` to account for the scoreboard message change and also includes those messages now
* Adds more HTTP error handling; we handle certain HTTP errors specifically (`400`, `401`, `403`, `404`, `429`) but back out early otherwise (for instance `500` series codes when Discord is having issues

There are still other things that we could do in the future to improve this; first and foremost refactoring this into something that handles this style of program better than Python, or using some other better message queue framework that would allow us to `block` until a message is received; start a task for that message and then `block` again.

That would prevent CPU utilization from being wasted simply looping waiting for a message to be enqueued; especially on empty servers.

### DIFF
--- a/rcon/scoreboard.py
+++ b/rcon/scoreboard.py
@@ -30,7 +30,7 @@ from rcon.webhook_service import (
     WebhookMessage,
     WebhookMessageType,
     WebhookType,
-    enqueue_message,
+    enqueue_scoreboard_message,
     get_message_edit_404_failure,
 )
 
@@ -442,7 +442,7 @@ def send_message(
         )
     else:
         logger.info(f"enqueuing {wh.message_id=} {key=}")
-        enqueue_message(
+        enqueue_scoreboard_message(
             message=WebhookMessage(
                 discardable=True,
                 edit=True,
@@ -450,7 +450,8 @@ def send_message(
                 message_type=WebhookMessageType.SCOREBOARD,
                 server_number=SERVER_NUMBER,
                 payload=wh.json,
-            )
+            ),
+            message_key=key,
         )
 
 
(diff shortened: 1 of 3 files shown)


rcon/scoreboard.py
```diff
@@ -442,15 +442,16 @@ def send_message(
         )
     else:
         logger.info(f"enqueuing {wh.message_id=} {key=}")
-        enqueue_message(
+        enqueue_scoreboard_message(
```
FlorianSW, line 445: would be nice to have a more generic function. Maybe something like `enqueue_message_last` or something to communicate that this method will discard older queued messages. And instead of message_key I would use the wording `message_group_key` or something.

message_key sounds like it is a unique id for that specific message, not for the group of messages where older messages are discarded.

rcon/webhook_service.py
```diff
@@ -178,6 +182,18 @@ class QueueParts:
     id: str
 
 
+@dataclass
+class ScoreboardQueueParts:
```
FlorianSW, line 186: Same as above: I would love a more generic approach. The webhook service should not need to know about specific types of messages or consumers of it's api.

FlorianSW (changes requested):

FlorianSW (approved):

PR #975, opened 2025-02-25
Split the docker build github action into frontend/backend

cemathey:
This splits the action that runs when we push a tag to build a release into an action for the frontend and an action for the backend.

There is no reason we need these to build sequentially; this does mean that one build could fail and the other could succeed but that was the case before anyway I think.

In any event; a failed build can be easily restarted through GitHub even without access to the Docker Hub repository.

Builds should now only be as slow as the slowest of the two build processes; so we should have images roughly 2x as fast as before.

### DIFF
--- a/.github/workflows/.github/workflows/docker-images-frontend.yml
+++ b/.github/workflows/.github/workflows/docker-images-frontend.yml
@@ -1,0 +1,39 @@
+name: build-docker-images-frontend
+
+on:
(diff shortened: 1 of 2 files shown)


FlorianSW (changes requested):
Instead of having two separate actions, we could simply put them as separate jobs. Jobs always run in parallel :)

PR #976, opened 2025-02-25
Persist game layout when layout was chosen for map

FlorianSW:


cemathey (approved):
Looks good to me other than the one comment.

Did we test this on a live server by chance?

FlorianSW:
There is one more question I was asking myself: Is persisting the resulting game layout (with all randomly selected caps already selected) enough? Or would it make sense to persist the _requested_ layout as well? 🤔 I personally would tend to only saving the resulting layout. wdyt

FlorianSW:
image
Not the best, but it works :D Tooltip over the map name when game layout is available, otherwise it's simply not doing anything.

rcon/types.py
```diff
@@ -492,6 +493,7 @@ class MapsType(TypedDict):
     server_number: Optional[int]
     map_name: str
     result: Optional[dict[str, int]]
+    game_layout: list[str]
```
cemathey, line 496: Shouldn't this also be a `dict[str, list[str]]` and not a `list[str]`?

rcon/models.py
```diff
@@ -509,6 +509,7 @@ class Maps(Base):
     map_name: Mapped[str] = mapped_column(nullable=False, index=True)
     # A dict with the result of the game mapped as Axis=int, Allied=int
     result: Mapped[dict[str, int]] = mapped_column(nullable=True)
+    game_layout: Mapped[list[str]] = mapped_column(JSON, nullable=False, default=dict)
```
cemathey, line 512: This is a `dict[str, list[str]]` now isn't it?

rcon/workers.py
```diff
@@ -71,7 +71,7 @@ def temporary_welcome_in(message, seconds, restore_after_seconds):
     )
 
 
-def get_or_create_map(sess, start, end, server_number, map_name):
+def get_or_create_map(sess: Session, start: datetime.datetime, end: datetime.datetime, server_number: int, map_name: str, game_layout: list[str]):
```
cemathey, line 74: This is a `dict[str, list[str]]` now isn't it?

rcon/workers.py
```diff
@@ -132,8 +133,9 @@ def _record_stats(map_info: MapInfo):
             sess=sess,
             start=start,
             end=end,
-            server_number=os.getenv("SERVER_NUMBER"),
+            server_number=int(os.getenv("SERVER_NUMBER")),
             map_name=map_info["name"],
+            game_layout=map_info["game_layout"] if "game_layout" in map_info else [],
```
cemathey, line 138: This is a `dict[str, list[str]]` now isn't it?

PR #998, opened 2025-03-21
Maint/golang wh service

cemathey:
I rewrote the Python webhook service in Go so we can be more CPU efficient.

We're now blocking instead of polling for both queued messages and transient messages (which use redis pub/sub).

I've used this locally a decent amount and as far as I can tell works properly; I'm not able to test it on our main server at the moment because we're on the VIP branch and don't want to complicate things.

I'm a novice at Go so by all means tear this bit apart; I kind of just slapped it together and did the minimum amount of work I could do to get it to work properly.

I noticed I still have some debug logging added in which I will strip; I don't think this is ready for release until we get someone more experienced with Go to at least give it a skim.

### DIFF
--- a/webhook_service/Dockerfile-webhook-service
+++ b/webhook_service/Dockerfile-webhook-service
@@ -1,0 +1,22 @@
+# Use official Go image as base
+FROM golang:1.24 AS builder
+WORKDIR /app
+
+# Copy go.mod and go.sum first (for caching dependencies)
+COPY go.mod go.sum ./
+RUN go mod download
+RUN go mod tidy
+COPY . .
+RUN go build -o webhook-service *.go
+
+# Final stage
+FROM golang:1.24
+WORKDIR /app
+
+# Copy the built binary from builder
+COPY --from=builder /app/webhook-service .
+
+COPY entrypoint.sh .
+RUN chmod +x entrypoint.sh
+
+ENTRYPOINT ["./entrypoint.sh"]
--- a/webhook_service/main.go
+++ b/webhook_service/main.go
@@ -1,0 +1,864 @@
+/*
+	The current Python web service is very inefficient because it polls.
+
+	Here's the goal of our new solution; use a master queue for messages.
+
+	Listen to the queue in a blocking fashion for CPU efficiency.
+
+	When we receive a message; examine it and determine if we already know
+	the Discord RL bucket for it; and if so send it to the queue for
+	that bucket.
+
+	If we don't; send it to the shared queue and once we've established the
+	RL bucket after a successful request to Discord; future messages will go
+	to the appropriate queue.
+
+	Each queue will also BLOP for efficiency and process one message at a time;
+	the key/value pairs for transient messages (scoreboard at this point)
+	will be managed by checking a set of key names and starting a worker task
+	for that key name
+
+	Workers for
+*/
+
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"math/rand"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/redis/go-redis/v9"
+)
+
+var maxQueueSize int64
+var maxMsgReattempts int
+var localRateLimit int
+var rateLimitWindowSize time.Duration
+var webhookIDPattern *regexp.Regexp
+var workerLookup map[string]*BucketWorker
+var logger *Logger
+
+type Logger struct {
+	debug   *log.Logger
+	info    *log.Logger
+	warning *log.Logger
+	error   *log.Logger
+	level   string
+}
+
+func (l *Logger) Debug(msg string) {
+	if l.level == "DEBUG" {
+		l.debug.Println(msg)
+	}
+}
+
+func (l *Logger) Info(msg string) {
+	if l.level == "DEBUG" || l.level == "INFO" {
+		l.info.Println(msg)
+	}
+}
+
+func (l *Logger) Warning(msg string) {
+	if l.level == "DEBUG" || l.level == "INFO" || l.level == "WARNING" {
+		l.warning.Println(msg)
+	}
+}
+
+func (l *Logger) Error(msg string) {
+	l.error.Println(msg)
+}
+
+func init() {
+	var err error
+
+	logPath := os.Getenv("LOGGING_PATH")
+	fmt.Print(logPath)
+	if logPath == "" {
+		logPath = "yourlogpathisnotset.log"
+	} else {
+		logPath = filepath.Join(logPath, "webhook_service.log")
+	}
+	logLevel := strings.ToUpper(os.Getenv("LOGGING_LEVEL"))
+	if logLevel == "" {
+		logLevel = "INFO"
+	}
+	file, err := os.OpenFile(logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to open log file: %v", err))
+	}
+	tag := getGitTag()
+
+	// Create logger instances with different prefixes
+	baseFlags := log.LstdFlags | log.Lshortfile
+	logger = &Logger{
+		debug:   log.New(file, "["+tag+"] [DEBUG] ", baseFlags),
+		info:    log.New(file, "["+tag+"] [INFO] ", baseFlags),
+		warning: log.New(file, "["+tag+"] [WARNING] ", baseFlags),
+		error:   log.New(file, "["+tag+"] [ERROR] ", baseFlags),
+		level:   logLevel,
+	}
+
+	maxQueueSize, err = strconv.ParseInt(os.Getenv("HLL_WH_MAX_QUEUE_LENGTH"), 10, 64)
+	if err != nil {
+		maxQueueSize = 150
+	}
+
+	var maxReattempts int
+	maxReattempts, err = strconv.Atoi(os.Getenv("HLL_WH_MAX_RETRIES"))
+	if err != nil {
+		maxMsgReattempts = 5
+	} else {
+		maxMsgReattempts = maxReattempts
+	}
+
+	var maxSends int
+	maxSends, err = strconv.Atoi(os.Getenv("HLL_LOCAL_MAX_SENDS_PER_SEC"))
+	if err != nil {
+		localRateLimit = 45
+	} else {
+		localRateLimit = maxSends
+	}
+
+	var windowSize int
+	windowSize, err = strconv.Atoi(os.Getenv("HLL_WH_SERVICE_RL_TIME_WINDOW"))
+	if err != nil {
+		rateLimitWindowSize = time.Duration(600 * int(time.Second))
+	} else {
+		rateLimitWindowSize = time.Duration(windowSize * int(time.Second))
+	}
+
+	webhookIDPattern = regexp.MustCompile(`webhooks/([0-9]+)/`)
+	workerLookup = make(map[string]*BucketWorker)
+}
+
+const (
+	inputQueue                 = "discord_webhook_queue:input"
+	firstTimeQueue             = "discord_webhook_queue:first_time"
+	bucketQueuePrefix          = "discord_webhook_queue:bucket:"
+	webhookErrors              = "discord_webhook_queue:webhook_id_errors"
+	transientChannel           = "discord_webhook_transient:channel"
+	message_404                = "discord_webhook:message_404"
+	bucketsHash                = "discord_webhook:webhook_buckets"
+	bucketRateLimitCountHash   = "discord_webhook_bucket_rl_count"
+	discordGloballyRateLimited = "discord_webhook:global_rate_limited"
+)
+
+// WebhookType represents the type of webhook service
+type WebhookType string
+
+const (
+	WebhookTypeDiscord WebhookType = "discord"
+)
+
+// WebhookMessageType represents the underlying type of a webhook message
+type WebhookMessageType string
+
+const (
+	WebhookMessageTypeLogLine     WebhookMessageType = "log_line"
+	WebhookMessageTypeLogLineChat WebhookMessageType = "log_line_chat"
+	WebhookMessageTypeLogLineKill WebhookMessageType = "log_line_kill"
+	WebhookMessageTypeTeamkill    WebhookMessageType = "log_line_teamkill"
+	WebhookMessageTypeAdminPing   WebhookMessageType = "admin_ping"
+	WebhookMessageTypeScoreboard  WebhookMessageType = "scoreboard"
+	WebhookMessageTypeAudit       WebhookMessageType = "audit"
+	WebhookMessageTypeOther       WebhookMessageType = "other"
+)
+
+// DiscordWebhookDict represents the payload for a Discord webhook
+type DiscordWebhookDict struct {
+	URL             string               `json:"url"`
+	WebhookID       string               `json:"webhook_id"`
+	MessageID       *string              `json:"message_id,omitempty"`
+	AllowedMentions *AllowedMentionsDict `json:"allowed_mentions,omitempty"`
+	Content         *string              `json:"content,omitempty"`
+	Embeds          []DiscordEmbedDict   `json:"embeds,omitempty"`
+}
+
+// Override unmarshaling so we can ignore extra fields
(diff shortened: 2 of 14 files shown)


webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	bucketsHash                = "discord_webhook:webhook_buckets"
+	bucketRateLimitCountHash   = "discord_webhook_bucket_rl_count"
+	discordGloballyRateLimited = "discord_webhook:global_rate_limited"
+)
+
+// WebhookType represents the type of webhook service
+type WebhookType string
+
+const (
+	WebhookTypeDiscord WebhookType = "discord"
```
FlorianSW, line 172: join this const declaration to the block above for readability reasons.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	URL             string               `json:"url"`
+	WebhookID       string               `json:"webhook_id"`
+	MessageID       *string              `json:"message_id,omitempty"`
+	AllowedMentions *AllowedMentionsDict `json:"allowed_mentions,omitempty"`
+	Content         *string              `json:"content,omitempty"`
+	Embeds          []DiscordEmbedDict   `json:"embeds,omitempty"`
+}
+
+// Override unmarshaling so we can ignore extra fields
+func (d *DiscordWebhookDict) UnmarshalJSON(data []byte) error {
```
FlorianSW, line 200: tbh: I don't get this :D Why do you need to do this? 🤔

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	return &BucketWorker{
+		QueueKey:  bucketQueuePrefix + bucketID,
+		rdb:       rdb,
+		ctx:       context.Background(),
+		rateLimit: &RateLimitState{BucketID: bucketID},
+	}
+}
+
+func GetWorker(bucket string, mu *sync.Mutex) (*BucketWorker, bool) {
+	mu.Lock()
```
FlorianSW, line 351: what race condition are you trying to protect against here? I think you can remove this lock for the Get method, couldn't you? In the worst case, it will return false for exists.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	defer mu.Unlock()
+	worker, exists := workerLookup[bucket]
+	return worker, exists
+}
+
+func SetWorker(bucket string, worker *BucketWorker, mu *sync.Mutex) {
```
FlorianSW, line 357: If you make `workerLookup` a struct, where the lookup map is a field and the mutex as well, then callers do not need to pass the mutex in.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	mu.Lock()
+	defer mu.Unlock()
+	workerLookup[bucket] = worker
+}
+
+// Keeps us under `HLL_LOCAL_MAX_SENDS_PER_SEC` across all requests
+func CheckLocalRateLimit(requests int, now time.Time) error {
```
FlorianSW, line 364: Why isn't this then a method on the `localRateLimitState` struct? It shouldn't use the global state that havily, especially as all the operations are done on fields of the `localRateLimitState` struct.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+
+	// If it's been more than 1 second since our last request; reset the window
+	if now.Sub(localRateLimitState.Window) >= time.Second {
+		localRateLimitState.Requests = requests
+		localRateLimitState.Window = now
+	}
+
+	// Don't exceed our configured max requests per second
+	if localRateLimitState.Requests >= localRateLimit {
+		localRateLimitState.mu.Unlock()
```
FlorianSW, line 376: no need to call `Unlock` explicitly, you already deferred that call.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+		return fmt.Errorf("local rate limit exceeded")
+	}
+
+	localRateLimitState.Requests++
+	return nil
+}
+
+func GetWorkerRateLimitSleepTime(rateLimit *RateLimitState) time.Duration {
+	rateLimit.mu.Lock()
```
FlorianSW, line 385: no need for a lock here, either, isn't it? This is a best-effort call and it should be acceptable that it might be incorrect in high-traffic scenarios. Especially, as the value can change _after_ the `GetWorkerRateLimitSleepTime` method returns. The caller can therefore not rely on this value anyway.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	if localRateLimitState.Requests >= localRateLimit {
+		localRateLimitState.mu.Unlock()
+		return fmt.Errorf("local rate limit exceeded")
+	}
+
+	localRateLimitState.Requests++
+	return nil
+}
+
+func GetWorkerRateLimitSleepTime(rateLimit *RateLimitState) time.Duration {
```
FlorianSW, line 384: This method should probably also be on the `RateLimitState` struct, as it is directly tied to it anyway.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+		return "", err
+	}
+
+	// Extract what we need to make our POST request to Discord
+	payload := map[string]any{}
+	if msg.Payload.Content != nil {
+		payload["content"] = *msg.Payload.Content
+	}
+	if len(msg.Payload.Embeds) > 0 {
+		payload["embeds"] = msg.Payload.Embeds
```
FlorianSW, line 408: this assignment can be done without the if condition. An empty slice (or nil for that matter) would mean the same as not having `embeds` added in the first place, iirc.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	}
+	if msg.Payload.AllowedMentions != nil {
+		payload["allowed_mentions"] = msg.Payload.AllowedMentions
+	}
+
+	var jsonPayload []byte
```
FlorianSW, line 414: use `jsonPayload, err := json.Marshal(payload)` then there is no need to declare the vars here.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+		return bucket, nil
+	} else {
+		return "", fmt.Errorf("unhandled HTTP status %d", resp.StatusCode)
+	}
+}
+
+// ProcessQueue handles a bucket queue with retries
+func (bw *BucketWorker) ProcessQueue(rdb *redis.Client) {
+	for {
+		bw.rateLimit.mu.Lock()
```
FlorianSW, line 514: I would always make a defer call to Unlock, just to make sure that all branches unlock the mutex. In this case, I think you covered this with explicit calls, but I would feel much safer still having this here :D

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	req, err := http.NewRequest(method, url, bytes.NewBuffer(payload))
+	if err != nil {
+		return nil, fmt.Errorf("request error: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("send error: %v", err)
+	}
+	defer resp.Body.Close()
```
FlorianSW, line 594: are we sure to never want to read the body? 🤔 I would've expected this call to be done at the calling side of `ExecuteWebhook`, but ok :)

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+
+	return resp, nil
+}
+
+// ProcessFirstTime discovers buckets for unknown webhooks
+func ProcessFirstTime(rdb *redis.Client, mu *sync.Mutex) {
+	ctx := context.Background()
+	for {
+		results, err := rdb.BLPop(ctx, 0, firstTimeQueue).Result()
```
FlorianSW, line 603: `BLPop` with 0 as a duration argument, it should be the same as `LPop`, isn't it? 🤔 I would then use `LPop` instead :)

Wait: There is no break condition in this for loop and no sleep or any other blocking call, isn't it? Doesn't this result in an indefinite loop that will be run as fast as the CPU allows? Doesn't sound very ressource efficient :D Didn't you want to use any other duration in `BLPop` than 0?

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+}
+
+func ProcessTransient(rdb *redis.Client, msg Message, mu *sync.Mutex) {
+	webhookID, err := ExtractWebhookID(msg.Payload.URL)
+	if err != nil {
+		logger.Error(fmt.Sprintf("could not parse webhook ID from: %s:%s", msg.Payload.URL, err))
+		return
+	}
+
+	ctx := context.Background()
```
FlorianSW, line 680: I see a lot of `context.Background` calls flying around here. The better approach would be to have a single context passed around for the calls to be made. Meaning: In the entrypoint method, use `context.Background` (or any other way to construct a context for that matter) and pass that as an argument to the methods that require a context. Usually this is the first argument to a method.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+}
+
+func main() {
+	logger.Info("Starting service")
+	rdb := SetupRedis()
+
+	// For the Docker health check
+	path := filepath.Join("/app", "webhook-service-healthy")
+	logger.Info(path)
+	exec.Command("touch", path).Start()
```
FlorianSW, line 724: shelling out to create a file... ooof :) What about using `os.Create` instead? :)

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+				}
+			}
+		}
+	}()
+
+	go ProcessFirstTime(rdb, &mu)
+	go SubscribeTransients(rdb, &mu)
+
+	// Keep running
+	select {}
```
FlorianSW, line 791: whenever there is a SIGTERM or SIGINT, the program should gracefully shutdown.Closing all redis connections, cleaning up any other stuff if necessary.

E.g. like here:
https://github.com/FlorianSW/hll-discord-server-watcher/blob/main/cmd/cmd.go#L94-L101
(just with closing the redis client and stuff)

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	matches := webhookIDPattern.FindStringSubmatch(url)
+	if len(matches) < 2 {
+		return "", fmt.Errorf("no webhook ID found in URL")
+	}
+	return matches[1], nil
+}
+
+func SetGloballyRateLimited(rdb *redis.Client, resetTime time.Time, resetAfter time.Duration) {
+	ctx := context.Background()
+	val := strconv.FormatInt(resetTime.Unix(), 10)
```
FlorianSW, line 825: `strconv.Itoa` is probably easier to understand then FormatInt in this simple case :)

FlorianSW: It could then also be inlined.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	rdb.Set(ctx, discordGloballyRateLimited, val, resetAfter)
+}
+
+// Add a hash entry anytime a bucket is rate limited so we can count
+// the number of rate limits within the user configured window size
+// and return it from the API
+func UpdateBucketRateLimitCount(rdb *redis.Client, bucket string) {
+	ctx := context.Background()
+	hashName := fmt.Sprintf("%s:%s", bucketRateLimitCountHash, bucket)
+	key := strconv.FormatInt(time.Now().Unix(), 10)
```
FlorianSW, line 835: strconv.Itoa

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+		"http_401": false,
+		"http_403": false,
+		"http_404": false,
+	}
+	rdb.HSet(ctx, key, payload)
+}
+
+func getGitTag() string {
+	// Run git describe --tags
+	cmd := exec.Command("git", "describe", "--tags")
```
FlorianSW, line 866: tbh, I don't think that the whole git repo should be in the final image. Instead, what about using `git describe --tags` in the build stage of the container and set the tag as an environment variable backed into the image? Then this would be a simple `os.Getenv` call.

FlorianSW (changes requested):
Sorry for the number of comments, but you said you are open for suggestions :P

I wrote this some times in the comments, but want to stress this again:
You could make this much more organised by splitting logic into their appropriate structs, create methods to construct them and use that instead.

I think also that splitting the structs and logical blocks into separate files (while being in the same package) would help work wtih the code tremedously as well.

webhook_service/Dockerfile-webhook-service
```diff
@@ -0,0 +1,22 @@
...
+
+# Copy go.mod and go.sum first (for caching dependencies)
+COPY go.mod go.sum ./
+RUN go mod download
+RUN go mod tidy
+COPY . .
+RUN go build -o webhook-service *.go
+
+# Final stage
+FROM golang:1.24
```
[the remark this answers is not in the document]
kamilkamilc: Make sure you also copy certs to scratch image `COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/`

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+
+var maxQueueSize int64
+var maxMsgReattempts int
+var localRateLimit int
+var rateLimitWindowSize time.Duration
+var webhookIDPattern *regexp.Regexp
+var workerLookup map[string]*BucketWorker
+var logger *Logger
+
+type Logger struct {
```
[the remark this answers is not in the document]
cemathey: Ah thanks; I was trying to wrap this up so I committed the crime of being so lazy it wrapped around to being more work than just looking up the right package.

webhook_service/Dockerfile-webhook-service
```diff
@@ -0,0 +1,22 @@
+# Use official Go image as base
+FROM golang:1.24 AS builder
+WORKDIR /app
+
+# Copy go.mod and go.sum first (for caching dependencies)
+COPY go.mod go.sum ./
+RUN go mod download
```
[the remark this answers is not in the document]
cemathey: I'll defer to you on that one.

webhook_service/main.go
```diff
@@ -0,0 +1,873 @@
...
+	baseFlags := log.LstdFlags | log.Lshortfile
+	logger = &Logger{
+		debug:   log.New(file, "["+tag+"] [DEBUG] ", baseFlags),
+		info:    log.New(file, "["+tag+"] [INFO] ", baseFlags),
+		warning: log.New(file, "["+tag+"] [WARNING] ", baseFlags),
+		error:   log.New(file, "["+tag+"] [ERROR] ", baseFlags),
+		level:   logLevel,
+	}
+
+	maxQueueSize, err = strconv.ParseInt(os.Getenv("HLL_WH_MAX_QUEUE_LENGTH"), 10, 64)
```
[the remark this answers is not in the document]
cemathey: No can do here; `rdb.LTrim` specifically requires an `int64`.

webhook_service/main.go
cemathey: > tbh: I don't get this :D Why do you need to do this? 🤔

I guess we don't really have to but there's a lot of extra shit that gets serialized into it and we only care about a few very specific things; I didn't want to include all the extra stuff so I just ignored those fields.

webhook_service/main.go
cemathey: After some reflection (and I think you mentioned it in a different comment?) I don't think we even need to rate limit here.

Each queue operates on a single bucket and shouldn't ever be competing for that struct.

webhook_service/Dockerfile-webhook-service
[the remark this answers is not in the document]
cemathey: I opted for alpine because it's always nice to have a shell on the Docker container you can exec into.

cemathey:
> Sorry for the number of comments, but you said you are open for suggestions :P
> 
> I wrote this some times in the comments, but want to stress this again: You could make this much more organised by splitting logic into their appropriate structs, create methods to construct them and use that instead.
> 
> I think also that splitting the structs and logical blocks into separate files (while being in the same package) would help work wtih the code tremedously as well.

Very much appreciated.

I made substantial changes to this if you would like to take another look.

FlorianSW:
@cemathey I looked at the PR, thanks for the updates :) I was wondering if you want to take a look at some (a lot actually) changes I made to the structure of the code and some minor details of how the different components work together. Do you want to take a look at them? I made a separate branch to not clutter them into this PR 😅 
https://github.com/MarechJ/hll_rcon_tool/compare/maint/golang_wh_service...maint/golang_wh_service_refactor?expand=1

What do you think? Do you get what I try to do with these changes? :)

PR #1007, opened 2025-04-02
Fix   tobruk images

ElGuillermo:
First time I do this. Formats and sizes are the same as others... I may miss something, though, so don't accept this PR blindly ;)


cemathey:
Did a test build; working as expected.

PR #1023, opened 2025-05-02
UI: Map settings

Dorfieeee:
Better views for map setting pages

image

### DIFF
--- a/rcongui/src/components/Header/nav-data.js
+++ b/rcongui/src/components/Header/nav-data.js
@@ -76,31 +76,40 @@ export const navMenus = [
     ],
   },
   {
-    name: "Maps",
-    icon: <MapIcon />,
     links: [
       {
-        name: "Change",
-        to: "/settings/maps/change",
-        icon: <RestartAltIcon />,
-      },
-      {
-        name: "Rotation",
-        to: "/settings/maps/rotation",
-        icon: <ThreeSixtyIcon />,
-      },
-      {
-        name: "Objectives",
-        to: "/settings/maps/objectives",
-        icon: <AdsClickIcon />,
-      },
-      {
-        name: "Votemap",
-        to: "/settings/maps/votemap",
-        icon: <HowToVoteIcon />,
+        name: "Maps",
+        to: "/settings/maps/list",
+        icon: <MapIcon />,
       },
     ],
   },
+  // {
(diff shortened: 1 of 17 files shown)


caranci:
Quick comments, without any consideration about effort or feasibility. :)
1. I'm not sure what the **copy** button does.
2. I think a **refresh** button is needed. A common flow is to mess with the rotation from yesterday, and then decide you want to start over.
3. To be honest, I'm not sure I like the tabs. I would rather go straight to the rotation settings. And having the **List** tab (i.e., change current map) first seems dangerous somehow...
4. It might be useful to have checkboxes on the weather combo-box to allow multiple-selection.
image
5. It would be useful if the available maps were effectively in a table-view with different columns for **name**, **game mode**, and **weather**, allowing people to sort by the columns to more effectively find the maps they want.
image
6. It would be convenient to allow drag-and-drop from the available maps list to the map rotation list.

caranci:
Fair enough!

caranci:
### Maps>Change page
1. Server live… didn't test any changing the map. Is there an "are you sure" pop-up? Might be a good idea if there isn't.

### Maps>Rotation>Builder page
1. Calling the refresh button "undo" is a bit misleading. Undo implies the last change will be reverted, but the button really refreshes the list from what is currently running on the server.
2. I like the "table-like" layout of the tiles in the map lists.
3. I like the check-boxes in the combo-boxes.
4. I didn't test the save button… server is seeded…

### Maps>Objectives>Map Objectives

1. Looks intuitive and seems to work.
2. Server live… didn't test "confirm" button. Is there an "are you sure" pop-up? Might be a good idea if there isn't.

### Maps>Votemap>Whitelist>Status

1. We aren't using it at the moment, so I don't know if the status works.
2. Looks easy to understand.

### Maps>Votemap>Whitelist>Map Whitelist

1. Search / filtering for the list of maps currently whitelisted would be useful.
2. When all the maps are in the whitelist the "No maps found -- Try adjusting your filters or search term" note is a bit confusing. There just aren't any maps at all because they are all in the whitelist. :) There is also a moment where the page renders all the maps in the available list before it filters the ones that are whitelisted… and this slight delay makes it seem like it was broken.

### Cosmetic
1. Why don't some of the list containers extend to the bottom of the page? 
image
image

PR #1024, opened 2025-05-03
Add Deadzonefences to Allowlist in Seeding Logic

2KU77B0N3S:
As allowlisted areas, players in these zones should not receive out-of-bounds warnings, even if they are temporarily outside.
Players driving trucks may cross deadzones as part of valid routes. Without allowlisting, they’d be unfairly warned or punished.

The original code applies player count and game mode conditions only to axisFences, potentially causing inconsistent enforcement for Allied players. Applying conditions to both fence sets ensures uniform behavior.

### DIFF
--- a/seeding/automod.go
+++ b/seeding/automod.go
@@ -134,85 +134,8 @@ func initialize(ctx context.Context, logger *slog.Logger, ac *crcon.AutoModSeedi
 				},
 			},
 		},
-		// Allowlist deadzonefences: Horizontal A2 to J2
-		{
-			X: &horizontalCaps[0],
-			Y: internal.Pointer(2),
-			Condition: &data.Condition{
-				Equals: map[string][]string{
-					"map_name": axisToAlliedHorizontalMaps,
-				},
-			},
-		},
-		{
-			X: &horizontalCaps[9],
-			Y: internal.Pointer(2),
-			Condition: &data.Condition{
-				Equals: map[string][]string{
-					"map_name": alliedToAxisHorizontalMaps,
-				},
-			},
-		},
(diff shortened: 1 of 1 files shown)


2KU77B0N3S:
Line 5-7 was moved always automatically from VSC

seeding/automod.go
```diff
@@ -132,8 +134,85 @@ func initialize(ctx context.Context, logger *slog.Logger, ac *crcon.AutoModSeedi
 				},
 			},
 		},
+		// Allowlist deadzonefences: Horizontal A2 to J2
```
FlorianSW, line 137: This looks more like A2 _and_ J9, instead of _to_.

Meaning, you probably wanted to have a fence more like this:
```
		{
			Y: internal.Pointer(2),
			Condition: &data.Condition{
				Equals: map[string][]string{
					"map_name": append([]string{}, axisToAlliedHorizontalMaps..., alliedToAxisHorizontalMaps...),
				},
			},
		},
```

And then remove the second fence added (the X: J, Y: 2, one). Same for the vertical maps then as well 🤔

seeding/automod.go
```diff
@@ -166,6 +245,82 @@ func initialize(ctx context.Context, logger *slog.Logger, ac *crcon.AutoModSeedi
...
```
FlorianSW, line 248: same as for axis fence.

FlorianSW (changes requested):
it's late and I had a long day, so take my code suggestions (the whole review as a matter of fact) with a grain of salt. I might've misunderstood what you did 😅

PR #1037, opened 2025-06-15
Fix u.current.values().forEach is not a function

Fragger:
Error appeared in Firefox 128 ESR and possibly other browsers
Screenshot_2025-06-14_21-56-51


FlorianSW:
Thanks for the catch and for fixing that. Sadly we seem to did not check for the compatibility of `forEach` on an Iterator 🙈

FlorianSW (approved):

PR #1041, opened 2025-06-16
Fix and update create/edit blacklist record

### LINKED ISSUE #887: Blacklist Bugs - Never Expire, Time Selection and Dev Console Errors.
Version: API Version: v11.2.0

A couple of issues I've noticed with blacklists feature

1. Blank for never expires - When you edit an expired blacklist item, you can't set the date to "blank". You have to change the blacklisted item expire date to at least a day in the future, and save the item; then, you can go in and remove the date and save the item. This is problematic, there should just be a checkbox that blanks out the date.
2. When editing a blacklist item, there is a dev console error 
```The specified value "2025-01-16T05:00:00+00:00" does not conform to the required format.  The format is "yyyy-MM-ddThh:mm" followed by optional ":ss" or ":ss.SSS". index-DDXinD0Z.js:38```
3. The + and - signs on all the options for hour, day, month don't work

Fragger:
Should fix #887 and act more like the add player to blacklist player action

Before:
Screenshot_2025-06-15_23-24-58

After:
Screenshot_2025-06-15_23-24-19

### DIFF
--- a/rcongui/src/components/Blacklist/BlacklistRecordCreateDialog.js
+++ b/rcongui/src/components/Blacklist/BlacklistRecordCreateDialog.js
@@ -5,16 +5,20 @@ import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
 import DialogContentText from "@mui/material/DialogContentText";
 import DialogTitle from "@mui/material/DialogTitle";
-import { FormControl, InputLabel, MenuItem, Select, Typography } from "@mui/material";
+import { Alert, FormControl, InputLabel, MenuItem, Select, Typography } from "@mui/material";
 import moment from "moment";
 import { getServerStatus, getSharedMessages } from "@/utils/fetchUtils";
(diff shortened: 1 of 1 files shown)


Dorfieeee:
There is an issue when on form save. The whole page needs a refresh. I am not planning to merge this PR.

image

Fragger:
That should fix that issue on form save now, expiresAt was getting set to "" when the dialog was closed.  Tested a few creates with and without it.

Fragger:
Any chance this could make it in even if the whole blacklist page is eventually going to get rewritten?  This at least fixes editing records to change durations like when a record gets expired by Barricade from being put on the wrong blacklist.

FlorianSW:
> Any chance this could make it in even if the whole blacklist page is eventually going to get rewritten? This at least fixes editing records to change durations like when a record gets expired by Barricade from being put on the wrong blacklist.

@Dorfieeee any opinion?

Dorfieeee (approved):
LGTM

PR #1053, opened 2025-06-27
Light mode improvements.

caranci:
- Add converted icons with black foreground and transparent background.
- Use new icons when light mode selected.
- Use default background colour for light mode (instead of grey / almost black).
- Use better tier level text colours for lightmode.

I haven't merged this with our production instance, but I have tested it and it seems good.

Sample screenshots below.

Screenshot 2025-06-27 182512
Screenshot 2025-06-27 182746
Screenshot 2025-06-27 182942


rcongui/src/components/game/useTheme.js
```diff
@@ -0,0 +1,3 @@
+import { useTheme } from "@mui/material/styles";
+
+export default useTheme;
```
FlorianSW, line 3: any reason why this separate export is needed? Why not importing `@mui/material/styles` directly as in the other files?

FlorianSW (commented):
Just one question. And sorry for the late review :(

Dorfieeee (approved):
LGTM now

rcongui/src/components/game/useTheme.js
Dorfieeee: Frankly there are 3 ways to import `useTheme` so... I had AI explain the subtle differences to me :D

caranci: So if the code is okay then I should click the button to merge?

PR #1057, opened 2025-07-11
remove author and linkdn info, change theme-color

Dorfieeee:


Dorfieeee:
The `theme-color` is dynamic based on the theme it does not need to be in index.html as it creates duplicate meta tags.

rcongui_public/src/hooks/use-theme-provider.tsx
```diff
@@ -1,3 +1,5 @@
+"use client"
```
FlorianSW, line 1: is that really needed? We do not have server-side rendering anywhere, aren't we? 🤔

FlorianSW (changes requested):

rcongui_public/src/hooks/use-theme-provider.tsx
Dorfieeee: No it is not. Brain seizure

FlorianSW: Once removed, I merge 🥳

FlorianSW (approved):

PR #1063, opened 2025-07-16
Update Dockerfile-frontend

ElGuillermo:
Debian Buster repo has been decommited, apt commands can't reach it anymore.

This isn't a prob in "vanilla" CRCONs, as this dockerfile is only used when building new Docker images.

I successfully built Docker images using the new "bookworm" Debian release.


caranci:
I noticed the same when setting up a new development environment. Image caching allowed existing environments to still work.

caranci (approved):

PR #1064, opened 2025-07-18
Feat: Votemap rework, multiple votes per player, player can add map to selection

Dorfieeee:
## UI Changes
- After successful vote, player can now see map's pretty name
- Show selection `!vm|!votemap` displays only list of maps and which map the player voted
- Map reminder contains the same as "show selection" and additional instruction text as of now
- Map selection index starts from 1. 0 is reserved for the new feature.

## TODO
- [x] - API permission migration

## New features
- [x] - Exclude maps of the same game mode with different environments
- [x] - Send reminder to vote manually
- [x] - User can choose whether to send the reminder on match start, end or both 
- [x] - Add new map to selection
- [x] - Remove map from selection 
- [x] - Guarantee the next map and thus pausing the voting for that round
- [x] - Allow voting for VIP players only 
- [x] - Allow voting based on white-list or ban-list flags
- [x] - Add vote - this feature is mainly meant for 3rd party apps - the vote can count as given vote_count or based on flags or vip if valid player_id and the player is in the database
- [x] - Players can add one more map to the selection. This map will be on the top of the list at index 0 and it will say "<_player_>'s pick: <_map_name_>". Only players with a specified flag will be able to use this feature. If no flags are set, all players can use this command. Once a player picked a map, it cannot be picked again until votemap restart. If there is any reward system ever introduced, the players might be able to rewrite the pick for reward points doubling with each change.
  - New command `!vm add help`
  - New command `!vm add <map_tag> [game_mode] [attackers | only if game_mode=offensive] [environment]` where game_mode, environment and attackers respectively default to warfare, day and allies
- [x] - Player's vote can have different vote count based on his flags. The flag with the lowest count will be used. Defaults to 1 count as of now. What does that mean? Players with flags and their respective vote counts:
  - set to `0` will practically disable player's ability to vote and the player will be informed in such a case with message "Player banned from voting"
  - set to `2+` will give players more decisions power
  - set to max value `100` may be given to admin users to outvote others.
  - 
image

## API Changes
- GET `get_votemap_status` returns the following object and is in the same order as when seen in the game message.
```
{
  enabled: bool
  paused: bool
  winning_map: str
  last_reminder: datetime
  results: { 
    map: MapLayer,
    votes_count: int,
    voters: { 
      player_name: str,
      player_id: str,
      vote_count: int 
    }[]
  }[]
}
```
- GET `get_votemap_config` and `set_votemap_config` contains new field
```
vote_flags = {
  flag: string
  vote_count: int
}[]
```
- GET `get_votemap_results` returns up to 100 historical votemap results
```
result = {
  ts: int
  map: MapLayer
  results: {
    map: MapLayer,
    votes_count: int
  }
}[]
```
- POST `add_map_to_votemap`, `remove_map_from_votemap` and `set_votemap_winner`
```
payload = {
  map_name: string
}
```
- POST `add_votemap_vote`
```
payload = {
  map_name: string
  player_id: string
  player_name: string
  vote_count: int | null
}
```
-  POST `send_votemap_reminder` forces reminder to vote that is displayed to all players except the ones who have voted and/or have reminders disabled

## Internal changes
 - [x] - Write new tests
- unified all redis keys names used by votemap as `votemap:<key>`
- VoteMap's state and command handling moved to separate classes
- New getters and setters for others to easily implement some custom features on their forks


rcon/vote_map.py
```diff
@@ -572,253 +571,329 @@ def register_vote(self, player_name: str, vote_timestamp: int, vote_content: str
...
 
             # take advantage of python dicts being ordered so the winner is always the first item
-            return {m: v for m, v in maps}
+            return {
+                maps.parse_layer(map_id): vote_count
+                for map_id, vote_count in ranked_maps
+            }
 
         except Exception:
             logger.exception("Can't produce vote overview")
```
timraay, line 633: This should either return an empty dict or raise.

timraay (commented):
For native Python types (`list`, `dict`, ...) you do not have to use their `typing` equivalent anymore for type annotations. `foo: typing.List[str]` therefore becomes `foo: list[str]`.

rcon/user_config/vote_map.py
```diff
@@ -76,18 +105,27 @@ class VoteMapUserConfig(BaseUserConfig):
     allow_consecutive_skirmishes: bool = Field(default=False)
     allow_default_to_skirmish: bool = Field(default=False)
     instruction_text: str = Field(default=INSTRUCTION_TEXT)
-    thank_you_text: str | None = Field(default=THANK_YOU_TEXT)
+    thank_you_text: str = Field(default=THANK_YOU_TEXT)
     no_vote_text: str = Field(default=NO_VOTE_TEXT)
     reminder_frequency_minutes: int = Field(ge=0, default=20)
     allow_opt_out: bool = Field(default=True)
-    help_text: str | None = Field(default="")
+    help_text: str = Field(default=HELP_TEXT)
```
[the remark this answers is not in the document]
Dorfieeee: It will return the default.

rcon/user_config/vote_map.py
```diff
@@ -70,24 +105,39 @@ class VoteMapUserConfig(BaseUserConfig):
...
+    allow_vip_only: bool = Field(default=False, description="Allow votemap for VIP players only.")
+    allow_flag_only: list[str] = Field(default_factory=list, title="Allow votemap for FLAGGED players only.", description="Players having one of these flags are allowed to run votemap commands.")
 
     @staticmethod
     def save_to_db(values: VoteMapType, dry_run=False):
         key_check(
             VoteMapType.__required_keys__, VoteMapType.__optional_keys__, values.keys()
         )
 
+        # model_fields = VoteMapUserConfig.model_fields.keys()
```
FlorianSW: this can be removed, can't it?

rcon/utils.py
```diff
@@ -4,7 +4,7 @@
 import secrets
 from datetime import datetime, timedelta, timezone
 from itertools import islice
-from typing import Any, Generic, Iterable, TypeVar
+from typing import Any, Generic, Iterable, TypeVar, cast
```
FlorianSW: looks unused, isn't it?

rcon/utils.py
```diff
@@ -285,7 +285,7 @@ class MapsHistory(FixedLenList[MapInfo]):
     def __init__(self, key="maps_history", max_len=500):
         super().__init__(key, max_len)
 
-    def save_map_end(self, old_map=None, end_timestamp: int = None):
+    def save_map_end(self, old_map: str, end_timestamp: int | None = None) -> MapInfo:
```
FlorianSW: this changes the behavior of this method (old_map is not None by default anymore?). Did you check if that has any implications on calling sides?

FlorianSW (commented):
was able to go through half of the code, only. Found nothing except some minor code-things.

rcon/utils.py
Dorfieeee: `save_map_end` is never called without providing the `old_map` argument.
I've been running this update for nearly a year without any issues.

rcon/user_config/vote_map.py
Dorfieeee: Actually I think the commented out code should be used instead of the manual instantiation.
```
        model_fields = set(VoteMapUserConfig.model_fields.keys())
        filtered_values = {k: v for k, v in values.items() if k in model_fields}
        
        validated_conf = VoteMapUserConfig.model_validate(filtered_values)
```

timraay (approved):

PR #1069, opened 2025-07-26
Feat: Go to map in the rotation

Dorfieeee:
Implementation of ?forgotten `map_ordinal` parameter in [Map command](https://gist.github.com/DeteCT0R/7d2837729d98eea8f82888e0c48c323c#map-map_name-map_ordinal).

```python
    def set_map(self, map_name: str, map_ordinal: int = 1) -> str:
        return self._str_request(f"map {map_name} {map_ordinal}", log_info=True)
```

Users are now able to go to a specific map in the rotation via the UI.
> e.g. First Driel is selected, the next map will be El Alamein, and Foy when the second Driel is selected.

image

### DIFF
--- a/rcon/commands.py
+++ b/rcon/commands.py
@@ -544,8 +544,8 @@ class ServerCtl:
     def set_welcome_message(self, message) -> str:
         return self._str_request(f"say {message}", log_info=True, can_fail=False)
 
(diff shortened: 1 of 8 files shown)


Dorfieeee:
This is not supported as of version U18.

PR #1075, opened 2025-07-26
Fix: User cannot download VIP files

Dorfieeee:
There was very stupid logic handling text/plain responses and if the response was of this type it always threw error.
While I was at it I managed to simplify the main request handling function.


Dorfieeee:
DO NOT MERGE

PR #1081, opened 2025-08-10
Display system usage information

Dorfieeee:
Added cpu, ram and disk usage information.
A minor change in styling the bottom section about connection and version details.

I believe this is useful information to have and save some logging into the environment directly.

🆕 API endpoint `/api/get_system_usage`
> example
```
{
    "cpu_usage": {
        "cores": 4,
        "percent": 10.4,
        "process_count": 4
    },
    "ram_usage": {
        "total": 15.244922637939453,
        "used": 7.427345275878906,
        "percent": 51.5
    },
    "disk_usage": {
        "total": 149.91522216796875,
        "used": 26.94037628173828,
        "percent": 18.7
    }
}
```

image

### DIFF
--- a/rcongui/src/components/layout/SideMenu.js
+++ b/rcongui/src/components/layout/SideMenu.js
@@ -57,19 +57,10 @@ export default function SideMenu() {
       <ServerStatus />
       <Divider />
       <MenuContent navigationTree={navMenus} />
-      <List dense>
-        <ListItem sx={{ height: 20, "& .MuiListItemText-root .MuiListItemText-primary": { fontSize: "0.75rem" } }}>
-          <ListItemText
-            sx={{ marginLeft: -0.5 }}
-            primary={<ConnectionStatus />}
-          />
-        </ListItem>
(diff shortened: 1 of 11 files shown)


FlorianSW:
Just some thoughts I have when reading the changes in the backend:
rconweb/api runs in the `backend` container (from `docker-compose.yml`), isn't it? What would `psutils` then return as CPU usage, disk space and memory?

For memory I would _assume_ the used memory for that container and as max memory either the system memory, or, when a limit is set in the `cgroups`, that? Or would it in that case also show the installed system memory? (Meaning, what is this percentage based on)

What about CPU: The CPU usage is the same, is it based on the processed from inside the container? I would assume so, as this is the whole idea of a container. Meaning, how much sense does this metric make?

For disk space: This is probably really interesting as well, what is the "disk" in this case? 🤔 

P.S.: Don't get me wrong here, I like the idea and it would probably make some debugging with people easier, who do not know how to get that information via their terminal (yet alone, know what a terminal is). These are merely questions that came up when I read the PR. From the code side itself, this PR would get an Approve from me :)

FlorianSW (approved):

PR #1103, opened 2025-10-05
Add support for U18 (RCONv2)

FlorianSW:
Just opened a PR for this so we can track the progres a bit easier, review changes as they come in and comment on them much easier.


Dorfieeee:
Impeccable

PR #1112, opened 2025-10-07
Map icons/misxsing stalingrad

FlorianSW:


Fragger:
These are also needed in rcongui_public for the public stats page

PR #1113, opened 2025-10-08
Fix get_perma_bans/get_temp_bans

Fragger:
dict was still being treated as a str by `_struct_ban` and resulting in `Invalid perma ban line:` log lines

### DIFF
--- a/rcon/rcon.py
+++ b/rcon/rcon.py
@@ -494,13 +494,19 @@ class Rcon(ServerCtl):
 
     @ttl_cache(ttl=60)
     def get_perma_bans(self) -> list[GameServerBanType]:
-        return [self._struct_ban(ban=x, type_=PERMA_BAN) for x in super().get_perma_bans()]
+        return [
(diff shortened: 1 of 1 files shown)


rcon/rcon.py
```diff
@@ -494,72 +494,22 @@ def remove_admin(self, player_id) -> bool:
...
+            PLAYER_ID: ban["userId"],
             # TODO FIX
             "timestamp": None,
-            "ban_time": date,
-            "reason": reason,
-            "by": by.replace('"', ""),
+            "ban_time": ban["timeOfBanning"],
+            "reason": ban["banReason"],
+            "by": ban["adminName"],
             "raw": ban,
```
FlorianSW, line 513: I don't know who actually uses this raw format. But this now changes from str to dict. Would you be able to join the dict values to a str which looks like the original one? :)

FlorianSW (changes requested):
Appart from that looks very good, thanks! :)

PR #1115, opened 2025-10-08
add Search Player in Team View

2KU77B0N3S:


Dorfieeee (changes requested):
Hey @2KU77B0N3S !

I quite don't understand the need for autocomplete with the select ability.
Apart from the fact that it breaks the UI even when only a couple players are selected, it does not allow to search for and select several players with similar names. That can be very helpful, when you search for community players.

I came up with another solution, where the search works as a filter for the entire team view.
This way the user can easily select all visible players or find that one particular player.
Also, scores and counts are only sums of all visible players.

What do you think?

https://github.com/user-attachments/assets/1b41fd4b-fbee-44e2-bd0f-dcecf64d79b2

PR #1117, opened 2025-10-10
fix: ensure that player_id is always treated as a string when add or remove vip

fxsobr:
When a VIP was added through the /api/add_vip endpoint, the response returned success (HTTP 200). However, when checking for the inserted VIP, it did not exist. After reviewing the logs, an error was identified — it occurred due to a comparison being made between an integer and a string.

the player_id parameter is typed as str in the function signature, but somewhere in the call chain, it's being passed as an integer.

Solution:
Ensure that player_id is converted to a string at the beginning of both the add_vip and remove_vip methods.

### DIFF
--- a/rcon/rcon.py
+++ b/rcon/rcon.py
@@ -88,7 +88,6 @@ LOG_ACTIONS = [
 ]
 logger = logging.getLogger(__name__)
 
-
 CTL: Optional["Rcon"] = None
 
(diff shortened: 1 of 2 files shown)


fxsobr (commented):
Created a new function to convert the player_id, wich was an integer, back to a string.

PR #1120, opened 2025-10-10
UI: General Settings rework

Dorfieeee:
The current design is annoying and does not work very well.
I hope this simple refresh makes things easier to work with.
image
image


Dorfieeee:
@FlorianSW it is a complete replacement of the previous code.

PR #1138, opened 2025-10-20
Fix live view kpm stat

Fragger:
Currently kpm on live view uses `current_playtime_seconds` from a players profile, which for the first match they play in works fine, but once the map flips and the kills reset this results in the kpm shown being very low and incorrect.  Instead limit the time used to the lower of `current_playtime_seconds` or the time since the match started.

With kills being broken and still returning 0 in rcon v2 this also is 0 but once kills is hopefully fixed this should result in a more useful metric for longer connected players.


Dorfieeee:
@Fragger are you sure the kills return 0? It seems to me the `get_team_view` endpoint returns 0 or the frontend does not handle it correctly I did not check. Other endpoints like the one used for public stats work and the kills are saved in the match history.

ElGuillermo:
There's also a "poor man's way" to calculate individual playtime :
(offense + defense) / 20 = playtime (min)

PR #1141, opened 2025-10-22
UI: Action form ability to search and add new records

Dorfieeee:
- A user can open WATCH, VIP, BAN, BLACKLIST, ADMIN and FLAG form dialogs from Records > Players page.
- A user can search for existing or add new records within the form ( only when opened on that page )
- A user can remove action recipients within the form

image
image


Fragger:
The search add/remove on the Records > Players page seems to all work nicely.

It looks like this is a problem before the tweak on the Settings > Vip page as well, but after picking a player by name or ID, then clearing them, and then trying to pick the same player again, doesn't work.

Dorfieeee:
That was an easy fix.

FlorianSW (approved):

PR #1152, opened 2025-10-31
Add CE OE DE and SE options to players live view

Fragger:


caranci (approved):
I like it. Works in my build.

PR #1154, opened 2025-11-02
hackalicious! enrich player kills from live game stats...

caranci:
... until hll server returns kills in player info again. i got sick of seeing 0 kills in the live and team views, but noticed that the watch kill rate module was still working.

this commit uses the cached live game stats to enrich the player stats in get_team_view with a kills count.

this works on my build (which includes some other customizations). i don't know what impact it might have on performance.


caranci:
changing to draft until abu finds out when a fix is expected from t17

Fragger:
Looks like this did get fixed with U19 and #1172/ 87216b2

PR #1160, opened 2025-11-04
feat: add multi server permissions

fxsobr:
## UI Changes
- **Server selector** now displays only servers the user has permission to access
- **Error page** shows clear message when accessing unauthorized server: "You do not have permission to access server X"
- **"Go to Home" button** redirects user to first allowed server instead of generic home page
- **"Switch Account" button** logs out current user and redirects to login page
- **Server selector** is disabled (but still visible) when user has access to only one server
- **No infinite loading** - permission errors are caught and displayed immediately

### Server Permission Model
- Administrators can now restrict specific users to specific servers in multi-server deployments
- **Default behavior:** Users with `can_view_other_crcon_servers` permission see all servers (backward compatible)
- **Opt-in restrictions:** Add `UserServerPermission` records to limit user to specific servers only
- **Superusers** always have unrestricted access regardless of permissions
- **Unique constraint** prevents duplicate permissions for same user/server combination

### GET `/api/get_server_list`
Returns filtered list of servers based on user permissions:
```json
{
  "result": [
    {
      "server_number": 2,
      "name": "Server 2", 
      "link": "http://server2.example.com:8010"
    },
    {
      "server_number": 3,
      "name": "Server 3",
      "link": "http://server3.example.com:8010"
    }
  ]
}
```

### Unauthorized Access Response
All API endpoints return 403 when user lacks server permission:
```json
{
  "result": null,
  "command": "get_connection_info",
  "failed": true,
  "error": "You do not have permission to access server 1. You only have access to servers: [2, 3]. Contact an administrator if you believe this is an error.",
  "text": "You do not have permission to access server 1...",
  "message": "You do not have permission to access server 1...",
  "server_number": 1,
  "allowed_servers": [2, 3]
}
```

### New Model: `UserServerPermission`
```python
class UserServerPermission(models.Model):
    user = ForeignKey(User, on_delete=CASCADE, related_name="server_permissions")
    server_number = IntegerField(help_text="Server number that the user can view")
    created = DateTimeField(auto_now_add=True)
    
    class Meta:
        unique_together = ("user", "server_number")
        ordering = ("server_number",)
```

### Migration: `0023_add_user_server_permission.py`
- Creates `api_userserverpermission` table

### Backend
- **New middleware:** `rconweb/api/middleware.py` - `ServerAccessMiddleware` class
- **Model addition:** `rconweb/api/models.py` - `UserServerPermission` model with docstring explaining behavior
- **Admin registration:** `rconweb/api/admin.py` - `UserServerPermissionInline` and `UserServerPermissionAdmin`
- **Endpoint filtering:** `rconweb/api/multi_servers.py` - `get_server_list` filters by user permissions
- **Settings update:** `rconweb/rconweb/settings.py` - Middleware registered after `AuthenticationMiddleware`

### Frontend
- **Auth hook refactor:** `rcongui/src/hooks/useAuth.js` - 3-step authentication flow with server access validation
- **Error state management:** New `permissionError` state separate from authentication errors
- **Smart navigation:** `handleGoToHome` fetches allowed servers and redirects to first one
- **Logout integration:** `handleSwitchAccount` calls logout API before redirecting
- **Selector update:** `rcongui/src/components/layout/SelectContent.js` - Disable instead of hide when single server

### Grant access to specific servers (restricted access)
```python
from django.contrib.auth import get_user_model
from rconweb.api.models import UserServerPermission

User = get_user_model()
user = User.objects.get(username="john")

# User can only access servers 2 and 3
UserServerPermission.objects.create(user=user, server_number=2)
UserServerPermission.objects.create(user=user, server_number=3)
```

### Grant access to all servers (default)
```python
# Don't create any UserServerPermission records
# User automatically sees all servers (if they have can_view_other_crcon_servers permission)
```

### Remove specific server access
```python
# Remove access to server 2
UserServerPermission.objects.filter(user=user, server_number=2).delete()
```

### Check user's allowed servers
```python
# Get all servers a user can access
allowed_servers = UserServerPermission.objects.filter(user=user).values_list('server_number', flat=True)
print(list(allowed_servers))  # [2, 3]
```

## Preview

[Preview](https://youtu.be/ozKsvdLDEE8)

### DIFF
--- a/rcongui/src/components/shared/ServerAccessDenied.jsx
+++ b/rcongui/src/components/shared/ServerAccessDenied.jsx
@@ -1,11 +1,11 @@
-import { Box, Card, CardContent, Stack, Typography, CircularProgress, Button } from "@mui/material";
+import { useQuery } from "@tanstack/react-query";
+import { Stack, Typography, Button, Paper } from "@mui/material";
 import LockIcon from "@mui/icons-material/Lock";
-import ServerIcon from "@mui/icons-material/Dns";
+import DnsIcon from "@mui/icons-material/Dns";
 import LogoutIcon from "@mui/icons-material/Logout";
 import { cmd } from "@/utils/fetchUtils";
-import { useQuery } from "@tanstack/react-query";
 
-export default function ServerAccessDenied({ errorMessage }) {
+export default function ServerAccessDenied() {
   const { data: availableServers = [], isLoading } = useQuery({
     queryKey: ["accessible_servers"],
     queryFn: cmd.GET_ACCESSIBLE_SERVERS,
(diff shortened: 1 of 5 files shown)


fxsobr:
@Dorfieeee noted, thanks!

Dorfieeee:
I did not know the permission exists for that endpoint and I thought that was something you introduced with this proposed change. I guess if a user does not have that permission he can still log in to another servers but he needs to guess the url 🤣

Dorfieeee:
I am thinking now why does the server list exclude the server the user is sending the request from.
For that reason the frontend app needs to send 2 requests to get the complete list and display the server selection dropdown component.
I would personally edited the `get_server_list` endpoint and return all servers and added an extra field to specify which one is the current one.

fxsobr:
@Dorfieeee, I made a few changes based on our chat, would be great if you could test them and let me know if you spot anything that could be improved or fixed.
I added an optional include_current parameter to the get_server_list endpoint so it can include the current server in the response with a current: true field.
Also tweaked global-state.js so the frontend now makes just one request instead of two.
Added a permission check in forward_request and forward_command to make sure users can only send changes to servers they actually have access to.
And finally, I updated the error handling — instead of showing the panel, it now displays a friendly error message. There’s also a new component just for showing server permission errors.

image

PR #1164, opened 2025-11-11
UI: map filters, map change actions

Dorfieeee:
Switch map filters from selects to checkboxes.
Added action side to map change page.
The user can easier reset the current map and change to the next map.

image


caranci (approved):
Looks okay in my build.

PR #1166, opened 2025-11-13
UI: all assets stored in one place

Dorfieeee:
Both rcongui(admin) and rcongui_public(stats) frontend applications share the same assets stored in root /assets directory.
These assets are copied to their respective containers during build time.
Added `copy-assets` commands for both frontend applications to import assets during development.
Updated README files to reflect these changes for devs.

+ 👍🏻  both FE apps using the same pool of assets
  + no need to copy images to multiple directories
+ 👍🏻  end users can replace images for their own before build time

- 👎🏻  FE dev needs to run one extra command


caranci:
I think I'm running this commit in my build without any issues.

PR #1181, opened 2025-12-05
Enable sorting on Player name in Public Stats

Fragger:
Had someone from the community ask me about being able to do this :)


Dorfieeee (approved):
👍🏻

PR #1182, opened 2025-12-05
New Analytics service

Dorfieeee:
A new service that captures the current HLL server player count and hosting device resource usage every 5 minutes.


config/supervisord.conf
```diff
@@ -103,6 +103,13 @@ startretries=5
 autostart=true
 autorestart=unexpected
 
+[program:analytics]
```
FlorianSW: I don't really like that this spawns _yet another_ python process, with it's own RCon connection to fetch the player count stuff.

What about moving this to the worker program that already exists. I believe it runs more often than 5 mins, hence we can probably find a suitable multiply of it and let it run every n-th run of the worker run?

FlorianSW (changes requested):

PR #1202, opened 2026-01-10
Fix: Votemap Whitelist page errors when whitelist contains non-existing

FlorianSW:
map

With the recent update, Stalingrad was renamed to a different map ID, however, Votemap whitelists might still contain the old IDs of the map. When that is the case, the votemap whitelist page simply errors, as these maps will be replaced with "undefined" instead of their map object counterpart.

I decided to simply remove the map IDs from the votemap whitelist selected maps list. An alternative would've been to make a migration server-side to replace the old map ID with the new one.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -46,6 +46,7 @@ then
   python -m rcon.user_config.seed_db
   # TODO: Temporarily handle converting users old scorebot URLs to the new format; remove in a few releases
   SERVER_NUMBER=${SERVER_NUMBER} LOGGING_PATH=/logs/ LOGGING_FILENAME=api_${SERVER_NUMBER}.log python -m rcon.cli port_legacy_scorebot_urls
+  SERVER_NUMBER=${SERVER_NUMBER} LOGGING_PATH=/logs/ LOGGING_FILENAME=api_${SERVER_NUMBER}.log python -m rcon.cli remove_orphaned_map_ids
   ./manage.py register_api
   cd rconweb
   ./manage.py collectstatic --noinput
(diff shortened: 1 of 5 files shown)


Dorfieeee:
I am not sure right now how votemap behaves when invalid map is given but yeah I think it would be much better to make a check while running the maintenance container so ensure that neither autosettings nor votemap or anything else does not contain some old ids and other values.

FlorianSW:
Sorry for my late response 🙈 

> I am not sure right now how votemap behaves when invalid map is given but yeah I think it would be much better to make a check while running the maintenance container so ensure that neither autosettings nor votemap or anything else does not contain some old ids and other values.

Agreed for the Votemap part. The autosettings part shouldn't be an issue. It's "just" a config to invoke API commands. Hence it should be the task of the api command to handle cases where "old" map IDs are still put in. Furthermore I feel that touching the autosettings config is an overkill for this (with all the potential parsing and mapping of API commands and map lists) for a potential issue that is limited.

I added a task to remove orphaned VoteMap IDs from the config in the latest commit. I also changed how the whitelist is persisted in redis. Before this, there were multiple ways on how the list was saved, either as a list of layers, or a list of strings, or a combination of both. That required several places to parse the list of maps, even though the type information already suggested that it is a set of parsed map layers. I made sure that we always only save the map IDs and parse the them when getting them back from redis. In that way the API can always safely expect that the type a method returns/accepts is acftually the contents of the value.

Dorfieeee (commented):
👍🏻

Dorfieeee (approved):
👍🏻

PR #1274, opened 2026-04-18
Get rid of unmaintained django trick

ElGuillermo:
- Replace cemathey's `django-directory` (https://github.com/cemathey/django-directory) fork from https://github.com/ASKBOT/django-directory with dedicated Django view.


ElGuillermo:
In hope we'll soon upgrade PostgreSQL and Django to their latest stable version, I've found a way to get rid of an Django "trick" cemathey used to display /logs files and folders using /api/logs endpoint in browser.
See : https://github.com/MarechJ/hll_rcon_tool/pull/291#discussion_r1284629093

The `django-directory` fork isn't mandatory anymore, replaced by a full dedicated view. User would still need to be logged as 'staff' to display the logs list and download logs.

image

Fragger (approved):
Gave this a try and it all seems to work well to me.  Nice to have the directory view with sizes and last modified.

PR #1279, opened 2026-04-26
Fix: Reduce duplicate log insert errors and Redis/RCON cache pressure

2KU77B0N3S:
This PR reduces avoidable backend load and log noise in high-traffic CRCON deployments.

It addresses three related operational issues:

- duplicate `log_line` inserts producing repeated database integrity errors
- Redis creating RDB snapshots too aggressively under normal CRCON write load
- concurrent cache misses causing multiple identical backend/RCON refreshes at the same time

## Problem

On busy servers, the log recorder can attempt to persist log lines that already exist. Because the `log_line` table already has a uniqueness constraint, these duplicate inserts currently surface as `IntegrityError` exceptions and noisy logs.

At the same time, the default Redis persistence setting:

```redis
save 60 1
```

can trigger frequent RDB saves, because CRCON naturally writes to Redis often. This can create unnecessary disk I/O and background save activity.

Finally, cached endpoints can suffer from a cache stampede when a hot key expires. Multiple requests may miss the cache at the same time and all regenerate the same value, increasing backend and RCON pressure.

### Log recorder

- Batch-builds log rows before inserting.
- Uses PostgreSQL `ON CONFLICT DO NOTHING` against the existing `unique_log_line` constraint.
- Keeps the previous SQLAlchemy object insert path as a fallback for non-PostgreSQL dialects.
- Avoids unnecessary work when there are no logs or player IDs to process.

### Redis persistence

Updates the default Redis snapshot policy from:

```redis
save 60 1
```

to:

```redis
save 900 1
save 300 100
save 60 10000
```

This keeps persistence enabled, but avoids triggering a background save for every small write burst.

### Redis cache refresh

Adds a lightweight Redis lock around cache regeneration.

When a cache key expires:

- the first request refreshes the value
- concurrent requests briefly wait for the refreshed value
- if Redis is unavailable, the original fallback behavior is preserved
- lock release is token-checked to avoid deleting another process' lock

## Expected Benefits

- Fewer noisy database integrity errors from duplicate log ingestion.
- Lower Redis background save frequency under normal write load.
- Reduced duplicate work during cache misses.
- Less burst pressure on expensive cached endpoints and RCON-backed calls.
- No schema changes required.

## Notes

This PR intentionally does not change blacklist retry behavior or RCON command retry semantics. It is limited to log recording, Redis persistence, and cache refresh behavior.

## Validation

Tested with:

```bash
python -m py_compile rcon/logs/recorder.py rcon/cache_utils.py
```

Also validated in a Docker image context with:

```bash
python -m py_compile rcon/logs/recorder.py rcon/cache_utils.py
```

### DIFF
--- a/config/redis.conf
+++ b/config/redis.conf
@@ -1,2 +1,4 @@
-save 60 1
+save 900 1
+save 300 100
+save 60 10000
 maxclients 100000
--- a/config/redis.conf
+++ b/config/redis.conf
@@ -1,4 +1,2 @@
-save 900 1
-save 300 100
-save 60 10000
+save 60 1
 maxclients 100000
--- a/rcon/cache_utils.py
+++ b/rcon/cache_utils.py
@@ -2,6 +2,8 @@ import functools
 import logging
 import os
 import pickle
+import time
+import uuid
 from contextlib import contextmanager
 from typing import Callable
 
@@ -70,6 +72,11 @@ class RedisCached:
             return self.key_prefix.encode() + b"__" + params
         return f"{self.key_prefix}__{params}"
 
+    def lock_key(self, key):
+        if isinstance(key, bytes):
+            return b"lock_" + key
+        return f"lock_{key}"
+
     @property
     def __name__(self):
         return self.function.__name__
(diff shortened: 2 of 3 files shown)


config/redis.conf
```diff
@@ -1,2 +1,4 @@
-save 60 1
-maxclients 100000
\ No newline at end of file
+save 900 1
```
FlorianSW: Why would we want to do that? You mentioned in the PR summary, that this is supposed to reduce rdb snapshots when bursting cache key changes, right?

With `save 60 1` we already make as napshot at max once every 60 seconds, isn't it? reducing that to 15 or even 5 minutes essentially means that we increase the amount of cache changes that one looses when redis crashes.

And especially in the case of a high-volume CRCon deployment (possibly with multiple servers), wouldn't that increase the chance to have over 100 or even 10k key changes (while I'm not even sure that we have 10k keys at all, that we could change to trigger the 60-second rule at all).

FlorianSW (commented):
The changes regarding inserting log lines in batch are a good idea. With regards to using postgres-specific logic: I would generally consider not doing that, however, mostly for consistency reasons. However, there is a propper fallback and the benefit is clear to me, hence I'm fine with that.

Regarding the change of redis rdb: See my comment, would love to hear your feedback.

Locking on refreshing in redis: The code overalls should work from what I see, however, it makes the whole stuff much more complicated. And especially hard to read :D Maybe you can find a more straight forward solution? I just reviewed it for a short time and didn't find a good solution, either, or a specific line I would point out that needs to be changed. Hence I just leave this more or less generic comment, sorry for that 😬

2KU77B0N3S (commented):
redis.conf fixed to have no change

FlorianSW (approved):

PR #1299, opened 2026-05-15
Proposal: Remove GSP specific server name change feature

FlorianSW:
Changing the server name is only supported for GTX, which is just one of many game service providers. The implementation of this feature is somewhat "heavy" as it requires FTP/SFTP access by CRCon to the actual game server. It also seems that no-one plans on extending this feature to other GSPs.

Furthermore: This works for the name of the game server, only. All other settings, such as the password of the server (which might be much more useful to the community) is not supported by this way, significantly limiting the usefulness of this feature from my point of view.

I therefore propose removing support for this feature from CRCon to simplify maintenance, dependency footprint, and the need to have Game Server FTP/SFTP credentials in environment variables.


caranci:
looks good... after a quick read.

Fragger (approved):
Gave this a try and looks good to me.  I think its probably the right move to remove this.

Fragger:
I fixed the requirements.txt merge conflict but looks like the backend test is failing from the migration that creates the default permissions still trying to add `can_view_server_name_change_config`

PR #1323, opened 2026-06-17
Add hook-specific admin ping trigger words

Foxx-l:
# Summary

Admin Ping webhooks (settings/webhooks/admin-ping) support multiple Discord hooks, but trigger words were applied globally. 
This PR adds an optional, hook specific "trigger_words" field. This way pings can be routed to dedicated hooks.
The hooks fire on both their specific and the global trigger words, in order to maintain backwards compatibility. 

# Changes

### rcon/user_config/webhooks.py

- New `TriggerWordMentionWebhook` model (subclass of `DiscordMentionWebhook`) adding a validated `trigger_words` field; integrated by `AdminPingWebhooksUserConfig.hooks`
Note: I'm not sure if im happy with the name `TriggerWordMentionWebhook`.

- Extracted the existing case-fold/strip/dedupe/sort logic into a shared `_normalize_trigger_words` helper, reused by both the global (`AdminPingWebhooksUserConfig`) and per-hook fields.
- `save_to_db` key check now allows the optional per-hook `trigger_words` key.

### rcon/discord_chat.py

- Each constructed webhook is kept paired with its config (`(webhook, hook_config)`) so per-hook trigger words/mentions are available at send time.
- New `hook_trigger_words(hook)` returns the shared + per-hook union; `create_admin_ping_message(log, hook)` is now evaluated per hook; `send_chat_message` loops hooks and enqueues only those that matched.

### rcon/message_variables.py

- `{admin_ping_trigger_words}` now aggregates the top-level list plus every hook's words (de-duplicated).

### rcongui/src/pages/settings/_data/webhooks/admin-ping.js
- settings notes document shared vs. per-hook trigger words

# Testing 

###  tests/test_discord_handler.py
Updated/added unit tests to pass hook config in `create_admin_ping_message`
- per-hook words fire only their own hook
- shared top-level words fire every hook

### tests/test_message_variables.py 
- Updated the mock for the aggregated variable

### DIFF
--- a/rcon/user_config/webhooks.py
+++ b/rcon/user_config/webhooks.py
@@ -91,6 +91,30 @@ class DiscordMentionWebhook(DiscordWebhook):
         return list(role_ids)
 
 
+def _normalize_trigger_words(vs: list[str]) -> list[str]:
+    """Lower-case, strip and de-duplicate trigger words."""
+    processed_words = set()
+    for v in vs:
+        processed_words.add(v.lower().strip())
+    return sorted(processed_words)
+
+
+class TriggerWordMentionWebhook(DiscordMentionWebhook):
+    """A mention webhook that may define its own trigger words.
+
+    These are added on top of the config level
+    ``AdminPingWebhooksUserConfig.trigger_words`` (which apply to every hook), so
+    a hook fires on the shared words plus any extra words defined here.
+    """
+
+    trigger_words: list[str] = pydantic.Field(default_factory=list)
+
+    @pydantic.field_validator("trigger_words")
+    @classmethod
+    def ensure_case_unique(cls, vs):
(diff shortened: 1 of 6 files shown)


rcon/user_config/webhooks.py
```diff
@@ -167,11 +187,11 @@ def save_to_db(values: AdminPingWebhookType, dry_run=False) -> None:
         for obj in raw_hooks:
             key_check(
                 WebhookMentionType.__required_keys__,
-                WebhookMentionType.__optional_keys__,
+                WebhookMentionType.__optional_keys__ | {"trigger_words"},
```
FlorianSW, line 190: can you explain this line? I don't get why you need/do that tbh 😬

PR #1329, opened 2026-06-27
Replace pip, black, isort, pyenv with uv and ruff

FlorianSW:
uv and ruff are blazingly fast and combine the features of currently multiple tools. Black and isort can be replaced by ruff format. The codestyle itself should be mostly the same (applying this also to the current repo to be consistent, that's where the changes in other files come from).

uv replaces pip and pyenv by aiming to be a full project setup tool, which manages dependencies and a python environment with the correct python version.


Fragger:
I'd like to see this in.  We probably just need to rerun the format after some bigger pr's are in to avoid merge conflicts and then merge this.

Fragger:
Should formatting also be enforced in PRs with a `ruff format --check` action?

FlorianSW:
The most interesting part to take a look on is probably this:
https://github.com/MarechJ/hll_rcon_tool/pull/1329/changes/7c6c70aa90fef97b809793597c55b00ae678a8a2

Some of our database fields handling datetimes are timezone-unaware. Ruff enforces that datetimes always have a timezone. I added UTC for that, as all relevant timezones already should be UTC. And postgres saves UTC for TZ-unaware times anyway. As I did not want to make a database migration to switch to a tz-aware timestamp DB type, I added a separate decorator handling this (so that datetimes coming from the DB have a TZ as well to ensure comparing them works).

Fragger:
> The most interesting part to take a look on is probably this: [7c6c70a](https://github.com/MarechJ/hll_rcon_tool/commit/7c6c70aa90fef97b809793597c55b00ae678a8a2)
> 
> Some of our database fields handling datetimes are timezone-unaware. Ruff enforces that datetimes always have a timezone. I added UTC for that, as all relevant timezones already should be UTC. And postgres saves UTC for TZ-unaware times anyway. As I did not want to make a database migration to switch to a tz-aware timestamp DB type, I added a separate decorator handling this (so that datetimes coming from the DB have a TZ as well to ensure comparing them works).

That's been on my radar for a while actually.  This is the right thing to do imo and the db does not need to be tz aware.

Fragger (approved):
Looks good to me.  Maybe want to give time for a couple other collaborator's to look through?

FlorianSW:
> Looks good to me. Maybe want to give time for a couple other collaborator's to look through?

I'll leave this PR for some days, if no-one objects, I just go ahead and merge it 😬

rconweb/rconweb/settings.py
```diff
@@ -37,13 +37,16 @@
...
+if HLL_MAINTENANCE_CONTAINER or HLL_WH_SERVICE_CONTAINER:
     ENVIRONMENT = "undefined"
+else:
+    try:
+        config = RconServerSettingsUserConfig.load_from_db()
+        ENVIRONMENT = re.sub(
+            "[^0-9a-zA-Z]+", "", (config.short_name or "default").strip()
+        )[:64]
+    except (ValidationError, IndexError):
+        ENVIRONMENT = "undefined"
```
Dorfieeee, line 49: The issue was the in `entrypoint` it runs `makemigrations` inside `maintenance` container where the `SERVER_NUMBER` is not set. 

The exception handling used to catch any `Exception` but in 326f2912 was changed to `(ValidationError, IndexError):` however `RconServerSettingsUserConfig.load_from_db()` raises `ValueError` when the `SERVER_NUMBER` is not set.

Dorfieeee (changes requested):
Image

Dorfieeee (approved):

PR #1331, opened 2026-07-06
Add post-match Team Balance automod (squad-based, steamroll-triggered)

brian-mcmullen:
## Summary

Adds a new **Team Balance** automod that rebalances the teams **after a match ends** by moving whole squads between sides, rather than shuffling individual players. Squads are evaluated by a configurable combat-effectiveness score so the mod moves the smallest set of squads needed to even things out.

Balancing only kicks in after a **steamroll**, which is deliberately defined by **match dynamics, not score margin**:

- **Fast match** — the match ended faster than a configurable threshold (default 30 min), or
- **Swap-aware win streak** — the same group has won repeatedly. Because HLL swaps Allies/Axis each match, a repeated winner shows up as the raw winner *alternating*; the streak logic accounts for that.

A long, decisive game (e.g. a 60-minute 5-0) is treated as a **grind, not a steamroll**, and does **not** trigger balancing.

## How it works

- Hooks `@on_match_end` (runs in the log loop, not the 5s punish loop).
- Reads `get_team_view()` for per-squad `combat/offense/defense/support`, squad `type`, and players.
- **Two passes**, each selecting the minimum set of whole squads to move:
  1. **Armor** — balanced by squad count, or by a score-gap trigger.
  2. **Infantry** — balanced by headcount and weighted score gap.
- Commander and artillery are never moved; recon is optionally excluded.

## Safety and guards

- **Seeding guard** — skips while the server is seeding (population at/below a threshold).
- **Minimum-player floor** — skips low-population matches.
- **Dry-run mode** — logs the planned moves without switching anyone.
- **Stale-event guard** — ignores match-end events older than 10 minutes, so re-read logs after a restart can never rebalance current teams or corrupt the streak history.
- **Per-player best-effort execution** — a failed switch is logged and skipped; only players actually moved are messaged and audited.
- Optional per-player switch message and **Discord audit** logging.
- Winner history is capped (30 entries) with a 14-day TTL; state is serialized as JSON (not pickle).

## Configuration

New `AutoModTeamBalanceUserConfig` (enabled off by default) exposing: enable/dry-run, fast-match minutes, win-streak threshold, seeding + min-player thresholds, armor balancing (count delta + score-gap trigger), infantry headcount delta + score-gap threshold, per-category score weights, recon exclusion, max players to switch, and the switch message. Wired into the settings UI, API commands, and `seed_db`.

## Permissions

Gated behind two dedicated permissions (`can_view_team_balance_config`, `can_change_team_balance_config`) with the accompanying migration.

## Testing

- 16 unit tests for the automod (steamroll gating, swap-aware streak, armor/infantry selection, seeding/min-player guards, dry-run, stale-event guard).
- 52 automod tests pass overall with no regressions in the existing automods.

## Notes

- Additive only — no existing behavior is changed; the mod is disabled by default.

## Update: optional average-level balancing

Adds an opt-in **"Balance by average level"** option (off by default) so a heavily level-stacked team (e.g. many level 500s vs level 5s) gets evened out:

- A large **average player-level gap** (above `level_gap_threshold`) becomes an additional match-end trigger, alongside the existing steamroll triggers (fast match / swap-aware win streak).
- Rebalancing is done as **headcount-preserving squad swaps** (a high-level squad is exchanged for a low-level one), so equal-sized teams stay even on numbers while high-level players get spread out.
- New config: `balance_by_level` and `level_gap_threshold` (surfaced in the settings UI via the model schema; no DB migration, config is JSON).

The win-streak trigger was already configurable via `win_streak_threshold` (`0` disables it).


rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+        return self.config.enabled
+
+    # ------------------------------------------------------------------ #
+    # Entry point
+    # ------------------------------------------------------------------ #
+    def on_match_end(self, rcon, struct_log) -> None:
+        try:
+            self._on_match_end(rcon, struct_log)
+        except Exception:
+            self.logger.exception("Team balance: unexpected error on match end")
```
FlorianSW: would be nice to have some mor einfo here, at least the exception message would be nice. The traceback would be helpful for debugging purposes.

rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+    config: AutoModTeamBalanceUserConfig
+
+    def __init__(
+        self, config: AutoModTeamBalanceUserConfig, red: redis.StrictRedis | None
+    ):
+        self.logger = logging.getLogger(__name__)
+        self.red = red
+        self.config = config
+
+    def enabled(self) -> bool:
```
FlorianSW: is this even used?

rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+        """Global on/off switch."""
+        return self.config.enabled
+
+    # ------------------------------------------------------------------ #
+    # Entry point
+    # ------------------------------------------------------------------ #
+    def on_match_end(self, rcon, struct_log) -> None:
+        try:
+            self._on_match_end(rcon, struct_log)
+        except Exception:
```
FlorianSW: can this actually ever happen? And if so, why would we want to have this catched with a generic error message and not bubble up?

rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+        if allied > axis:
+            return "allies"
+        if axis > allied:
+            return "axis"
+        return None
+
+    def _record_and_get_winners(self, winner: Optional[str]) -> list[Optional[str]]:
+        """Persist the winner (newest first) and return recent winners."""
+        winners: list[Optional[str]] = []
+        if self.red is not None:
```
FlorianSW: in what cases does the redis client being None make sense? I would expect that this is being sorted out in the init function: If redis is None, it should raise an exception and never end up in this state, isn't it? The code in this class should always safely assume that redis is not None.

rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+
+    @staticmethod
+    def _match_duration_minutes(struct_log) -> Optional[float]:
+        """Duration of the just-ended match, in minutes (None if unknown)."""
+        end_ms = struct_log.get("timestamp_ms")
+        if not end_ms:
+            return None
+        end_s = end_ms / 1000
+        try:
+            start_s = MapsHistory()[0].get("start")
```
FlorianSW: what is when the map history is empty? This should be handled explicitly, not by a try/except block.

rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+        except Exception:
```
FlorianSW: what exception would be possible here? Looks superfluous.

rcon/automods/team_balance.py
```diff
@@ -0,0 +1,719 @@
...
+                    "level_sum": sum(lvl for _, _, lvl in kept),
+                    "player_ids": [pid for pid, _, _ in kept],
+                    "player_names": [pname for _, pname, _ in kept],
+                }
+            )
+        return squads
+
+    def _is_movable_infantry(self, squad: dict) -> bool:
+        stype = squad.get("type")
+        if stype in _ALWAYS_PROTECTED_TYPES or stype == "armor":
```
FlorianSW: why isn't armor in the `_ALWAYS_PROTECTED_TYPES` list as well? Its only use is here, hence it should contain all of the protected types, shouldn't it?

FlorianSW (changes requested):
Apart from the comments I've the folowing in my head while reading the PR summary and the code:

1. What happens if, during moving players, the side is getting "full"? Did you try that out? I remember that the game server allows us to move players via RCon even when the the target side already has more than 50 players, right?
2. From a player experience perspective, I don't know if first moving a player to another side and then messageing them is the best we can do. This is a rather disruptive action, hence I would've expected that players first get a message, than a little cooldown and are moved afterwards. What do you think?

ElGuillermo:
Moving players, even after a message, isn't a good idea : I ran such a script some months ago and it only led to a massive disagreement (and, worst, loss of players).

rcon/automods/team_balance.py
brian-mcmullen: removed this catch so unexpected errors bubble to the existing hook dispatcher that provides the exception message + traceback

rcon/automods/team_balance.py
brian-mcmullen: my bad I must've missed that while tooling around seeing if I could get this to work

rcon/automods/team_balance.py
brian-mcmullen: removed this catch so unexpected errors bubble to the existing hook dispatcher

rcon/automods/team_balance.py
brian-mcmullen: Redis is now required by the constructor, which raises an explicit error if it is missing

rcon/automods/team_balance.py
brian-mcmullen: This now explicitly checks whether MapsHistory() is empty and returns None when there is no match-start record

rcon/automods/team_balance.py
brian-mcmullen: removed; the empty-history case is handled explicitly instead

rcon/automods/team_balance.py
brian-mcmullen: Armor is movable, but only during the dedicated armor-balancing pass, so it is not “always protected.” I renamed the set to _INFANTRY_EXCLUDED_TYPES and included armor there, making it clear that armor is excluded only from infantry balancing

brian-mcmullen:
> Apart from the comments I've the folowing in my head while reading the PR summary and the code:
> 
>     1. What happens if, during moving players, the side is getting "full"? Did you try that out? I remember that the game server allows us to move players via RCon even when the the target side already has more than 50 players, right?
> 
>     2. From a player experience perspective, I don't know if first moving a player to another side and then messageing them is the best we can do. This is a rather disruptive action, hence I would've expected that players first get a message, than a little cooldown and are moved afterwards. What do you think?

The armor and infantry selection passes already excluded moves that would put the target team above 50. I also added the same check to level-swap selection and recheck the target team immediately before each squad is moved; if the whole squad no longer fits, it is skipped. There is no atomic squad-transfer RCON command, so this is the safest available guard against players joining between selection and execution.

Agreed. Selected players are now notified before any move, then the automod waits a configurable 20 seconds by default. The message explains that the squad has been selected and will only be moved if the target team still has room. Administrators can set the delay to 0 if they explicitly want immediate switching.

Ultimately, I do agree that forced moves are disruptive. This is a feature that my server has been asking for, for some time. Repeated steamrolls kills our player count, especially in the middle of the night with fewer if any admins awake and on. I made this disabled by default, with a dry-run mode, and added that warning before moving. I did also make it attempt to move entire squads so that players who are friends can attempt to stay together. The way I see it, server admins don't have to use this tool, but hopefully when they do, its less disruptive.

PR #1332, opened 2026-07-06
Fix stale Live View: stop caching live API responses (frontend no-store + private nginx /api)

brian-mcmullen:
## Summary

Fixes the Live View player list and game header (map, score, player counts) not updating during a session, even though the logs panel updates fine. The only workaround users found was to log out and back in. This primarily affects CRCON hosts that are behind reverse proxies that cache requests.

## Root cause

The shared fetch wrapper `requestFactory` in `rcongui/src/utils/fetchUtils.js` issued every request with `cache: "default"`, which lets the browser use its normal HTTP cache. As a result:

- `get_gamestate` and `get_team_view` are **GET** requests, so their responses were eligible for the browser HTTP cache and were served **stale** from memory/disk cache.
- The logs panel stayed fresh because `get_recent_logs` is a **POST** request, and browsers never cache POST responses.
- Logging out and back in forces a full reload, which revalidates the cache for the first fetch, so the data appeared to "unfreeze" briefly.

This was verified by hitting the same endpoints directly (e.g. Postman), which **always returned fresh data** — confirming the staleness was introduced by the browser HTTP cache, not by the backend, the Redis `ttl_cache`, or the RCON layer. In the browser's Network panel the stale requests show `200 OK` served from `(disk cache)` / `(memory cache)`.

## Fix

Set `cache: "no-store"` in `requestFactory` so live API responses are always fetched from the network and never served from the browser cache. This applies to all API calls (they are dynamic, permission-gated, live-ops data that should not be browser-cached); static assets are served separately by nginx and are unaffected.

```diff
-    cache: "default",
+    cache: "no-store",
```

## Testing

- `esbuild` parses the changed file cleanly.
- Verified locally that the Live View team/player and gamestate queries (which already poll on a 10s `refetchInterval`) now reflect fresh data on each poll instead of being frozen by the HTTP cache.

## Update: server-side header for the private `/api` (proxy caches)

The frontend `cache: "no-store"` change stops the **browser** caching, but a caching **reverse proxy** in front of CRCON (WHM/Apache, Cloudflare, etc.) can still serve stale GET responses regardless of what the browser requests. The definitive fix is a server-side response header.

The `nginx.conf` **public** interface already sends `Cache-Control: no-store` on `location /api`, but the **private/admin** interface (which serves the Live View) did not — so its `get_gamestate` / `get_team_view` GET responses were cacheable. This is directly observable: the public stats page (`get_live_game_stats`, `get_live_scoreboard`, `get_public_info`) updates correctly because it goes through the public interface, while the admin Live View on the private interface went stale.

This adds the same `no-store` headers to the private `location /api`, so no cache layer (browser or proxy) stores live API responses.


rcongui/src/utils/fetchUtils.js
```diff
@@ -26,7 +26,11 @@ async function requestFactory({
   const requestOptions = {
     method,
     mode: "cors",
-    cache: "default",
+    // Never serve API responses from the browser HTTP cache. These endpoints
+    // return live data (e.g. get_gamestate, get_team_view); with the default
+    // cache mode, GET responses could be served stale from disk/memory cache
+    // while POST responses (logs) stayed fresh, freezing the live view.
+    cache: "no-store",
```
FlorianSW: this looks a bit duplicated and off. The HTTP client should (in a browser context) from my point of view not decide if a response is supposed to be cached or not. Especially after adding the nginx cache directives, do you think we can get rid of this part here?

rcongui/nginx.conf
```diff
@@ -69,6 +69,14 @@ server {
         }
 
         location /api {
+            # Never let the browser or any reverse proxy (WHM/Apache, Cloudflare, ...)
+            # cache live API responses. Matches the public interface below; without this
+            # a caching reverse proxy serves stale GET responses (get_gamestate,
+            # get_team_view) while POST-based logs stay fresh.
+            if_modified_since off;
```
FlorianSW, line 76: I mean, the most efficient way would be to support a proper no-cache workflow. For that the backend would need to support returning some kind of "last modified" information. This would allow us to better control what responses can be cached and which can't. E.g. the list of maps response is probably fine being cached for some minutes or hours as it does not change regularly.

However, not caching these endpoints for now is good enough.

FlorianSW (changes requested):

FlorianSW (approved):

PR #1345, opened 2026-08-09
Refactor votemap cache migrations and enhance Redis database handling

Dorfieeee:
- Updated entrypoint to use the new cache_migrations module.
- Introduced main.py for centralized migration logic.
- Added redis_databases.py for Redis client management across databases.
- Implemented votemap.py for handling votemap state migrations.
- Cleaned up maps_history.py by removing redundant functions.
- Added tests for votemap cache migration to ensure data integrity during upgrades.

This refactor improves the structure and maintainability of the votemap migration process.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -17,7 +17,7 @@ then
   # Upgrade durable Redis structures before any server-specific container starts.
   # The command discovers every populated logical Redis database so multi-server
   # installations are migrated in the same maintenance pass.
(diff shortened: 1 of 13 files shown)


Dorfieeee:
There are no functional changes really only splitting the vote_map into several files in its own package and creating redis migration just as for the MapsHistory as these two caches persist across updates and restarts. This will also ensure the whitelist is persisted after the redis key change so users don't have to create their whitelist again with the upcoming release

Fragger (approved):
👍

PR #1347, opened 2026-08-09
Reduce Redis scan load

bumseb1ene:
## Summary

Reduce Redis `SCAN` traffic generated by the webhook service and cache invalidation paths.

## Root cause

The webhook service discovers all queue keys on every loop iteration. With the default `HLL_WH_LOOP_SLEEP_TIME=0.006`, both queue key scans run roughly 167 times per second even when the set of queues has not changed.

In addition, several parameterless cached methods use `cache_clear()`. That method scans the complete Redis database for the function prefix, although these methods have exactly one cache key and can be invalidated directly.

## Changes

- Cache discovered webhook queue keys for one second while continuing to check the contents and existence of known queues on every service loop.
- Add `invalidates_for()` for exact invalidation of parameterless cached functions.
- Use exact invalidation for gamestate-derived caches, autobalance settings, idle autokick settings, and the player list on connect.
- Add unit coverage for queue discovery refreshes, deleted queues, and exact invalidation before and after a mutation.

New webhook queues may take up to one second to be discovered. Existing queues continue to be processed at the configured loop interval.

## Validation

- Python compile check passed.
- Black check passed for the changed formatted files.
- 12 directly relevant unit tests passed.
- Tested in a live two-server deployment: Redis `SCAN` traffic fell from approximately 1,798 calls/s to 8 calls/s (99.6% reduction) while both backends, webhook processing, and RCON supervisor processes remained operational.

### DIFF
--- a/rcon/webhook_service.py
+++ b/rcon/webhook_service.py
@@ -2,6 +2,7 @@ import asyncio
 import math
 import os
 import random
+import time
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
@@ -91,6 +92,11 @@ try:
     HLL_WH_LOOP_SLEEP_TIME = float(os.getenv("HLL_WH_LOOP_SLEEP_TIME"))
 except (ValueError, TypeError):
     HLL_WH_LOOP_SLEEP_TIME = 0.006
+
+# Queue contents still get checked every service loop. Redis key discovery is
+# intentionally slower because newly created webhook queues can tolerate this
+# small delay and SCAN is otherwise executed hundreds of times per second.
+QUEUE_KEY_REFRESH_INTERVAL_SECS = 1.0
 
 # Global datastructures to support associating hooks with locks
 _RATE_LIMIT_BUCKETS: defaultdict[str, asyncio.Lock | None] = defaultdict(lambda: None)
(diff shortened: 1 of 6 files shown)


rcon/webhook_service.py
```diff
@@ -830,6 +836,37 @@ def get_all_queue_keys_not_empty(
     return populated_queues + populated_transient_messages
 
 
+@dataclass
+class WebhookQueueKeyCache:
```
FlorianSW, line 840: Wouldn't it make more sense to split the two tasks here:
- one (even less frequent than 1 once every second) task to fetch _existing_ queue IDs (even if they are empty). Queue IDs are probably not needed to be discovered more often than every couple of seconds.
- another task that takes the _known_ queue IDs and just checks their emtiness with llen/exists

I think that this would allow us to reduce the amount of traffic towards redis keyspace even more, what do you think?

FlorianSW (commented):
Apart from the conceptual comment, it looks good.

rcon/webhook_service.py
bumseb1ene: Thanks! The current WebhookQueueKeyCache already separates both frequencies logically: queue IDs are discovered once per second, while the known IDs are checked with LLEN/EXISTS on every service loop. However, I agree that this separation is not very explicit and discovery can run less frequently.
One caveat is that Redis removes list keys once they become empty, so empty queue IDs cannot be discovered through SCAN. Persistently tracking those would require a separate queue registry, for example a Redis set updated during enqueueing.
For this PR I can make the two phases explicit and increase discovery to every five seconds. A persistent registry could then remove recurring keyspace scans entirely in a follow-up.

FlorianSW: just do the last part for now :) Should be good enough, if I understood it correctly 😅

FlorianSW: @bumseb1ene are you still working on this? :)

PR #1349, opened 2026-08-11
HLL Vietnam support

### LINKED ISSUE #1298: HLL Vietnam support
Vietnam is around the corner, so it'd be good to get started on adding support for it. I'm creating this issue mainly to outline all the information I know and have received thus far.

---

Right off the bat, I don't think much work is needed to support it, luckily. RCON is practically identical. The only substantial differences are the new roles, factions, weapons, maps, etc.

- Factions: IDs can be found below. DM me for icons.
- Roles: IDs can be found below. DM me for icons.
- Maps: TBA
- Weapons: DM me for names.
https://github.com/timraay/hllrcon/blob/feat/hll-vietnam/hllrcon/responses.py#L151-L260>

Note that above link also outlines changes in how the game returns platform names for both `GetServerInformation->players` and `GetServerInformation->serverconfig` commands.

---

### What needs to happen:
(in my opinion; subject to change)
- Add a `HLL_GAME` value to `default.env`.
  - Supported values are `hll` and `hllv`.
  - Defaults to `hll` if not provided to maintain backwards compatibility.
  - Expose via `/api/get_public_info`
  - Not sure yet what else we might want to use this for but I bet it'll be very useful for development.
  - Alternative idea would be to detect this dynamically via RCON, however, I imagine it'll be less convenient from a developer's perspective.
- Add support for the new roles.
- Add support for the new factions.
- Add support for the new weapons (mainly for stats).
- Maybe add an indicator and/or cosmetic changes to the UI?
- The relation between `PlayerAccount` and `PlayerSoldier` needs to be changed from one-to-one to one-to-many.
  - Each `PlayerAccount` now corresponds to multiple `PlayerSoldier`s, namely one for each game.
  - This will cause a number of cascading changes to both front- and back-end.
  - This will break many third party extensions. See if we can somehow preserve some of the old API behavior?

Maps and game modes will have to be done at a later stage.

---

### We're looking for contributions!
If you are willing to take on (parts of) this task, please leave a comment below and open a PR.

Dorfieeee:
Resolves: https://github.com/MarechJ/hll_rcon_tool/issues/1298


caranci:
maybe have the code validate the value is either "hll" or "hllv".
maybe in `ServerInfo.from_env()`?

rcon/types.py
```diff
@@ -169,6 +170,32 @@ class ServerInfoType(TypedDict):
...
+    password: str | None = None
+    game: str = "hll"
+
+    @classmethod
+    def from_env(cls) -> "ServerInfo":
+        return cls(
+            host=os.getenv("HLL_HOST"),
+            port=os.getenv("HLL_PORT"),
+            password=os.getenv("HLL_PASSWORD"),
+            game=os.getenv("HLL_GAME") or "hll",
```
FlorianSW: should validate against possible values, `hll` or `hllv`?

FlorianSW (commented):
apart from that looks good so far.

timraay (approved):

tests/test_game_profiles.py
```diff
@@ -0,0 +1,209 @@
...
+    assert hll_layer.map.id == "carentan"
+    assert hllv_layer.id == "carentan_warfare"
+    assert hllv_layer.map.id == UNKNOWN_MAP_NAME
+    assert hllv_layer.game_mode is GameMode.DOMINATION
+
+
+def test_hllv_profile_loads_hllv_catalog_with_logical_sides():
+    layer = HLLV_PROFILE.parse_layer("wdeve_offensivenva_day")
+
+    assert layer.map.id == "WDEV_E"
```
timraay, line 78: This is gonna break when upgrading hllrcon v2.0.0.1 by the way, which will change it to be `wdeve` instead.

PR #1350, opened 2026-08-12
Fix/consecutive offensive maps in selection when disabled

Dorfieeee:
when creating a new selection it is possible old `current_map` was cached - switched from using `current_map` to `MapsHistory.current_map` and use `current_map` as a fallback value


Fragger (approved):
Looks good and makes sense 👍

FlorianSW (approved):

PR #1352, opened 2026-08-12
Fix/increase rate limit for public info endpoint

Dorfieeee:
Increase from 10 to 60 request per minute per IP as it was uncommon to run into 403 on the public stats pages

### DIFF
--- a/rconweb/api/views.py
+++ b/rconweb/api/views.py
@@ -88,7 +88,7 @@ def get_version(request):
 
 @csrf_exempt
 @require_http_methods(["GET"])
-@ratelimit(key='ip', rate='10/m')
+@ratelimit(key='ip', rate='60/m')
 def get_public_info(request):
     cached_cur_map = MapsHistory().get_current_map()
     if not cached_cur_map:


Fragger (approved):
I'm curious if this will be enough for the many installations with reverse proxy's in front of everything since Django will only be seeing the proxy's IP.

FlorianSW:
> since Django will only be seeing the proxy's IP.

Are you sure? If that is the case, we should definitely revisit this change: reverse proxies generally add a X-Forwarded-For header, which we should take instead. On top of that, we would need to gave some kind o "trusted proxies" config, where the server admin can enter the reverse proxy IP, so that we only take the header when the request is coming from that proxy (otherwise it would be easy to bypass the ratelimiter).

FlorianSW (approved):
apart from that, this looks okayish :)

Fragger:
> Are you sure?

I'm not but what I was seeing with some brief reading.

Fragger:
https://github.com/jsocol/django-ratelimit/issues/55 seems to be correct, maybe we do something like that

Fragger:
Did some digging and testing on a live crcon and rate limit is just seeing the reverse proxy IP as the one used for the rate limit key.  Maybe we do an env variable that is end user configurable for what header to use as the IP so different setups can all be accommodated for, be it X-Forwarded-For, CF-Connecting-IP, or not set set for keeping the default REMOTE_ADDR.  We could do trusted proxies as well but generally if that's being used I'd think a setup should already be where you can't access it except through the proxy.

MartijnHols:
A properly configured reverse proxy will update the IP with the appropriate header, but only if it is validated to avoid just about anyone being able to spoof the IP by providing the header when by passing the reverse proxy. This shouldn't need to be solved in this project and allowing users to specify a header without doing any sort of validation will only open them up for exploitation

see nginx's `set_real_ip_from` and `real_ip_header`

PR #1354, opened 2026-08-13
Add CLIENT_IP_HEADER .env var and middleware

Fragger:
This along with #1352 should fix ratelimit for those with reverse proxies in front of their CRCON installs.  Without this ratelimit only see's the proxies IP and thus everyone shares the limit.

### DIFF
--- a/rconweb/api/middleware.py
+++ b/rconweb/api/middleware.py
@@ -1,0 +1,61 @@
+import logging
+import os
+
+logger = logging.getLogger("rconweb")
+
+
+class ClientIpHeaderMiddleware:
+    """
+    Rewrites ``request.META["REMOTE_ADDR"]`` using the value of a
+    configurable HTTP header, so that ``django-ratelimit`` (and anything
+    else in Django that relies on ``REMOTE_ADDR``, e.g. IP bans) sees the
+    real client IP when CRCON is running behind a reverse proxy / load
+    balancer, rather than the IP of the proxy itself.
+
+    Configure the header to trust via the ``CLIENT_IP_HEADER`` environment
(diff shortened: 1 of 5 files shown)


rconweb/api/middleware.py
```diff
@@ -0,0 +1,61 @@
...
+        self.meta_key = "HTTP_" + header.upper().replace("-", "_") if header else None
+
+        if self.meta_key:
+            logger.info(
+                "ClientIpHeaderMiddleware enabled: REMOTE_ADDR will be set from the '%s' header (%s)",
+                header,
+                self.meta_key,
+            )
+
+    def __call__(self, request):
```
FlorianSW, line 50: this should definitely have the possibility to configure trusted/known proxy servers. When configured, the actual remote_addr should be checked to be one of the configured trustworthy proxies, and if they do not match, do not replace the IP (or even better: reject the request and return a 401/403/something).

FlorianSW (changes requested):

rconweb/api/middleware.py
Fragger: I started looking into this more but it looks like all the backend ever sees for the remote_addr is the frontend's ip.  We will probably need to look at moving the ratelimits to the frontend on nginx like emathy is talking about in the discord.

FlorianSW: I believe you have to add a bit of configuration to nginx to ingest the x-forwarded-for header. It is not automatic. While I think about it, this probably means we do not need the configurable header on this side at all 😅

Fragger: Hopefully #1379 takes care of this in a way that makes a little more sense.

PR #1362, opened 2026-08-13
Fix map recording in log_loop

Dorfieeee:
The service could not parse map as it used general util functions and so it never started recording stats and map was always `unknown`


FlorianSW (approved):
Tbh, I don't know if I get all of the changes correctly, but they look right 👍

Dorfieeee:
Basically when it registered a new HLL:V map(match) it could never parse and save the correct map from the log as it tried to parse it based on HLL map collection. The map was always `unknown` and then inside the loop service it compares the saved map `unknown` with the current map taken from game_state that was correctly parsed resulting in never running the stats recording as it thought the map has changed but not yet started

timraay (approved):

PR #1363, opened 2026-08-14
UI/frontend game env

Dorfieeee:
pass HLL_GAME to frontend and base assets on it it will be used for other things later


FlorianSW (commented):
Wouldn't it make sense to have assets that are shared by both game types (when they are the same), and game-specific assets? So we could avoid duplicating them?

Dorfieeee:
I was planning to work this backwards once the time for UI changes comes but sure there are images that can be shared and some can be removed entirely

Dockerfile-frontend
```diff
@@ -10,26 +14,34 @@ COPY rcongui/ .
 RUN npm ci
 ENV REACT_APP_API_URL=/api/
 COPY .git/ .git/
-COPY assets/images/ /code/public
-COPY assets/rcongui/ /code/public
+COPY assets/shared/ /code/public
+COPY assets/${HLL_GAME}/images/ /code/public
```
FlorianSW, line 18: uhm, doesn't this mean that the shared docker image we built for people on docker hub does only contain either game images? Depending on what env variable we use during image built?

Shouldn't the image have all assets for both games and decide on what to use based on tan indicator from the backend?

FlorianSW (changes requested):

PR #1367, opened 2026-08-15
Add temporary placeholder art for the six HLL:V maps

Laurence-Walpole:
**These are generated stand-ins, not real captures.** Flagging that up front so it's a deliberate choice rather than a surprise — see *What these actually are* below. The intent is a stopgap that's replaceable in place, not a claim that Vietnam art is done.

### The problem

No Vietnam map art ships, on `master` or in v12.0.1 — `assets/images/` has 60 WWII icons, 60 banners and 20 tactical maps, and no `wdev*` anywhere. Every Vietnam layer carries `image_name = "<map id>-day.webp"`, and the frontends request:

```
admin   /maps/icons/<id>-day.webp    /tac-maps/<id>.webp
public  /maps/<id>-day.webp          /maps/icons/<id>-day.webp
```

None of those exist, so every Vietnam layer draws a broken image: the map list, the change-map page, the rotation and votemap builders, the objectives page backdrop, and the public scoreboard.

Worth noting for anyone who has looked for this and not found it: it never appears as a 404. nginx's SPA fallback answers the missing paths with `index.html` at **HTTP 200, `Content-Type: text/html`**, so the browser simply receives HTML where it expected a webp.

### What these actually are

Six maps × three sizes = 18 files, 120 KB total. Each is a flat colour gradient carrying the map's real name, three-letter tag and ID, generated with ImageMagick. The tactical maps additionally carry a 5×5 grid and are labelled "placeholder tactical map".

They match the existing conventions exactly — 64×36 icon, 960×540 banner, 1000×1000 tactical map — because the filenames and dimensions are the only part that needs to be stable. Real captures drop straight over the same paths with no code change.

Six maps, all `_day`:

| ID | Map |
|---|---|
| `wdeva` | Vạn Tường |
| `wdevb` | Quảng Ngãi |
| `wdevc` | Huế Outskirts |
| `wdevd` | Đăk Tô Airfield |
| `wdeve` | Cam Ranh Port |
| `wdevf` | Thanh Hòa Bridge |

If you would rather hold out for real art, the alternative fix is a graceful fallback instead: there is currently no `onError` on any of the six `` sites, and `getMapLayerImageSrc` falls back to `unknown.webp` only when `image_name` is *undefined*, never when the file is absent. Happy to send that instead of, or as well as, this — just say which you prefer.

### Scope

Adds files only. No existing asset is modified, so WWII is untouched. Verified running on two live Vietnam servers: all 18 paths serve as `image/webp` on both the admin and public sites, and the WWII assets still serve unchanged.


Dorfieeee:
It seems Vietnam images provided in #1369

Dorfieeee (approved):

PR #1368, opened 2026-08-15
Don't let an unreachable game server block CRCON's web UI from booting

cemathey:
Tested this and it works on a container without a valid RCON connection to a game server.

Worth fixing because otherwise people can't get to their admin portal if they only have a single game server and they can't use the web UI or API to make valid offline changes (auto settings, etc.)

`remove_orphaned_map_ids` runs in the startup script.


Fragger (approved):
Looks good to me and works.  Turned a server off and restarted CRCON with this and everything came up.

FlorianSW (approved):

PR #1371, opened 2026-08-16
Feat/split assets

Dorfieeee:
Frontend assets are now split into shared, hll, and hllv directories. A single Docker image includes both game asset sets, while `HLL_GAME `is passed at container runtime. The `entrypoint `validates this value, generates runtime frontend configuration, and copies the selected game’s assets into the public directories. This allows the same prebuilt image to support either game without rebuilding.

NOTE: Passing `HLL_GAME` is the foundation for further `frontend` work


docker-templates/one-server.yaml
```diff
@@ -74,6 +74,7 @@ x-frontend: &frontend
   image: ${FRONTEND_DOCKER_REPOSITORY}:${TAGGED_VERSION}
   environment:
     RCONWEB_EXTERNAL_ADDRESS: ${RCONWEB_EXTERNAL_ADDRESS}
+    HLL_GAME: ${HLL_GAME:-hll}
```
Fragger: This should go down in environment in the frontend_1 section since it is unique to that.

docker-templates/ten-servers.yaml
```diff
@@ -177,6 +178,9 @@ services:
           - supervisor
   frontend_2:
     <<: *frontend
+    environment:
+      RCONWEB_EXTERNAL_ADDRESS: ${RCONWEB_EXTERNAL_ADDRESS}
```
Fragger: Don't need RCONWEB_EXTERNAL_ADDRESS in the individual frontend_n's since it is in the frontend template section

docker-templates/ten-servers.yaml
```diff
@@ -83,6 +83,7 @@ x-frontend: &frontend
   image: ${FRONTEND_DOCKER_REPOSITORY}:${TAGGED_VERSION}
   environment:
     RCONWEB_EXTERNAL_ADDRESS: ${RCONWEB_EXTERNAL_ADDRESS}
+    HLL_GAME: ${HLL_GAME:-hll}
```
Fragger: Same as the one server file down in frontend_1

Fragger:
~Should the map images for hllv be working in this?~ I forgot to update my compose to have HLL_GAME in the frontend XD

Fragger (approved):
Looks good and works when I tested.

PR #1378, opened 2026-08-17
Fix/rate limit

Dorfieeee:

PR #1379, opened 2026-08-17
Add trusted-proxy support for X-Real-IP forwarding

Fragger:
nginx now sets X-Real-IP on every request proxied to the backend. Its value is nginx's own $remote_addr, which is only rewritten from a configurable header (RCONWEB_REAL_IP_HEADER, default X-Forwarded-For) when the request's direct peer matches an entry in RCONWEB_TRUSTED_PROXIES; otherwise it stays the address of whoever actually connected. This lets CRCON sit behind a reverse proxy or a Cloudflare Tunnel without letting clients spoof their own IP.

- rcongui/entrypoint.sh: generate /etc/nginx/realip.conf from the two new env vars on every container start, and run nginx -t before starting.
- rcongui/nginx.conf: include the generated config and forward X-Real-IP from every backend-proxied location.
- rconweb/api/middleware.py: new XRealIpMiddleware trusts X-Real-IP and overwrites REMOTE_ADDR, so django-ratelimit's key="ip" (login, scoreboard, API) see the real client instead of the frontend container's address.
- default.env, docker-templates/*.yaml: document and wire through RCONWEB_REAL_IP_HEADER / RCONWEB_TRUSTED_PROXIES.


Fragger:
Going with #1378 instead

PR #1380, opened 2026-08-17
HLLV support for config files and auto settings

Dorfieeee:
Updated all config-backed automods to use `(game, server_number)` from the active `ServerInfo`. This includes squad automods, TK autoban, API validation/getters, and game-specific role/map/weapon validation.

There is DB migration where the UserConfig is modified:
[key, value] -> [name, game, server_number, value]
Every config is migrated as game=hll by default. The key was used to query the config and its value was in {server_number}_{config_name} format. The number and name is extracted from the key during the migration process.

UserConfig is now being validated depending on the HLL_GAME so for example 'officer' role will not validate when HLL_GAME=hllv.

I am not 💯 sure it all works correctly. I was able to successfully warn a player for a disallowed role set in Seeding automod.


Fragger (approved):
Upgrade instructions for already setup hllv servers if you want your settings migrated, skip if you want to start fresh:
Before CRCON update backup settings like in https://github.com/MarechJ/hll_rcon_tool/wiki/Additional-Setup-%E2%80%90-Backup-and-restore-settings for each hllv server, take note of the CRCON server number for each and replace the three 1's below with the number and run the command for each server/number:
```
cd /root/hll_rcon_tool
docker compose exec -it backend_1 python -m rcon.cli get_user_settings 1 /logs/backup_settings_1.json
```

Upgrade CRCON

Restore settings, again run command for each hllv server/number:
```
cd /root/hll_rcon_tool
docker compose exec -it backend_1 python -m rcon.cli set_user_settings 1 /logs/backup_settings_1.json --dry-run False
```

Bounce CRCON for settings to take effect:
```
docker compose down
docker compose up -d --remove-orphans
```

PR #1382, opened 2026-08-18
Bind mount /config as ro in docker compose

Fragger:
Probably should have always been like this


FlorianSW (approved):
Looks good, but didn't test it in a local docker run.

PR #1398, opened 2026-08-20
fix: configure Django settings before ASGI imports

fxsobr:
Daphne failed to start because `api` was imported before `DJANGO_SETTINGS_MODULE` was set, causing `ImproperlyConfigured` and breaking WebSocket connections and live log streaming.

Move the settings initialization before the guarded imports to prevent Ruff from reordering them again.


Fragger (approved):
Looks good and cleans up that error on backend startup, thanks!

PR #1393, opened 2026-08-19
Feat/hllv role faction icons

wienni112:
## Summary

Adds missing frontend assets for Hell Let Loose: Vietnam faction and role support.

The HLL Vietnam roles are already exposed by the game/API, but several corresponding role icons were missing from the CRCON frontend assets. This caused missing or incorrect icons to be displayed for these roles.

This PR adds the required role icon variants as well as the missing NVA faction emblem.

## Added role icons

The following HLL Vietnam roles now have dedicated frontend assets:

- `grenadier`
- `helicopterlogisticsofficer`
- `helicopterpilot`
- `mortargunner`
- `mortarobserver`
- `mortarsupport`
- `specialist`
- `squadleader`

Each role includes both variants used by the existing frontend:

- `<role>.png`
- `<role>_black.png`

This adds 16 role icon files in total.

## Squad Leader

HLL Vietnam exposes the squad leader role as:

`"squadleader"`

The existing frontend did not contain corresponding:

- `squadleader.png`
- `squadleader_black.png`

The HLL Vietnam Squad Leader insignia matches the existing Officer insignia closely, so the existing Officer artwork is intentionally reused for these two assets.

This keeps the visual style consistent with the existing CRCON role icons while allowing the frontend's existing role-based asset lookup to work without additional special-case logic.

## Added NVA faction icons

The following faction assets were added:

- `nva.webp`
- `nva_dark.webp`
- `nva_colored.webp`

These provide the missing NVA faction emblem variants required by the existing frontend team/faction icon lookup.

## Asset paths

Role icons:

`assets/images/icons/roles/`

Faction icons:

`assets/images/icons/teams/`

## Testing

The assets were tested on a live HLL Vietnam CRCON instance.

Verified:

- NVA faction emblem displays correctly
- HLL Vietnam role icons load correctly
- regular role icon variants display correctly
- black role icon variants display correctly
- Squad Leader resolves to its new asset
- existing frontend asset lookup works without additional frontend code changes

The assets were also tested directly in the running frontend container under:

`/var/www/icons/roles/`

## Scope

This PR only adds frontend image assets.

It does not modify:

- role parsing
- role validation
- squad type detection
- squad leader detection
- automod behavior

HLL Vietnam role handling and leader detection are handled separately.

PR #1399, opened 2026-08-20
Add Vietnam roles and teams icons

Dorfieeee:
This expands on #1393


Fragger (approved):
Does this replace #1393 then?

Dorfieeee:
@Fragger yes it does. It did not have the team icons and it was PR based on outdated master branch before the assets split. I am not merging that...

PR #1404, opened 2026-08-21
Downgrade redis back to 7.4.0 but bump to 7.4.1

Fragger:
Pull #1386 broke barricade


ElGuillermo (approved):

Fragger:
I think #1405 will be preferable to this, haven't noticed anything else that is broken besides that.

FlorianSW:
Superseded with #1405

PR #1405, opened 2026-08-21
Disable socket timeout like redis-py 7.x

Fragger:
Looks like this is an alternative fix to #1404 suggested from https://github.com/redis/redis-py/issues/4091


ElGuillermo (approved):

FlorianSW (approved):
fine for me like that. It would probably be good to investigate what the BZPOPMIN timeout is, and make the socket timeout a bit higher than that, right?

Fragger:
> fine for me like that. It would probably be good to investigate what the BZPOPMIN timeout is, and make the socket timeout a bit higher than that, right?

I believe disabling the socket timeout on this side does just that.

PR #1409, opened 2026-08-24
Blacklist page refactor

Dorfieeee:
A complete overhaul of the `Records > Blacklist page`
This change introduces better design with two different views - cards and tabular
It also fixes the message template bug so the `reason` templates can be used in the form now.
There is a support for creating multiple reports at once.

> Cards view
image
> Tabular view
image


Fragger:
It looks like get_blacklist_records?exclude_expired=false is not being parsed into a boolean so that being set makes it always true and this is always filtering out expired records now.  See https://github.com/MarechJ/hll_rcon_tool/blob/b7f0b21d10066a5d69eb4d853ec4d4aecb0f5845/rconweb/api/views.py#L272 either just don't set it or we need to fix get_blacklist_records to parse it properly.

Fragger (approved):
Looks good and all seems to work from what I've tried.  Only thought is maybe want to remember if the user is on card or table view instead of needing to change that each time.

PR #1411, opened 2026-08-25
Fix redis migration to exclude global db by default

Dorfieeee:
I found `votemap:version` in the global redis database = 0 which was caused by a redis db discovery in a migration code.


Fragger (approved):
Looks good and makes sense

PR #1414, opened 2026-08-25
Enhance team detection logic in PlayerStats based on units

Dorfieeee:
Since the unit changes are being tracked the team detection does not need to solely rely on kills


Fragger (approved):
This looks good and works well from a couple of matches I checked :)

FlorianSW (approved):

PR #1422, opened 2026-08-27
Fix Maps → Objectives page crashing on Conquest servers

cemathey:
Conquest doesn't return the Sector_* parameters SetSectorLayout expects (only Warfare/Offensive do), so get_objective_rows() always 400s. The frontend only excluded Skirmish from the objectives page.

- Fold Conquest into the existing unsupported-mode check alongside Skirmish (helpers.js, hll/index.jsx).
- Generalize error handling by matching the API command name + HTTP 400 for any future game modes that fail.

Built and verified this on a server that was set to conquest:

image

### DIFF
--- a/rcongui/src/pages/settings/maps/objectives/error.jsx
+++ b/rcongui/src/pages/settings/maps/objectives/error.jsx
@@ -3,12 +3,13 @@ import { Stack, Typography, Button, Paper } from "@mui/material";
 
 export default function RouteError() {
   const error = useRouteError();
-  const isSkirmishMode =
-    error.message === "skirmish" ||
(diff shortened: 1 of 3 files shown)


Dorfieeee:
How could this happen? HLL:WW2 does not have conquest yet.

Fragger:
The experimental branch on WW2 has Conquest

Dorfieeee:
I mean we can exclude Conquest from this as it has no effect but the error arises from the fact there are no Conquest map layers in the codebase so get_objective_rows always returns 400 as the map id is not recognized

Dorfieeee (approved):

Issue #1285, opened 2026-04-27
duplicate key value violates unique constraint "unique_log_line"


Fragger:
To add to a discussion around #1279 without hijacking the pull request:

I've previously noticed the duplicate log line errors that get emitted to the logs somewhat often, had done some digging into it and to me it seemed like the `.order_by(desc(LogLine.event_time))` in `_get_new_logs` in `rcon/logs/recorder.py` was part of the problem since its non-deterministic when there happens to be logs with the same time.  Ex: the match starting when a player joins or other messages that generate together.  Its something the current test cases around the function fail to capture correctly since the behavior of postgres changes from just doing a reverse table scan with a small mostly empty table (like during tests) to using indexes once the table is full enough (like in actual use).  I have an instance running with the order by changed to `.order_by(desc(LogLine.event_time), LogLine.id)` and that seems to do the trick although I just hadn't gotten the time to put into making test cases that show why its needed and works.

Also doing some profiling, `_get_new_logs` is called very often and thus the query here is done a lot, it might be good to add an index like `CREATE INDEX ix_log_lines_server_event_time_id ON public.log_lines USING btree (server, event_time DESC, id)`.

FlorianSW:
> I just hadn't gotten the time to put into making test cases that show why its needed and works

it will probably be difficult to make a test case for that case. Ordering on a row that _can_ have the same value for multiple rows is simply not stable and guaranteed at all. However, on a small dataset that doesn't change (e.g. in a test case), the order might _look_ stable and be the same on multiple queries. In a big table with constantly changing contents, multiple queries with the same value in the column used with order by might return different orderings (or even the same but "wrong" ones, based on the expectation of the code that did the query).

Adding a unique key to the order by clause is therefore something I would support, even when we are not able to make a proper test case for that. Simply as this makes sense from how the database is ordering and when the order is considered "stable".

One thing that came to my mind, however:
```
.order_by(desc(LogLine.event_time), LogLine.id)
```
Are you really ordering ascending, which is the default, isn't it? Wouldn't it make more sense to sort desc as well, meaning that we always get the "latest" log lines on top? 🤔 

> Also doing some profiling, _get_new_logs is called very often and thus the query here is done a lot, it might be good to add an index like CREATE INDEX ix_log_lines_server_event_time_id ON public.log_lines USING btree (server, event_time DESC, id).

Having an index on this lookup makes sense, absolutely. We already seem to have an index that covers this partially, only:
```
[...]

PR #1423, opened 2026-08-27
Fix/log recorder last log

Fragger:
This should address #1285.  Insert logs oldest to newest and then tie break matching timestamps by the newer log line id.  Add some indexes to cover how log lines gets queried by game logs and the recorder.

### DIFF
--- a/alembic/versions/5cf1dd099fd3_improve_indexing_on_log_lines.py
+++ b/alembic/versions/5cf1dd099fd3_improve_indexing_on_log_lines.py
@@ -1,0 +1,44 @@
+"""Improve indexing on log_lines
+
+Revision ID: 5cf1dd099fd3
+Revises: 3f12a7b9c4d1
+Create Date: 2026-08-24 18:00:16.458454
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '5cf1dd099fd3'
+down_revision = '3f12a7b9c4d1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("CREATE EXTENSION IF NOT EXISTS pg_trgm;")
+    op.drop_index(op.f('ix_log_lines_player1_steamid'), table_name='log_lines')
+    op.drop_index(op.f('ix_log_lines_player2_steamid'), table_name='log_lines')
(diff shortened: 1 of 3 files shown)


alembic/versions/5cf1dd099fd3_improve_indexing_on_log_lines.py
```diff
@@ -0,0 +1,44 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("CREATE EXTENSION IF NOT EXISTS pg_trgm;")
+    op.drop_index(op.f('ix_log_lines_player1_steamid'), table_name='log_lines')
+    op.drop_index(op.f('ix_log_lines_player2_steamid'), table_name='log_lines')
+    op.create_index('ix_log_lines_player1_name_event_time', 'log_lines', ['player1_name', 'event_time'], unique=False)
+    op.create_index('ix_log_lines_player1_name_trgm', 'log_lines', ['player1_name'], unique=False, postgresql_using='gin', postgresql_ops={'player1_name': 'gin_trgm_ops'})
```
FlorianSW, line 22: Why does this need to be a GIN index, especially with trgm?

Fragger: You're talking GIN vs GiST?  It has to be one of them with trgm.  From what I've looked at GIN seems to be the better option for player names and the type columns since they both have tons of repetitive data

FlorianSW: Sorry, no, maybe my question was not really good: Why are we using trgm instead of a "classical" normal index? 🤔

Fragger: It's for the LIKE matches, anytime the exact toggle isn't switched on that's what gets used in those queries.

FlorianSW: Ah, oh damn, I didn't remember that we have that feature :D Then I understand it, I thought we have "exact" matches only.

FlorianSW (approved):

caranci:
I just merged master into my deployment branch, and the index creation took quite a long time (about 20 minutes).

The maintenance container was marked as unhealthy and the full container bring-up was stopped... can `create index concurrently` work here to avoid blocking until the indexes are created?

@Fragger

PR #1425, opened 2026-08-27
Fix Seeder VIP notifications for HLL Vietnam

wienni112:
## Summary

This fixes two issues affecting Seeder VIP Discord notifications:

* Dry-run notifications were indistinguishable from live notifications.
* The current map could be displayed as `unknown Domination`, especially on Hell Let Loose: Vietnam servers.

## Changes

* Adds a clear dry-run prefix to Seeder VIP Discord notifications:

```text
🧪 DRY RUN – SIMULATION
No VIPs or in-game reward messages will be sent.
```

* Applies the prefix to both:

  * seeding progress notifications,
  * seeding complete notifications.
* Uses the current map returned by `get_gamestate()` instead of the controller's potentially stale `current_map` property.
* Adds tests covering dry-run and live Discord message formatting.

## Previous behavior

Seeder VIP notifications sent to Discord during dry-run mode looked identical to live notifications. Administrators could therefore not tell whether VIP rewards and in-game reward messages were actually being sent.

The Discord embed also used `rcon_api.current_map.pretty_name`. On HLL Vietnam servers, this property could remain initialized as unknown and produce:

```text
Current Map
unknown Domination
```

## New behavior

Dry-run notifications are clearly identified as simulations.

The current map is taken from the same gamestate used for player counts and remaining time. This correctly displays Vietnam maps, for example:

```text
Current Map
Vạn Tường Warfare
```

Existing VIP rewards and in-game player messages are unchanged.

## Testing

* Verified Python syntax using `py_compile`.
* Verified the changes using `git diff --check`.
* Added tests for dry-run and live message formatting.
* Tested the patched Seeder VIP service against a live HLL Vietnam server.
* Confirmed that:

  * dry-run notifications are visibly marked,
  * the correct player counts are displayed,
  * `Vạn Tường Warfare` is displayed instead of `unknown Domination`,
  * the `seed_vip` process restarts and remains running without errors.

### DIFF
--- a/rcon/seed_vip/service.py
+++ b/rcon/seed_vip/service.py
@@ -25,6 +25,14 @@ from rcon.user_config.seed_vip import SeedVIPUserConfig
 
 logger = getLogger(__name__)
 
+DRY_RUN_DISCORD_PREFIX = (
+    "🧪 **DRY RUN – SIMULATION**\nNo VIPs or in-game reward messages will be sent.\n\n"
+)
+
+
+def format_discord_message(message: str, dry_run: bool) -> str:
+    return f"{DRY_RUN_DISCORD_PREFIX}{message}" if dry_run else message
+
 
 def run():
     config = SeedVIPUserConfig.load_from_db()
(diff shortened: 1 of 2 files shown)


rcon/seed_vip/service.py
```diff
@@ -25,6 +25,14 @@
 
 logger = getLogger(__name__)
 
+DRY_RUN_DISCORD_PREFIX = (
+    "🧪 **DRY RUN – SIMULATION**\nNo VIPs or in-game reward messages will be sent.\n\n"
```
FlorianSW: Isn't that a bit longish? What about just the first line? And then appending the message after it without a newline?

FlorianSW (changes requested):

rcon/seed_vip/service.py
wienni112: Good point. I’ll shorten the prefix to the first line and append the notification message directly after it without an additional newline.

PR #1427, opened 2026-08-27
General design improvements

Dorfieeee:
3 new themes:
- HLL WW2
- HLL Vietnam
- AI Sloperino

and High Contrast theme updated

Design coherence:
- make sure buttons and other elements are styled the same way throughout the codebase as some components had static styling that did not reflect the theme styling
- a few responsive changes - the search form on Records > Game Logs page for example which was being squished
- better light/dark theming

> High Contrast
image

> HLL WW2
image

> HLL Vietnam
image

> AI Slopperino
image


Fragger:
Themes look good.  I probably need to see this on a server with players on it but is it smashing the logs on the live stats too small?  I guess the players table just resized before depending on the columns enabled and the rest was left to the logs.

Fragger (approved):

PR #1434, opened 2026-08-31
Make backend build multi-stage, version in package

Fragger:
uv build into a package and use uv-dynamic-versioning/hatchling to put the git tag as the version into the package.  Read the package version instead of needing to ship the .git directory and git in the image


FlorianSW (approved):
Yeah, absolutely awesome! Getting rid of shipping the entire .git directory as well as needing to shell-out to read the git version is really nice :)

PR #1435, opened 2026-08-31
frontend build improvements/exec in entrypoint

Fragger:
nit some things like #1434 and use exec as the final call in the entrypoint script.

### DIFF
--- a/rcongui/entrypoint.sh
+++ b/rcongui/entrypoint.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env sh
 
-export REACT_APP_VERSION=$(cat /code/tag_version)
+REACT_APP_VERSION=$(cat /code/tag_version)
+export REACT_APP_VERSION
 
 game=${HLL_GAME:-hll}
 case "$game" in
@@ -11,7 +12,7 @@ case "$game" in
         ;;
 esac
 
-export HLL_GAME=${game}
+export HLL_GAME="${game}"
 
 # Vite environment variables are compiled into the bundle. Generate the
 # deployment-specific values at container startup instead.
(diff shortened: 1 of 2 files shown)


rcongui/entrypoint.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env sh
 
-export REACT_APP_VERSION=$(cat /code/tag_version)
+REACT_APP_VERSION=$(cat /code/tag_version)
```
FlorianSW, line 3: any reason for splitting this into two distinct statements?

rcongui/entrypoint.sh
```diff
@@ -11,23 +12,26 @@ case "$game" in
...
 printf '%s\n' "$runtime_config" > /var/www/runtime-config.js
 printf '%s\n' "$runtime_config" > /var/www_public/runtime-config.js
 
-# Copy game specific assets to nginx root
-cp -R "/code/assets/$game/images/." /var/www/
-cp -R "/code/assets/$game/rcongui/." /var/www/
-cp -R "/code/assets/$game/images/." /var/www_public/
-cp -R "/code/assets/$game/rcongui_public/." /var/www_public/
+# Link game specific assets to nginx root
+cp -Rs "/code/assets/$game/images/." /var/www/
```
FlorianSW, line 24: good catch! 🥳

rcongui/entrypoint.sh
```diff
@@ -11,23 +12,26 @@ case "$game" in
...
+cp -Rs "/code/assets/$game/rcongui/." /var/www/
+cp -Rs "/code/assets/$game/images/." /var/www_public/
+cp -Rs "/code/assets/$game/rcongui_public/." /var/www_public/
 
 if [ ! -f "/certs/cert.crt" ] || [ ! -f "/certs/key.key" ]; then
     echo "No certificates found. Generating self signed"
-    openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /certs/key.key -out /certs/cert.crt -subj "/C=US/ST=Oregon/L=Portland/O=Company Name/OU=Org/CN=$RCONWEB_EXTERNAL_ADDRESS" 
+    openssl ecparam -name secp384r1 -genkey -noout -out /certs/key.key
```
FlorianSW, line 31: This actually changes the key we use for these certs from RSA to ECDSA, which might not be supported by every client.
Two things that make this argument pointless: Any reasonable up-to-date browser (which this is intended to be used against) should already support ECDSA. Anda s this is a self-signed cert, we could safely assume that anyone who wants/needs control over what key to use, that they can do so by generating a keypair on theirown.

FlorianSW (commented):
just one little nitpick (first comment). Apart from that, looks good :)

rcongui/entrypoint.sh
Fragger: It fixes a ShellCheck warning, doesn't really matter here (especially since this doesn't have set -e) but one less line lighting up as a warning for me :joy: https://www.shellcheck.net/wiki/SC2155

FlorianSW (approved):

PR #1436, opened 2026-08-31
exec in entrypoint/handle signals better, tidy up

Fragger:
Use exec in entrypoint for single process services.  Fixup backend/web service so it tracks both daphne and gunicorn, gracefully shuts down and cleans up if either die, and have docker restart them.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -5,76 +5,97 @@ set -x
 
 env
 # Only run the database migrations in the maintenance container
-if [ "$1" == 'maintenance' ]
-then
-  alembic upgrade head
-  # Convert old stye player IDs to new style (md5 hash)
-  # TODO: we can eventually remove this in a few releases once old installs have updated their game servers and CRCON
-  # SERVER_NUMBER is mandatory and not otherwise set in the maintenance container; only want it to run once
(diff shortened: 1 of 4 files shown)


entrypoint.sh
```diff
@@ -5,76 +5,97 @@ set -x
...
+
+if [ "$1" == 'supervisor' ]; then
+    if [ "$HLL_HOST" == '' ]; then
+        echo "HLL_HOST not set"
+        exit 1
+    fi
+    sleep 10
+    env >> /etc/environment
+    export LOGGING_FILENAME="supervisor_${SERVER_NUMBER}.log"
+    if [ -f "/config/supervisord_${SERVER_NUMBER}.conf" ]; then
```
caranci, line 96: i think the previous code would fallback to `supervisord.conf` if `supervisord_N.conf` wasn't valid whereas the change will not.
just making sure this is intended since it wasn't mentioned in the comment.

Fragger: Yes and good to mention actually.  It would also start the 2nd instance if the first crashed for some reason, neither which is probably what we really want.  If the intended config is invalid then the container should probably fail to start like this will now instead of silently using the fallback.

caranci: Sounds good!

FlorianSW (approved):
Looks good. However, there should definitely be a migration guide for people to do for their compose files :)

PR #1444, opened 2026-09-02
Add an opt-in Python process supervisor so service loops can share a …

caranci:
Add an opt-in Python process supervisor so service loops can share a preloaded interpreter via fork/CoW instead of paying a full rcon.cli import per Supervisord child.

- Parse existing supervisord.conf and speak the XML-RPC subset the Services UI uses (getAll/get/start/stop, Supervisord state and fault codes).
- Dispatch known Python loops through supervisor-only run_<name> adapters with lazy imports; do not wire these into rcon.cli yet.
- Fork adapted programs from a helper that preloads hllrcon and rcon.maps so extra services mostly add unique heap; CRCON_SUPERVISOR_FORK=0 falls back to a fresh worker interpreter.
- Exec INI command= as-is for programs without an adapter (workers, cron, scheduler); backend/gunicorn stays out of scope.
- Gate the arbiter behind CRCON_USE_PROCESS_SUPERVISOR (default off) in entrypoint.sh, default.env, and compose templates.
- Cover spawn, RPC locking, registry/fork/worker paths with tests; add pytest-cov as a dev dependency.


caranci:
I have this running on our CRCON install... seems to work fine.

Fragger:
Haven't looked further but seeing some log line output changes with some weird characters after this.
`[2026-09-02 03:51:34,617][INFO]` before
and then
`[2026-09-02 03:52:12,056][ESC[32mINFOESC[0m]` after

FlorianSW:
> Haven't looked further but seeing some log line output changes with some weird characters after this. `[2026-09-02 03:51:34,617][INFO]` before and then `[2026-09-02 03:52:12,056][ESC[32mINFOESC[0m]` after

These should be escape sequences for colors 🤔 Not sure, why these would be added with this PR, though 🙈 Do you run your container with a tty maybe? 🤔

Fragger:
It might be that this is hooking the console output and redirecting to a file instead of letting the logger config do the file write.  Looks like the color escapes are part of the default log format string for console output but not file.

caranci:
Sorry. Busy with work and long weekend. :)

It looks like the root-cause was the fork was freezing the logging config for the children to the supervisor config... and then pushing the console output to the logfile.

The solution is to reconfigure the child logging to the correct logger after the fork.

Fragger (approved):
This seems to work from what I've tried.  Since its optional I say we ship it, take feedback from those who try it and do improvements as necessary.

PR #1449, opened 2026-09-04
chore(i18n): refine zh-Hans translations

farshore07:
- Fix a character encoding error
- Localize part of the translated text


ElGuillermo (approved):

farshore07:
wow so quick

PR #1451, opened 2026-09-05
Revert add_vip RCON payload key from Comment back to Description

cemathey:
Revert back to the correct command name to fix the `AddVip` command.


FlorianSW (approved):

Fragger:
This breaks the comment field on WW2 but fixes it on Vietnam

PR #1453, opened 2026-09-05
Add colored Live Log names and expandable player IDs

kilo689:
## Summary

- Adds optional team-colored player-name columns to Live Logs in the drop down menu.
- Adds optional team-colored player-name "message" columns to Live Logs  in the drop down menu.
- Adds an Expand/Collapse control that displays complete player IDs beneath names.
- Keeps the new columns hidden by default and preserves existing table preferences.
git1
git2
git3
git4

## Testing

- Production frontend build passed.
- Prettier formatting passed.
- Git diff validation passed.
- Verified the branch contains only three Live Logs source files.


Fragger (approved):
Looks good and all checks out and works for me when testing :)

PR #1477, opened 2026-09-21
Fix game logs recovery worker

Dorfieeee:
1. It wasn't correctly IDing logs
2. If it found server logs there were not saved it did not store them


Fragger (approved):
Looks good, hard to test until some matches run through it besides the actual test cases so guess we will see. :)

